PR #4, opened 2015-10-23
Add StaleSequenceException for Ringbuffer 

cangencer:
Exception is thrown when the read sequence is less than the head sequence.

### DIFF
--- a/Hazelcast.Net/Hazelcast.Core/HazelcastException.cs
+++ b/Hazelcast.Net/Hazelcast.Core/HazelcastException.cs
@@ -57,4 +57,20 @@ namespace Hazelcast.Core
         public DistributedObjectDestroyedException(Exception cause) : base(cause.Message) { }
  
     }
+
+    /// <summary>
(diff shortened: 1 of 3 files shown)


Hazelcast.Net/Hazelcast.Core/HazelcastException.cs
```diff
@@ -57,4 +57,20 @@ public class DistributedObjectDestroyedException : HazelcastException
...
+    {
+        public StaleSequenceException()
+        {
+        }
+
+        public StaleSequenceException(string message) : base(message)
+        {
+        }
+    }
 }
```
ihsandemir: should we add new line here?

ihsandemir:
:+1:

Hazelcast.Net/Hazelcast.Core/HazelcastException.cs
cangencer: As Github lost track of this fork somehow, I will fix the formatting in the documentation PR.

PR #8, opened 2015-10-27
Fixed several issues with ClientNearCache

cangencer:
* When maximum size is reached, the eviction function should evict items according to policy
* On Map.Get() the deserialised result should be stored in the cache
* On Map.GetAll it should not remove items from the collection it is iterating on
* EntryEvents should be handled according to their type
* Disabled Offheap in memory format as it is not supported
* Add several tests for all of the above

### DIFF
--- a/Hazelcast.Test/Hazelcast.Client.Test/ClientNearCacheTest.cs
+++ b/Hazelcast.Test/Hazelcast.Client.Test/ClientNearCacheTest.cs
@@ -14,132 +14,373 @@
 * limitations under the License.
 */
 
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
-using System.Threading;
 using Hazelcast.Client.Proxy;
 using Hazelcast.Config;
 using Hazelcast.Core;
+using Hazelcast.IO.Serialization;
 using Hazelcast.Util;
 using NUnit.Framework;
 
 namespace Hazelcast.Client.Test
 {
-
     [TestFixture]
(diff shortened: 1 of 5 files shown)


Hazelcast.Test/Hazelcast.Client.Test/ClientNearCacheTest.cs
```diff
@@ -14,132 +14,339 @@
...
             {
-                map.Put("key" + i, "value" + i);
+                _map.Put("key" + i, "value" + i);
                 keys.Add("key" + i);
             }
 
-            map.GetAll(keys);
-            var mapImpl = map as ClientMapProxy<object, object>;
-            var clientNearCache = mapImpl.NearCache;
+            _map.GetAll(keys);
```
sancar: I remember we had problems with similar tests. 
The scenario that is causing tests were as follows.
- This getAll populates the near cache.
- An invalidation event caused by map.put comes and invalidates a key. 
- And this results in less than expected number of items in the cache. 

One solution is to open the client after put is done and its events are already fired. 
An example test. 
https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapNearCacheTest.java#L516

cangencer: I've updated the tests to have invalidation disabled, except for those tests that specifically test invalidation, which should prevent this issue from happening.

sancar:
:+1:

PR #9, opened 2015-10-27
Changes to HeapData layout to match the changes in Java client

cangencer:
See https://github.com/hazelcast/hazelcast/pull/6577

### DIFF
--- a/Hazelcast.Net/Hazelcast.IO.Serialization/HeapData.cs
+++ b/Hazelcast.Net/Hazelcast.IO.Serialization/HeapData.cs
@@ -16,24 +16,22 @@
 
 using System;
 using System.Text;
-using Hazelcast.IO;
-using Hazelcast.Net.Ext;
 using Hazelcast.Util;
 
 namespace Hazelcast.IO.Serialization
 {
     public sealed class HeapData : IData
     {
-        internal const int TypeOffset = 0;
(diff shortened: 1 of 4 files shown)


Hazelcast.Net/Hazelcast.IO.Serialization/HeapData.cs
```diff
@@ -87,14 +86,19 @@ public int GetTypeId()
...
         public int GetHeapCost()
         {
             // reference (assuming compressed oops)
-            int objectRef = Bits.IntSizeInBytes;
-            return objectRef + (data != null ? ArrayHeaderSizeInBytes + data.Length : 0);
+            var objectRef = Bits.IntSizeInBytes;
+            return objectRef + (_data != null ? ArrayHeaderSizeInBytes + _data.Length : 0);
+        }
+
+        public bool IsPortable()
```
ihsandemir: why do we have this helper method? If we have this why not have isDataSerializable() method?

cangencer: good question, this method seems to be ported from the Java Client, and seems only used in some test related code. I'll leave it for now in case it will be useful in the future..

ihsandemir:
:+1:

PR #12, opened 2015-10-30
Logging improvements, provide ability to pass own ILoggerFactory

cangencer:
Make it possible to configure the logging level for the console logger


cangencer:
verify

sancar:
:+1:

PR #15, opened 2015-11-02
make sure ExecutionService is used instead of Task.Factory directly

cangencer:
…and give name to all the threads being used.


cangencer:
retry

cangencer:
verify

PR #21, opened 2015-11-13
Make opening a connection asynchronous as part of the invocation.

cangencer:
Change invocations to be timeout based rather than retry count based.


cangencer:
verify

sancar:
:+1:

PR #22, opened 2015-11-16
Update SerializationService to bring it in line with Java client

cangencer:
Update default serializers to be constant, and add serializers for List
and LinkedList

### DIFF
--- a/Hazelcast.Net/Hazelcast.IO.Serialization/StreamSerializerAdapter.cs
+++ b/Hazelcast.Net/Hazelcast.IO.Serialization/StreamSerializerAdapter.cs
@@ -19,11 +19,9 @@ namespace Hazelcast.IO.Serialization
     internal sealed class StreamSerializerAdapter<T> : ISerializerAdapter
     {
         private readonly IStreamSerializer<T> _serializer;
-        private readonly SerializationService _service;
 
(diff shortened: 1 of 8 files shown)


Hazelcast.Net/Hazelcast.IO.Serialization/StreamSerializerAdapter.cs
```diff
@@ -54,29 +52,6 @@ public ISerializer GetImpl()
             return _serializer;
         }
 
-        public override bool Equals(object o)
```
asimarslan: Don't we need Equal and HashCode methods ?

asimarslan:
What about putting the .Net default serialization in the last step as Java did. It won't be compatible with other clients but would work perfectly for .Net clients.

Hazelcast.Net/Hazelcast.IO.Serialization/StreamSerializerAdapter.cs
cangencer: I removed them becausd they were not in any code path (never used)

cangencer:
I've added support for C# Serializable now, which uses type ID -101, which probably should be added to the Java list of constants so that it's reserved.

hasancelik:
verify

cangencer:
verify

asimarslan:
:+1:

PR #26, opened 2015-11-25
Changes to bring serialization in order with the Java client

cangencer:
Global serializer should have lower priority than CLR serialization, unless
an override is specified.


cangencer:
verify

asimarslan:
:+1:

PR #27, opened 2015-11-25
Add IdentifiedDataSerializable as a convenience class

cangencer:
and add an example for using IdentifiedDataSerializable.


cangencer:
verify

asimarslan:
:+1:

PR #44, opened 2016-02-12
Fixed serialization compatibility issue between Java and C# client

### LINKED ISSUE #43: Read/WriteObject in SerializationService non-compliant with Java equivalents
The ReadObject and WriteObject methods within the SerializationService class perform a null check on the object in question by either doing a ReadBoolean to determine if the object is null or a WriteBoolean to record the object as null. 

**ReadObject (C#)**

```
try
            {
                var isNull = input.ReadBoolean();
                if (isNull)
                {
                    return default(T);
                }
....
```

**WriteObject (C#)**
```
var isNull = obj == null;
            try
            {
                output.WriteBoolean(isNull);
                if (isNull)
                {
                    return;
                }
                var serializer = SerializerFor(obj);
                output.WriteInt(serializer.GetTypeId());
                serializer.Write(output, obj);
            }
            catch (Exception e)
            {
                throw HandleException(e);
            }
```

The Java equivalents readObject and writeObject in the AbstractSerializationService class do not read or write any boolean from the stream to check for "nullness". The result of this is that any stream that utilizes these methods will fail because the C# client reads/writes one extra byte (for the "nullness" boolean). This causes reading and writing to fail. ~~A quick check on the audit of both serialization service classes shows that the C# client has always done this and the Java equivalent hasn't done this, meaning these methods would have never worked~~. Updated: According to @cangencer this was changed in 3.5.5. I'm very new to Hazelcast (and C#/.NET itself) as well as Github (go easy on me :)), but can anyone confirm that my findings are indeed an issue? Removing these read/write boolean calls allows me to utilizes these methods.

Here are the Java equivalents to the above:

**readObject**
```
public final <T> T readObject(final ObjectDataInput in) {  
        try {  
            final int typeId = in.readInt();  
            final SerializerAdapter serializer = serializerFor(typeId);  
            if (serializer == null) {  
                if (active) {
                    throw newHazelcastSerializationException(typeId);
                }
                throw new HazelcastInstanceNotActiveException();
            }
            Object obj = serializer.read(in);
            if (managedContext != null) {
                obj = managedContext.initialize(obj);
            }
            return (T) obj;
        } catch (Throwable e) {
            throw handleException(e);
        }
    }
```

**writeObject**
```
@Override
    public final void writeObject(final ObjectDataOutput out, final Object obj) {
        if (obj instanceof Data) {
            throw new HazelcastSerializationException(
                    "Cannot write a Data instance! " + "Use #writeData(ObjectDataOutput out, Data data) instead.");
        }
        try {
            SerializerAdapter serializer = serializerFor(obj);
            out.writeInt(serializer.getTypeId());
            serializer.write(out, obj);
        } catch (Throwable e) {
            throw handleException(e);
        }
    }
```

cangencer:
The serialization of objects was not compatible between Java and C#
as the additional null flag for objects was removed from 3.6 in Java but
not from the C# client

Fixes #43

### DIFF
--- a/Hazelcast.Net/Hazelcast.IO.Serialization/SerializationService.cs
+++ b/Hazelcast.Net/Hazelcast.IO.Serialization/SerializationService.cs
@@ -75,7 +75,8 @@ namespace Hazelcast.IO.Serialization
             _dataOutputQueue = new ThreadLocalOutputCache(this);
             _portableContext = new PortableContext(this, version);
             _dataSerializerAdapter =
-                CreateSerializerAdapterByGeneric<IIdentifiedDataSerializable>(new DataSerializer(dataSerializableFactories));
+                CreateSerializerAdapterByGeneric<IIdentifiedDataSerializable>(
+                    new DataSerializer(dataSerializableFactories));
             _portableSerializer = new PortableSerializer(_portableContext, portableFactories);
             _portableSerializerAdapter = CreateSerializerAdapterByGeneric<IPortable>(_portableSerializer);
             _nullSerializerAdapter = CreateSerializerAdapterByGeneric<object>(new ConstantSerializers.NullSerializer());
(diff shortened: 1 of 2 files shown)


asimarslan:
:+1:

PR #46, opened 2016-02-12
Promote connections to owner connection if one was available

cangencer:
When owner connection drops, if another connection was already available 
it should be promoted to owner


sancar:
:+1:

PR #47, opened 2016-02-19
Use hostname as is for addresses, and throw exception during authentication if address can't be resolved.

cangencer:


cangencer:
verify

cangencer:
verify

cangencer:
verify

sancar:
:+1:

PR #71, opened 2016-05-24
[Backport] Improve heartbeat handling from client

cangencer:
Also replace heartbeat and partition threads with scheduling service.


cangencer:
verify

PR #79, opened 2016-06-07
SSL support for client-server communication

asimarslan:


Hazelcast.Test/Hazelcast.Client.Test/ClientSSLTest.cs
```diff
@@ -0,0 +1,91 @@
...
+            HazelcastClient.ShutdownAll();
+            StopRemoteController(RemoteController);
+        }
+
+        [Test]
+        public void TestMapSSLEnabled()
+        {
+            Setup(true, false, true, ValidCertName);
+            var map = Client.GetMap<string, string>(TestSupport.RandomString());
+            map.Put("key", "value");
```
cangencer: I think it would be clearer and enough to assert that client is connected, by checking lifecycle maybe?

Hazelcast.Net/Hazelcast.Client.Connection/ClientConnection.cs
```diff
@@ -49,6 +50,7 @@ internal sealed class ClientConnection
         private readonly BlockingCollection<ISocketWritable> _writeQueue = new BlockingCollection<ISocketWritable>();
 
         private volatile Socket _clientSocket;
+        private volatile Stream _stream;
```
cangencer: this shouldn't need to be volatile, but instead readonly as it's only created once.

Hazelcast.Net/Hazelcast.Client.Connection/ClientConnection.cs
```diff
@@ -113,11 +115,26 @@ internal sealed class ClientConnection
 
                 _builder = new ClientMessageBuilder(invocationService.HandleClientMessage);
 
+                var networkStream = new NetworkStream(_clientSocket, false);
+                if (clientNetworkConfig.GetSSLConfig().IsEnabled())
+                {
+                    var sslStream = new SslStream(networkStream, false, (sender, certificate, chain, sslPolicyErrors) => RemoteCertificateValidationCallback(sender, certificate, chain, sslPolicyErrors, clientNetworkConfig), null);
```
cangencer: very long line, can be broken in two

Hazelcast.Net/Hazelcast.Client.Connection/IClientConnection.cs
```diff
@@ -0,0 +1,29 @@
+using System;
+using System.Net;
+using Hazelcast.Core;
+using Hazelcast.IO;
+
+namespace Hazelcast.Client.Connection
+{
+    internal interface IClientConnection
```
cangencer: I would prefer we get rid of this interface unless it's needed, as was before.

Hazelcast.Net/Hazelcast.Client.Connection/ClientConnection.cs
```diff
@@ -423,29 +410,12 @@ private void WriteQueueLoop()
...
-                        }
+                        var remaining = _sendBuffer.Remaining();
+                        var pos = _sendBuffer.Position;
 
-                        
+                        _stream.Write(_sendBuffer.Array(), pos, remaining);
 
-                        _sendBuffer.Position += sendByteSize;
-                        //logger.Info("SEND BUFFER CALLBACK: pos:" + sendBuffer.Position + " remaining:" + sendBuffer.Remaining());
+                        _sendBuffer.Position += remaining;
```
cangencer: This part is redundant as _stream.Write() will be sending the whole buffer.

cangencer:
👍 , agreeing that the build.bat file will be updated with another PR to download correct versions from maven directly.

PR #85, opened 2016-08-17
Fix for GetDistributedObjects()

cangencer:
When GetDistributedObjects() is called with existing objects created from
other clients, there would be an exception thrown if the proxy had
generic type arguments.

Now, proxies created through GetDistributedObjects() will have parameters
of type "Object", but can be recreated when asked for with a specific
generic type.


asimarslan:
👍

PR #96, opened 2016-11-22
[FORWARD-PORT]fix unobserved task exception

asimarslan:
fix unobserved task exception
fix client message byte encode
fix invocation service operation timeouts


asimarslan:
Remote Controller does not work with hazelcast 3.8-SNAPSHOT.

PR #98, opened 2017-01-24
Map RemoveAll with predicate support implementation.

asimarslan:


Hazelcast.Net/Hazelcast.Client.Proxy/ClientMapProxy.cs
```diff
@@ -277,6 +277,12 @@ public bool Replace(TKey key, TValue oldValue, TValue newValue)
             return MapReplaceIfSameCodec.DecodeResponse(clientMessage).response;
         }
 
+        public void RemoveAll(IPredicate predicate)
```
ihsandemir: Do we have RemoveAll working correctly when near-cache enabled? I could not locate related code. Is there a code somewhere which handles this: https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client/src/main/java/com/hazelcast/client/proxy/NearCachedClientMapProxy.java#L156

PR #101, opened 2017-02-20
Predicates improved to provide Predicate Builder functionality

asimarslan:


Hazelcast.Test/Hazelcast.Client.Test/PredicateExtTest.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+
+        [Test]
+        public virtual void TestPredicate_key_property()
+        {
+            FillMap();
+            var predicate = Predicates.Key("myint").GreaterThanOrEqual(50);
+            var values = map.Values(predicate);
+            Assert.AreEqual(50, values.Count);
```
ihsandemir: Should we also test if the entries are as expected for keys >= 50 in addition to the total count testing? Similar for all test cases.

asimarslan: This is not map predicate test. Instead it just check the returned predicate also run on map.

PR #104, opened 2017-03-08
Paging predicate implementation

asimarslan:


samshanmukh:
Hello,

I am not really able understand this previous email for which I am
replying, As it does not contain much of detailed information at all and
after after opening those links which is present in that mail I could see
some useful coding of hazelcast in the site done by csharp language. But I
would be delightful only if I can get any informatic or descriptive type of
intsructed mail for those of the programmes.

Regards
Shanmukh

asimarslan:
verify

Hazelcast.Test/Hazelcast.Client.Test/PagingPrediateTest.cs
```diff
@@ -0,0 +1,171 @@
...
+using Hazelcast.Client.Model;
+using Hazelcast.Config;
+using Hazelcast.Core;
+using Hazelcast.Util;
+using NUnit.Framework;
+
+namespace Hazelcast.Client.Test
+{
+    [TestFixture]
+    public class PagingPrediateTest : SingleMemberBaseTest
```
ihsandemir: typo: PagingPrediateTest -> PagingPredicateTest

Hazelcast.Test/Hazelcast.Client.Test/PagingPrediateTest.cs
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        [Test]
+        public virtual void TestValuePaging_with_comparator()
+        {
+            FillMap2();
+            var predicate = new PagingPredicate(3, null, new CustomComparator(2, IterationType.Value));
+
+            var page1 = map.Values(predicate);
+            Assert.That(new[] {"A", "BB", "CCC"}, Is.EquivalentTo(page1.ToArray()));
```
ihsandemir: you can also test nextPage call with custom comparator.

Hazelcast.Test/Hazelcast.Client.Test/PagingPrediateTest.cs
```diff
@@ -0,0 +1,171 @@
...
+                .AddDataSerializableFactory(IdentifiedFactory.FactoryId, new IdentifiedFactory());
+        }
+
+        internal static IMap<object, object> map;
+
+        private void FillMap()
+        {
+            for (var i = 0; i < 10; i++)
+            {
+                map.Put("key-" + i, "value-" + i);
```
ihsandemir: May be we can also put an entry with different key but the same value, e.g. : ("key-30", "value-2") . values test may fail in this test due to this issue: https://github.com/hazelcast/hazelcast/issues/9968 but it may pass if the java and csharp hashcodes are equal for the same object.

PR #113, opened 2017-07-18
invocation timeout fixed to respect long running operations

### LINKED ISSUE #112: Client does not respect long running operations
.Net client has a hard coded 250 sec of invocation timeout. This prevents submitting long running operations such as executeOnKey or complicated predicate queries. 
After the timeout hit, waiting invocation responded by TimeoutException. But the server side continue to execute and respond the actual result. On this case;

```
Received an unhandled exception from a task!: System.InvalidOperationException: An attempt was made to transition a task to a final state when it had already completed. 
at System.Threading.Tasks.Task`1.set_Result(TResult value) 
at Hazelcast.Client.Spi.SettableFuture`1.set_Result(T value) 
at Hazelcast.Client.Spi.ClientInvocationService.HandleResponseMessage(IClientMessage response) 
at Hazelcast.Client.Spi.ClientInvocationService.<>c__DisplayClass4.<HandleClientMessage>b__1() 
at System.Threading.Tasks.Task.InnerInvoke() 
at System.Threading.Tasks.Task.Execute()
```

Java client has no similar timeout and waits the server side forever. The trivial solution is to implement the same behaviour.

asimarslan:
fixes #112


asimarslan:
verify

asimarslan:
verify

asimarslan:
verify

Hazelcast.Net/Hazelcast.Client.Spi/ClientInvocationService.cs
```diff
@@ -269,10 +269,17 @@ private void HandleInvocationException(ClientInvocation invocation, Exception ex
...
                 }
                 //Fail with exception
-                if (!invocation.Future.IsComplete)
+                try
                 {
                     invocation.Future.Exception = exception;
                 }
+                catch (InvalidOperationException e)
+                {
+                    if (Logger.IsFinestEnabled())
```
ihsandemir: We made this log as warning at Java: https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java#L377  This may be also set as warning.

asimarslan:
Listener reregistration is broken, a separate fix will be sent. I'll ignore the failing test for now.

Hazelcast.Net/Hazelcast.Client.Spi/ClientInvocationService.cs
asimarslan: Maybe we should change Java side to debug 🤔 . 
There is nothing for customer can do with this log as it is only a debug message that should be reported back to us.

asimarslan:
verify

asimarslan:
verify

asimarslan:
verify

PR #115, opened 2017-07-24
ByteArrayObjectDataOutput mem leak

ampeltsi:
ByteArrayObjectDataOutput were kept per thread (on ThreadLocalOutputCache) and this lead to
memory piling (on clients' side) after heavy usage
(mulitple threads / tasks using Hazelcast client)

for the time being, and as far as I tested it it seems to be working properly.


asimarslan:
Hi @ampeltsi,

Thanks for finding this bug and submitting the fix. I have created the [issue](https://github.com/hazelcast/hazelcast-csharp-client/issues/126) for this problem. For the solution of the leak we don't want to remove the pooling mechanism. 

Please refer to [this PR](https://github.com/hazelcast/hazelcast-csharp-client/pull/125) for the fix.

PR #117, opened 2017-09-06
partition predicate implementation

asimarslan:


Hazelcast.Test/Hazelcast.Client.Test/PartitionPredicateTest.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                for (var k = 0; k < ItemsPerPartition; k++)
+                {
+                    map.Put(GenerateKeyForPartition(Client, p), p);
+                }
+            }
+            
+            partitionKey = TestSupport.RandomString();
+            partitionId = partitionService.GetPartitionId(partitionKey);
+            predicate = new PartitionPredicate(partitionKey, Predicates.True());
```
ihsandemir: Can we also have a test where Predicate is different from TruePredicate?

asimarslan: the target predicate is irrelevant here from .net client perspective. It's only extracted and executed on the member partition.

PR #121, opened 2017-10-02
fix wrong version send to server on auth

### LINKED ISSUE #111: Cluster member shows wrong version when the client is connected
When a Hazelcast .NET 3.8.1 client connected to a cluster, members reports the connectivity as follows:

```
INFO: [192.168.56.1]:5701 [dev] [3.8.1] Received auth from Connection[id=1, /127.0.0.1:5701->/127.0.0.1:11255, endpoint=null, alive=true, type=CSHARP_CLIENT], successfully authenticated, principal : ClientPrincipal{uuid='81d52973-8480-441d-a2b2-c0a8aeb49fe6', ownerUuid='455b9a86-453b-480c-acaa-025e54fab2f8'}, owner connection : true, client version : 3.8.0
```

As it is seen, the client version seems as 3.8.0 which is 3.8.1 in real.

asimarslan:
fixes #111

### DIFF
--- a/Hazelcast.Net/Hazelcast.Util/EnvironmentUtil.cs
+++ b/Hazelcast.Net/Hazelcast.Util/EnvironmentUtil.cs
@@ -42,7 +42,8 @@ namespace Hazelcast.Util
         public static string GetDllVersion()
         {
             var version = typeof(EnvironmentUtil).Assembly.GetName().Version;
-            var versionStr = string.Format("{0}.{1}.{2}", version.Major, version.Minor, version.Revision);
+            var versionStr = version.Build == 0 ? string.Format("{0}.{1}", version.Major, version.Minor)
+                : string.Format("{0}.{1}.{2}", version.Major, version.Minor, version.Build);
             return versionStr;
         }
     }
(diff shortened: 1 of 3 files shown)


Hazelcast.Test/Hazelcast.Client.Test/EnvironmentUtilTest.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace Hazelcast.Client.Test
+{
+    public class EnvironmentUtilTest
+    {
+        [Test]
+        public void TestDLLVersion()
+        {
+            var version = typeof(EnvironmentUtil).Assembly.GetName().Version.ToString();
+            Assert.True(version.StartsWith(EnvironmentUtil.GetDllVersion()));
+            Assert.False(EnvironmentUtil.GetDllVersion().EndsWith(".0"));
```
mustafaiman: I did not understand why client version has to always end with ".0". Is it not the patch version at the end of the string?

Hazelcast.Net/Hazelcast.Util/EnvironmentUtil.cs
```diff
@@ -42,7 +42,8 @@ internal class EnvironmentUtil
         public static string GetDllVersion()
         {
             var version = typeof(EnvironmentUtil).Assembly.GetName().Version;
-            var versionStr = string.Format("{0}.{1}.{2}", version.Major, version.Minor, version.Revision);
```
mustafaiman: Has `version.Revision` been renamed to `version.Build` or was it something completely different?

asimarslan: A DLL version is composed of four parts : Major.Minor.Build.Revision. We only use first three of them.  I just used the wrong one previously and now fixed

PR #124, opened 2017-10-10
Disabling socket linger option is not working. 

asimarslan:
`SocketOptions.GetLingerSeconds()` is 3 seconds by default. Socket lingering can be disabled by entering zero value:

`SocketOptions.SetLingerSeconds(0)` -> disables socket lingering.

### DIFF
--- a/Hazelcast.Net/Hazelcast.Client.Connection/ClientConnection.cs
+++ b/Hazelcast.Net/Hazelcast.Client.Connection/ClientConnection.cs
@@ -77,15 +77,17 @@ namespace Hazelcast.Client.Connection
             {
                 _clientSocket = new Socket(isa.AddressFamily, SocketType.Stream, ProtocolType.Tcp);
                 _clientSocket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.KeepAlive, true);
-
-                var lingerOption = new LingerOption(true, 5);
                 if (socketOptions.GetLingerSeconds() > 0)
                 {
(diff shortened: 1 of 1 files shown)


Hazelcast.Net/Hazelcast.Client.Connection/ClientConnection.cs
```diff
@@ -77,15 +77,16 @@ internal sealed class ClientConnection
...
-                var lingerOption = new LingerOption(true, 5);
+                var lingerOption = new LingerOption(true, socketOptions.GetLingerSeconds());
                 if (socketOptions.GetLingerSeconds() > 0)
                 {
-                    lingerOption.LingerTime = socketOptions.GetLingerSeconds();
+                    _clientSocket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.Linger, lingerOption);
+                }
+                else
+                {
+                    _clientSocket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.DontLinger, lingerOption);
```
ihsandemir: you may consider changing with _clientSocket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.DontLinger, new LingerOption(true, 0)); Currently I understand that the lingerOption will be constructed with negative linger seconds, which most probably have the same affect as described here https://msdn.microsoft.com/en-us/library/system.net.sockets.lingeroption.lingeroption(v=vs.110).aspx but it will be more clear this way.

PR #125, opened 2017-10-13
Memory and Handle leak fixes

asimarslan:
* Serialization service thread local buffer pool reimplemented to use Weak References to Threads.
* StripedTaskScheduler does not dispose internal thread pool


Hazelcast.Net/Hazelcast.IO.Serialization/SerializationService.cs
```diff
@@ -37,7 +37,8 @@ internal class SerializationService : ISerializationService
         private readonly Dictionary<Type, ISerializerAdapter> _constantTypesMap =
             new Dictionary<Type, ISerializerAdapter>(ConstantSerializersSize);
 
-        private readonly ThreadLocalOutputCache _dataOutputQueue;
+//        private readonly ThreadLocalOutputCache _dataOutputQueue;
```
sancar: leftover ?

Hazelcast.Net/Hazelcast.IO.Serialization/BufferPool.cs
```diff
@@ -0,0 +1,134 @@
...
+                _strongReferences.Add(Thread.CurrentThread, bufferPool);
+                _threadLocal.Value = new WeakReference(bufferPool);
+                return bufferPool;
+            }
+            else
+            {
+                var bufferPool = _threadLocal.Value.Target as BufferPool;
+                if (bufferPool == null)
+                {
+                    throw new HazelcastInstanceNotActiveException();
```
sancar: In Java client, we are using `HazelcastClientNotActiveException` to differentiate this with the server exception. 
We should add same exception to c# client some time if not in this prd.

asimarslan: Yes we should discuss and do it on a separate PR.

Hazelcast.Net/Hazelcast.IO.Serialization/BufferPool.cs
```diff
@@ -0,0 +1,134 @@
...
+    internal class BufferPoolThreadLocal : IDisposable
+    {
+        private readonly ISerializationService _serializationService;
+        private readonly ThreadLocal<WeakReference> _threadLocal = new ThreadLocal<WeakReference>();
+        private ConditionalWeakTable<Thread, BufferPool> _strongReferences;
+
+        public BufferPoolThreadLocal(ISerializationService serializationService)
+        {
+            _serializationService = serializationService;
+            _strongReferences = new ConditionalWeakTable<Thread, BufferPool>();
```
ihsandemir: should we start the name of the variable as strong or weak?

Hazelcast.Net/Hazelcast.IO.Serialization/BufferPool.cs
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        public BufferPool Get()
+        {
```
ihsandemir: since dispose sets the _strongReferences to null, should we do a null check here and create it if null.

Hazelcast.Net/Hazelcast.IO.Serialization/BufferPool.cs
asimarslan: the purpose of this Map is to keep strong references of BufferPools. I used the same names as on Java client implementation.

Hazelcast.Net/Hazelcast.IO.Serialization/BufferPool.cs
asimarslan: Dispose is not recoverable operation, once it is disposed, client will always shutdown and no need to recreate it.

ihsandemir: Java client implementation looks recoverable.

Hazelcast.Test/SimpleMapTestFromClient.cs
```diff
@@ -146,8 +147,8 @@ public static void StatDisplayTask()
 
         private static void Main(string[] args)
         {
+            Environment.SetEnvironmentVariable("hazelcast.logging.level", "Info");
```
ihsandemir: Was this change supposed to be in this PR?

asimarslan: yes, a minor cleanup

Hazelcast.Net/Hazelcast.IO.Serialization/BufferPool.cs
asimarslan: This class itself is recoverable but it's not used as a recoverable class.  Its clear method only called from a shutdown.
Java client uses a method named `clear` to clear it, where I named it as `Dispose` and implement interface `IDisposable` to imply that this class  is not recoverable.

PR #137, opened 2017-11-15
map and multimap listener improvements, DataAwareEventEntry added

### LINKED ISSUE #135: Map types listeners missing optimizations
Map type listeners are not inline with Java client implementation improvements.

* Events should be registered to sub listeners to not receive unnecessary events.
* Event objects are not lazy
* Event object cannot provide server merged value to client

asimarslan:
fixes #135

### DIFF
--- a/Hazelcast.Net/Hazelcast.Client.Proxy/ClientMapProxy.cs
+++ b/Hazelcast.Net/Hazelcast.Client.Proxy/ClientMapProxy.cs
@@ -56,7 +56,8 @@ namespace Hazelcast.Client.Proxy
         {
             ValidationUtil.IsNotNull(aggregator, "aggregator");
             ValidationUtil.IsNotNull(predicate, "predicate");
-            var request = MapAggregateWithPredicateCodec.EncodeRequest(GetName(), ToData(aggregator), ToData(predicate));
+            var request =
+                MapAggregateWithPredicateCodec.EncodeRequest(GetName(), ToData(aggregator), ToData(predicate));
             var response = Invoke(request);
             var resultParameters = MapAggregateWithPredicateCodec.DecodeResponse(response);
             return ToObject<TResult>(resultParameters.response);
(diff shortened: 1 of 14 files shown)


asimarslan:
verify

asimarslan:
verify

asimarslan:
verify

Hazelcast.Net/Hazelcast.Core/EntryEvent.cs
```diff
@@ -57,22 +108,24 @@ public virtual TValue GetOldValue()
...
+        public virtual TValue GetMergingValue()
+        {
+            return _mergingValue;
+        }
+
         public override string ToString()
         {
-            return "EntryEvent {" + GetSource() + "} key=" + GetKey() + ", oldValue=" + GetOldValue() + ", value=" +
-                   GetValue() + ", event=" + GetEventType() + ", by " + Member;
+            return string.Format("EntryEvent{ {0}, key={1}, oldValue={2}, value={3}, mergingValue={4} }",
```
sancar: order of oldValue and value seems mixed ?

Hazelcast.Net/Hazelcast.Client.Proxy/ClientMapProxy.cs
```diff
@@ -563,8 +631,7 @@ public ISet<TKey> KeySet()
         {
             var request = MapKeySetCodec.EncodeRequest(GetName());
             var result = Invoke(request, m => MapKeySetCodec.DecodeResponse(m).response);
-
-            return ToSet<TKey>(result);
+            return new ReadOnlyLazySet<TKey>(result, GetContext().GetSerializationService());
         }
 
         public IDictionary<TKey, TValue> GetAll(ICollection<TKey> keys)
```
ihsandemir: Would it be desirable to return lazy dictionary?

Hazelcast.Test/Hazelcast.Client.Test/ClientMapTest.cs
```diff
@@ -592,6 +593,34 @@ public void TestListener()
...
+            map.Remove("key1");
+            map.Put("key1", "value2");
+            map.Clear();
+            map.Put("key1", "value2");
+            map.EvictAll();
+            map.Put("key2", "value2");
+            map.Evict("key2");
+            map.Put("key3", "value2", 1l, TimeUnit.Seconds);
+
+            Assert.IsTrue(listener.GetLatch(EntryEventType.Added).WaitOne(TimeSpan.FromSeconds(10)));
```
ihsandemir: You may want to put this line right after map.Put("key1", "value1"); just to make sure that the event listener update did not mistakenly update the wrong latch but at the end since all latches are set, even if it is updated wrong (i.e. wrong callback method), the test will pass.

Hazelcast.Net/Hazelcast.Core/EntryEvent.cs
asimarslan: done

Hazelcast.Net/Hazelcast.Util/ReadOnlyLazySet.cs
```diff
@@ -0,0 +1,144 @@
...
+
+using System.Collections;
+using System.Collections.Generic;
+using Hazelcast.IO.Serialization;
+
+namespace Hazelcast.Util
+{
+    public class ReadOnlyLazySet<T> : ISet<T>
+    {
+        private readonly IList<IData> list;
```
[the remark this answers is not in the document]
asimarslan: This lazy set is used for lazy deserialization of result return from server and they are always List. Name is not confusing . It's using an internal list for its internal data. Keeping an internal set will add an extra conversion step from set to list, which is not straightforward in c#.

Hazelcast.Net/Hazelcast.Util/ReadOnlyLazySet.cs
```diff
@@ -0,0 +1,144 @@
...
+
+        public ReadOnlyLazySet(IList<IData> list, ISerializationService serializationService)
+        {
+            this.list = list;
+            this.serializationService = serializationService;
+        }
+
+        public IEnumerator<T> GetEnumerator()
+        {
+            return new ReadOnlyLazyList<T>.Enumerator(list, serializationService);
```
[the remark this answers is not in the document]
asimarslan: There is no such thing as List enumerator. C# only have enumerator similar to iterator in java.
List and set can share the implementation. As this lazy set uses internal list, it is just used for iteration.

Hazelcast.Test/Hazelcast.Client.Test/ClientMapTest.cs
asimarslan: make sense, I'll fix it

Hazelcast.Net/Hazelcast.Client.Proxy/ClientMapProxy.cs
asimarslan: Yes but lazy dictionary is not so easy to implement. We may improve it on its own PR.

PR #141, opened 2017-11-16
Transactions should send the invocations to correct member for unisocket clients

### LINKED ISSUE #140: Client Transactions does not respect unisocket behavior
Client transactions operates on random connections which override unisocket behaviour.

```cs
        [Test]
        public void TestListenerWithNonSmartRouting()
        {
            var cm  = ((HazelcastClientProxy) _client).GetClient().GetConnectionManager();
            for (int i = 0; i < 100; i++)
            {
                var options = new TransactionOptions().SetTransactionType(TransactionOptions.TransactionType.TwoPhase);
                var context = _client.NewTransactionContext(options);

                context.BeginTransaction();
                var tmap = context.GetMap<int, string>("test");
                var tmap2 = context.GetMap<int, string>("test2");

                tmap.Set(i, "value");
                tmap2.Set(i, "value");
                context.CommitTransaction();
                Assert.AreEqual(1, cm.ActiveConnections.Count);
            }
        }

```

asimarslan:
fixes #140

### DIFF
--- a/Hazelcast.Examples/Transactions/TransactionBenchmark.cs
+++ b/Hazelcast.Examples/Transactions/TransactionBenchmark.cs
@@ -1,0 +1,64 @@
+﻿// Copyright (c) 2008-2017, Hazelcast, Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+// http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using Hazelcast.Client;
+using Hazelcast.Config;
+using Hazelcast.Transaction;
+
+namespace Hazelcast.Examples.Transactions
+{
+    internal class TransactionBenchmark
+    {
+        public static void Run(string[] args)
+        {
+            Environment.SetEnvironmentVariable("hazelcast.logging.level", "info");
+            Environment.SetEnvironmentVariable("hazelcast.logging.type", "console");
+
+            var config = new ClientConfig();
+            config.GetGroupConfig().SetName("dev").SetPassword("dev-pass");
+            config.GetNetworkConfig().AddAddress("127.0.0.1:5701")
+                .SetSmartRouting(false);
+
+            var client = HazelcastClient.NewHazelcastClient(config);
+            var map = client.GetMap<int, string>("test");
+            var map2 = client.GetMap<int, string>("test3");
+            map.Clear();
+            map2.Clear();
+
+            long totalDuration = 0;
+            int count = 100;
+            for (int i = 0; i < count; i++)
+            {
+                TransactionOptions options = new TransactionOptions()
+                    .SetTransactionType(TransactionOptions.TransactionType.TwoPhase);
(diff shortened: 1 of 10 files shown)


Hazelcast.Examples/Transactions/TransactionBenchmark.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        public static void Run(string[] args)
+        {
+            Environment.SetEnvironmentVariable("hazelcast.logging.level", "info");
+            Environment.SetEnvironmentVariable("hazelcast.logging.type", "console");
+
+            var config = new ClientConfig();
+            config.GetGroupConfig().SetName("dev").SetPassword("dev-pass");
+            config.GetNetworkConfig().AddAddress("127.0.0.1:5701")
+                .SetSmartRouting(false);
```
mustafaiman: Why do we set the benchmark to use unisocket client? Somebody may try this sample and assume that this is as fast as it gets.

asimarslan: Transactions use single connection so it does not scale I guess but I can update the name or add a comment

Hazelcast.Net/Hazelcast.Client.Spi/ClientListenerService.cs
```diff
@@ -84,7 +84,7 @@ public ClientListenerService(HazelcastClient client)
         public string RegisterListener(IClientMessage registrationMessage, DecodeRegistrationResponse responseDecoder,
             EncodeDeregisterListenerRequest encodeDeregisterListenerRequest, DistributedEventHandler eventHandler)
         {
-            Debug.Assert(!Thread.CurrentThread.Name.Contains("eventRegistration"));
+            Debug.Assert(Thread.CurrentThread.Name == null || !Thread.CurrentThread.Name.Contains("eventRegistration"));
```
sancar: Why did we need this change ? It does not look safe.

asimarslan: Default thread name in .net is null. We want to assert that the thread is not eventRegistration thread. But the thread can be a user thread which its name by default null.

Hazelcast.Net/Hazelcast.Client.Proxy/TransactionContextProxy.cs
```diff
@@ -22,48 +22,49 @@ namespace Hazelcast.Client.Proxy
...
 
         public TransactionContextProxy(HazelcastClient client, TransactionOptions options)
         {
-            Client = client;
-
+            _client = client;
             var clusterService = (ClientClusterService) client.GetClientClusterService();
-            TxnOwnerNode = clusterService.GetRandomMember();
+
+            TxnOwnerNode = client.GetClientConfig().GetNetworkConfig().IsSmartRouting() ?
```
ihsandemir: Java uses connection for the transaction but .Net seems to use member.  Which one is preferable or more correct?

asimarslan: That's a discussion beyond this PR :)

ihsandemir: This will be a behaviour difference between Java and .Net. Java destroys the transactions when connection is closed but .Net does InvokeOnMember for the txn requests and if a new connection to the member, it will continue.

asimarslan: On server side if connection is closed it also removes all transactions. Any new connection cannot recover these transactions so .net client will not continue but instead it will receive a ` TransactionException("No transaction context found for txnId:" + txnId)` from server.

Although there is a difference in handling the closed connection,  there isn't any risk for transactional behaviour.

PR #155, opened 2018-02-15
client statistics implementation

asimarslan:


asimarslan:
run-net

Hazelcast.Net/Hazelcast.Client/Statistics.cs
```diff
@@ -0,0 +1,296 @@
...
+                () => ((long) (DateTime.Now - Process.GetCurrentProcess().StartTime).TotalMilliseconds).ToString());
+            _gauges.Add("runtime.usedMemory", () => Process.GetCurrentProcess().WorkingSet64.ToString());
+            _gauges.Add("executionService.userExecutorQueueSize", () => EmptyStatValue);
+        }
+
+        private void FillMetrics(StringBuilder stats, ClientConnection ownerConnection)
+        {
+            AddStat(stats, "lastStatisticsCollectionTime", Clock.CurrentTimeMillis());
+            AddStat(stats, "enterprise", "false");
+            AddStat(stats, "clientType", "CSHARP");
```
mustafaiman: should this be `CSP`? See https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/ClientTypes.java#L40

asimarslan: No, it is from this Enum 
https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/core/ClientType.java#L24

Hazelcast.Net/Hazelcast.Client.Spi/ProxyManager.cs
```diff
@@ -63,8 +63,9 @@ public string AddDistributedObjectListener(IDistributedObjectListener listener)
             var isSmart = _client.GetClientConfig().GetNetworkConfig().IsSmartRouting();
             var request = ClientAddDistributedObjectListenerCodec.EncodeRequest(isSmart);
             var context = new ClientContext(_client.GetSerializationService(), _client.GetClientClusterService(),
-                _client.GetClientPartitionService(), _client.GetInvocationService(), _client.GetClientExecutionService(),
-                _client.GetListenerService(),
+                _client.GetClientPartitionService(), _client.GetInvocationService(),
+                _client.GetClientExecutionService(),
+                _client.GetListenerService(), _client.GetNearCacheManager(),
                 this, _client.GetClientConfig());
             //EventHandler<PortableDistributedObjectEvent> eventHandler = new _EventHandler_211(this, listener);
```
ihsandemir: why not remove this line?

Hazelcast.Net/Hazelcast.Client/Statistics.cs
```diff
@@ -0,0 +1,296 @@
...
+
+        private readonly IDictionary<string, Func<string>> _gauges = new Dictionary<string, Func<string>>();
+
+        private volatile Address _ownerAddress;
+
+        public Statistics(HazelcastClient client)
+        {
+            _client = client;
+            _enabled = EnvironmentUtil.ReadBool("hazelcast.client.statistics.enabled") ?? false;
+            _period = EnvironmentUtil.ReadInt("hazelcast.client.statistics.period.seconds") ?? 3;
```
ihsandemir: at Java we have this log, you may add it as well: https://github.com/hazelcast/hazelcast/blob/v3.9.2/hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java#L134

We also do it at start method. If not enabled, no need to print the message.

Hazelcast.Net/Hazelcast.Client/Statistics.cs
```diff
@@ -0,0 +1,296 @@
...
+            _gauges.Add("os.totalSwapSpaceSize", () => EmptyStatValue);
+
+            _gauges.Add("runtime.availableProcessors", () => Environment.ProcessorCount.ToString());
+            _gauges.Add("runtime.freeMemory", () => EmptyStatValue);
+            _gauges.Add("runtime.maxMemory", () => Process.GetCurrentProcess().MaxWorkingSet.ToInt64().ToString());
+            _gauges.Add("runtime.totalMemory", () => Process.GetCurrentProcess().WorkingSet64.ToString());
+            _gauges.Add("runtime.uptime",
+                () => ((long) (DateTime.Now - Process.GetCurrentProcess().StartTime).TotalMilliseconds).ToString());
+            _gauges.Add("runtime.usedMemory", () => Process.GetCurrentProcess().WorkingSet64.ToString());
+            _gauges.Add("executionService.userExecutorQueueSize", () => EmptyStatValue);
```
ihsandemir: do we not have user executor queue for .net?

Hazelcast.Net/Hazelcast.Client/Statistics.cs
```diff
@@ -0,0 +1,296 @@
...
+            AddStat(stats, "lastStatisticsCollectionTime", Clock.CurrentTimeMillis());
+            AddStat(stats, "enterprise", "false");
+            AddStat(stats, "clientType", "CSHARP");
+            AddStat(stats, "clientVersion", DllVersion);
+            AddStat(stats, "clusterConnectionTimestamp", ownerConnection.ConnectionStartTime);
+            AddStat(stats, "clientAddress", ownerConnection.GetLocalSocketAddress());
+            AddStat(stats, "clientName", _client.GetName());
+
+            var credentials = _client.GetClientConfig().GetCredentials();
+//            if (!(credentials is UsernamePasswordCredentials))
```
ihsandemir: why is this commented out?

Hazelcast.Net/Hazelcast.Client/Statistics.cs
```diff
@@ -0,0 +1,296 @@
...
+            _gauges.Add("runtime.usedMemory", () => Process.GetCurrentProcess().WorkingSet64.ToString());
+            _gauges.Add("executionService.userExecutorQueueSize", () => EmptyStatValue);
+        }
+
+        private void FillMetrics(StringBuilder stats, ClientConnection ownerConnection)
+        {
+            AddStat(stats, "lastStatisticsCollectionTime", Clock.CurrentTimeMillis());
+            AddStat(stats, "enterprise", "false");
+            AddStat(stats, "clientType", "CSHARP");
+            AddStat(stats, "clientVersion", DllVersion);
```
ihsandemir: Java client is not sending this. It may be obtained from server side. Actually the client type can also be obtained from server side endpoint.

Hazelcast.Net/Hazelcast.Net.Ext/AtomicInteger.cs
```diff
@@ -75,12 +75,12 @@ public int IncrementAndGet()
 
         public void Set(int newValue)
         {
-            _val = newValue;
+            Interlocked.Exchange(ref _val, newValue);
```
ihsandemir: What is the advantage of using Interlocked here? I assume the .Net int operations are atomic by default.

Hazelcast.Net/Hazelcast.Client/Statistics.cs
```diff
@@ -0,0 +1,296 @@
...
+                // suppress exception, do not print too many messages
+                if (Logger.IsFinestEnabled())
+                {
+                    Logger.Finest("Could not send stats ", e);
+                }
+            }
+        }
+
+        private void RegisterMetrics()
+        {
```
ihsandemir: Do you think that we should test for actual runtime functioning of the gauges and not register them if the method is throwing an exception?

Hazelcast.Net/Hazelcast.Client/Statistics.cs
asimarslan: that's still missing

Hazelcast.Net/Hazelcast.Client/Statistics.cs
asimarslan: I thinks we should always send principal. I'll delete the commented code

Hazelcast.Net/Hazelcast.Client/Statistics.cs
asimarslan: Client will sent its version . Server side shouldn't add data to it.

Hazelcast.Net/Hazelcast.Net.Ext/AtomicInteger.cs
asimarslan: Interlocked is equivalent of Java's Atomic classes. This is a bug to be fixed to make set operation atomic

Hazelcast.Net/Hazelcast.Client/Statistics.cs
asimarslan: We have PR builders for linux and windows so that we can test on various OS. I think we should keep it.

Hazelcast.Net/Hazelcast.Client/Statistics.cs
```diff
@@ -0,0 +1,293 @@
...
+            _ownerAddress = ownerAddress;
+            return null;
+        }
+
+        private void PeriodicStatisticsSendTask()
+        {
+            var ownerConnection = GetOwnerConnection();
+            if (null == ownerConnection)
+            {
+                Logger.Finest("Can not send client statistics to the server. No owner connection.");
```
mustafaiman: `GetOwnerConnection` method does not print repetitive logs if the owner does not support statistics. However, this method prints repetitive logs when server does not support statisticks, right? It kind of reverses what `GetOwnerConnection` does about logging.

Also, do you need `Logger.isFinestEnabled` check here like in `GetOwnerConnection`?

Hazelcast.Net/Hazelcast.Client/Statistics.cs
mustafaiman: Isn't this redundant though? Clients send their type and version in AuthenticationTask.

Hazelcast.Net/Hazelcast.Client/Statistics.cs
ihsandemir: What I meant by word "test" was this:
```
try {
   tryGetStat();
  registerSinceNoException();
} catch () {
 // do not register
}
```
And do the above code at production code, that is not the test code.

Hazelcast.Net/Hazelcast.Client/Statistics.cs
asimarslan: I've added statistic enabled info log

Hazelcast.Net/Hazelcast.Client/Statistics.cs
asimarslan: As rule of thumb; client sends its all stat data and it's not gathered from multiple sources.

Hazelcast.Net/Hazelcast.Client/Statistics.cs
asimarslan: Added controlled registration step

Hazelcast.Net/Hazelcast.Client/Statistics.cs
ihsandemir: should we have a PR for Java for this additional field, so that Java is always our reference implementation?

PR #159, opened 2018-02-27
fix client proxy manager

### LINKED ISSUE #144: ProxyManager is broken
ProxyManager is responsible of 

* creating proxy objects and persist them to provide single instance
* DistributedObjectListener event delivery for these proxies

C# generics are different than Java where you cannot ignore generic types at runtime.
For that reason, a proxy created with generic parameters cannot be cast to different generic parameters at run-time.  We need to replace these proxies with non-generic versions of them internally.

There is also a race condition that cannot handle concurrent proxy creation.

> Original bug report:
> 
> Seeing alot of WARNING level log on concurrent :
> ```
> [WARNING] - [:19] Hazelcast.Client.Spi.ProxyManager: Replacing old proxy for default of type Hazelcast.Client.Proxy.ClientMapProxy`2[System.String,System.String] with Hazelcast.Client.Proxy.ClientMapProxy`2[System.String,System.String]
> ```

asimarslan:
fixes #144


asimarslan:
run-linux

Hazelcast.Net/Hazelcast.Client.Spi/ProxyManager.cs
```diff
@@ -106,105 +104,137 @@ public void Destroy()
...
+                }
+            }
+
+            if (isAssignable)
+            {
+                _proxies.TryRemove(objectInfo, out clientProxy);
+                return GetOrCreateProxy<T>(service, id);
+            }
+
+            throw new InvalidCastException(string.Format("Distributed object is already created with {0}", proxyArgs));
```
mustafaiman: This gives the impression that the problem is to re-create an existing proxy whereas the problem is the incompatibility of the types. I would say `...already created with incompatible types {0}`

asimarslan:
run-net

PR #160, opened 2018-03-05
client statistics code sample

asimarslan:


Hazelcast.Examples/Client/ClientStatisticsExample.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    internal class ClientStatisticsExample
+    {
+        private static void Run(string[] args)
+        {
+            //This will enable client statistics 
+            Environment.SetEnvironmentVariable("hazelcast.client.statistics.enabled", "true");
+            
+            //set the statistics send period, default value is 3 seconds
+            Environment.SetEnvironmentVariable("hazelcast.client.statistics.period.seconds", "3");
```
ihsandemir: you may use a non-default value for the example.

Hazelcast.Examples/Client/ClientStatisticsExample.cs
```diff
@@ -0,0 +1,39 @@
...
+            
+            var config = new ClientConfig();
+            config.GetNetworkConfig().AddAddress("127.0.0.1");
```
ihsandemir: do we need this line? don't we add 127.0.0.1 by default if the list is empty?

Hazelcast.Examples/Client/ClientStatisticsExample.cs
```diff
@@ -0,0 +1,39 @@
...
+            var client = HazelcastClient.NewHazelcastClient(config);
+            //After client connected you can use Management Center to visualize client statistics
+
```
ihsandemir: you may put some sleep here to give the user some time to visualise the statistics.

Hazelcast.Examples/Client/ClientStatisticsExample.cs
asimarslan: no we need it, there isn't any default

Hazelcast.Examples/Client/ClientStatisticsExample.cs
asimarslan: Let me put some client activity instead

Hazelcast.Examples/Client/ClientStatisticsExample.cs
ihsandemir: this behaviour seems to differ from Java client.

PR #183, opened 2018-09-17
Fixes interval/heartbeat mixup in ClientConnectionManager

pveentjer:
The values for timeout and interval are mixed up.


ihsandemir:
@pveentjer  Unit test?

pveentjer:
Since the error is so obvious, I don't see the added value of a unit test. Apart from that I have no clue how to write one for .NET or how to trigger the behaviour.

PR #184, opened 2018-09-17
Refactor Heartbeat timer to prevent GC 

asimarslan:
Refactor Heartbeat timer to prevent GC  collecting it on new version's of .net framework.

see: https://stackoverflow.com/questions/4962172/why-does-a-system-timers-timer-survive-gc-but-not-system-threading-timer


asimarslan:
run-net

PR #192, opened 2018-09-27
Update documentation for Get(object key) to take structs into account.

drieseng:
The documentation for `IMap<TKey,TValue>.Get(object key)` states that it will return null if the map does not contain this key.

When **TValue** is a struct, this is not correct. I've updated the doc so it applies to both reference and non-reference types.

### DIFF
--- a/Hazelcast.Net/Hazelcast.Core/IMap.cs
+++ b/Hazelcast.Net/Hazelcast.Core/IMap.cs
@@ -490,10 +490,12 @@ namespace Hazelcast.Core
         void ForceUnlock(TKey key);
 
         /// <summary>
-        ///     Returns the value for the specified key, or <c>null</c> if this map does not contain this key.
+        ///     Returns the value for the specified key, or the default value for <typeparamref name="TValue"/>
+        ///     if this map does not contain this key.
         /// </summary>
         /// <remarks>
-        ///     Returns the value for the specified key, or <c>null</c> if this map does not contain this key.
+        ///     Returns the value for the specified key, or the default value for <typeparamref name="TValue"/>
+        ///     if this map does not contain this key.
         ///     <p>
         ///         <b>Warning:</b>
         ///     </p>


asimarslan:
Hi @drieseng,
Thanks for the PR.  
Can you please sign and send the contributor agreement on https://hazelcast.atlassian.net/wiki/spaces/COM/pages/6357071/Hazelcast+Contributor+Agreement
in order to merge this PR.

Hazelcast.Net/Hazelcast.Core/IMap.cs
```diff
@@ -490,10 +490,12 @@ public interface IMap<TKey, TValue> : IDistributedObject
         void ForceUnlock(TKey key);
 
         /// <summary>
-        ///     Returns the value for the specified key, or <c>null</c> if this map does not contain this key.
+        ///     Returns the value for the specified key, or the default value for <typeparamref name="TValue"/>
+        ///     if this map does not contain this key.
         /// </summary>
         /// <remarks>
-        ///     Returns the value for the specified key, or <c>null</c> if this map does not contain this key.
+        ///     Returns the value for the specified key, or the default value for <typeparamref name="TValue"/>
```
ihsandemir: does it not return null if not exist?

asimarslan:
verify

zpqrtbnk:
Hey. Stephan here. We are about to release version 4.0 of the Hazelcast.Net client, and our README and documentation have completely been overhauled, making this PR... unfortunately irrelevant. Sorry for not merging earlier, there has been a period where we were not paying sufficient attention to PRs, and we want to be better at it in the future. I want to thank you for your contribution, and hope to see you again here!

As for the purpose of the PR, the documentation in the new client has been updated to read:

```
/// <summary>
/// Gets the value for the specified key.
/// </summary>
/// <param name="key">The key.</param>
/// <returns>The value for the specified key, if any; otherwise <c>default(TValue)</c>.</returns>
/// <remarks>
/// <para>This methods <strong>interacts with the server-side <c>MapStore</c></strong>.
/// If no value for the specified key is found in memory, <c>MapLoader.load(...)</c> is invoked
/// to try to load the value from the <c>MapStore</c> backing the map, if any.</para>
/// </remarks>
Task<TValue> GetAsync(TKey key);
```

which indeed mentions `default(TValue)` instead of `null`.

Issue #187, opened 2018-09-20
After owner node is killed, client still tries to connect to it even new owner is selected


sertugkaya:
Client version 3.9.1
Configurations are default, except adding 3 member's IP'S.
Client continuously does put/get operations to the cluster and owner member is brought down.
After this member change, client stops reading data (read operations does not return) from remaining 2 members even if there's connection to both. Connection is also detected by monitoring the network on the hosts.

Also, as seen in the [client logs](https://github.com/hazelcast/hazelcast-csharp-client/files/2403052/client_log_when_ownernode_down.txt), owner member with ip .89 is killed a couple mins after client connects to the cluster. After that it continuously tries to connect .89.

sertugkaya:
Attached is newer logs from client and some of the members, tests done with .Net 3.9.3 version. 
Same scenario is tested (killing an owner node) and same results are observed.
[Owner Node Restarted.zip](https://github.com/hazelcast/hazelcast-csharp-client/files/2445569/Case.1.-.Owner.Node.Restarted.zip)

PR #197, opened 2018-10-16
fixes owner restart causes client operations hang problem

asimarslan:
fixes issue #187

### DIFF
--- a/Hazelcast.Net/Hazelcast.Client.Spi/ClientExecutionService.cs
+++ b/Hazelcast.Net/Hazelcast.Client.Spi/ClientExecutionService.cs
@@ -16,6 +16,7 @@ using System;
 using System.Threading;
 using System.Threading.Tasks;
 using Hazelcast.Core;
+using Hazelcast.Net.Ext;
 using Hazelcast.Util;
 
 #pragma warning disable CS1591
@@ -24,6 +25,7 @@ using Hazelcast.Util;
     internal sealed class ClientExecutionService : IClientExecutionService
     {
         private readonly TaskFactory _taskFactory = Task.Factory;
+        private readonly AtomicBoolean _live = new AtomicBoolean(true);
 
         public ClientExecutionService(string name, int poolSize)
         {
@@ -41,6 +43,10 @@ using Hazelcast.Util;
 
         public void ScheduleWithFixedDelay(Action command, long initialDelay, long period, TimeUnit unit, CancellationToken token)
         {
+            if (!_live.Get())
+            {
+                throw new HazelcastException("Client is shut down.");
+            }
             ScheduleWithCancellation(command, initialDelay, unit, token).ContinueWith(task =>
             {
                 if (!task.IsCanceled)
@@ -52,6 +58,10 @@ using Hazelcast.Util;
 
         public Task ScheduleWithCancellation(Action command, long delay, TimeUnit unit, CancellationToken token)
         {
+            if (!_live.Get())
+            {
+                throw new HazelcastException("Client is shut down.");
+            }
             var tcs = new TaskCompletionSource<object>();
             var timer = new Timer(o =>
             {
@@ -79,6 +89,10 @@ using Hazelcast.Util;
 
         public Task Schedule(Action command, long delay, TimeUnit unit)
         {
+            if (!_live.Get())
+            {
+                throw new HazelcastException("Client is shut down.");
+            }
             var tcs = new TaskCompletionSource<object>();
             var timer = new Timer(o =>
             {
@@ -96,6 +110,7 @@ using Hazelcast.Util;
 
         public void Shutdown()
         {
+            _live.Set(false);
         }
 
         internal Task SubmitInternal(Action action)
--- a/Hazelcast.Net/Hazelcast.Client.Spi/ClientPartitionService.cs
+++ b/Hazelcast.Net/Hazelcast.Client.Spi/ClientPartitionService.cs
@@ -28,7 +28,7 @@ using Hazelcast.Util;
  namespace Hazelcast.Client.Spi
 {
     internal sealed class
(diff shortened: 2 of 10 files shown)


Hazelcast.Test/Hazelcast.Client.Test/ClientRetryTest.cs
```diff
@@ -177,5 +178,54 @@ public void TestRetryTimeout()
...
+                    }
+                }
+            });
+            thread.Start();
+
+            Thread.Sleep(10000);
+
+            _remoteController.shutdownMember(_cluster.Id, member.Uuid);
+
+            Thread.Sleep(10000);
```
ihsandemir: why are we sleeping 10 secs here?

Hazelcast.Test/Hazelcast.Client.Test/HazelcastTestSupport.cs
```diff
@@ -39,8 +39,13 @@ public class HazelcastTestSupport
 
         public HazelcastTestSupport()
         {
-            Environment.SetEnvironmentVariable("hazelcast.logging.level", "finest");
+#if DEBUG
```
ihsandemir: was this for local testing purposes? do we have DEBUG set all the time?

Hazelcast.Net/Hazelcast.Client.Spi/ClientPartitionService.cs
```diff
@@ -48,9 +48,22 @@ public ClientPartitionService(HazelcastClient client)
 
         public Address GetPartitionOwner(int partitionId)
         {
-            Address rtn;
-            _partitions.TryGetValue(partitionId, out rtn);
-            return rtn;
+            Address partitionOwner = null;
+            while (_live.Get() && !_partitions.TryGetValue(partitionId, out partitionOwner))
+            {
+                Logger.Finest("Address of a partition cannot be null. Retrying to get it...");
```
ihsandemir: is the partitionOwner set correctly when the code is here? if not, should we continue?

Hazelcast.Net/Hazelcast.Client.Spi/ClientExecutionService.cs
```diff
@@ -96,6 +110,7 @@ public Task Schedule(Action command, long delay, TimeUnit unit)
 
         public void Shutdown()
         {
+            _live.Set(true);
```
sancar: this should be `false`

Hazelcast.Test/Hazelcast.Client.Test/HazelcastTestSupport.cs
asimarslan: This is removed on non Debug compiles. We only use Debug on local development, all ther case are not DEBUG.

Hazelcast.Test/Hazelcast.Client.Test/ClientRetryTest.cs
```diff
@@ -177,5 +179,58 @@ public void TestRetryTimeout()
...
+            var terminate = new ManualResetEvent(false);
+            var task = Task.Factory.StartNew(() =>
+            {
+                while (!terminate.WaitOne(0))
+                {
+                    var st = new Stopwatch();
+                    st.Start();
+                    var result = map.GetAll(dict.Keys);
+                    var took = st.ElapsedMilliseconds;
+                    if (took > 1500)
```
ihsandemir: Time sensitive testing may be problem at different environments especially when too many processes running on the same machine (e.g. parallel dockers, we observed during compatibility test runs).

Hazelcast.Net/Hazelcast.Client.Spi/ClientPartitionService.cs
ihsandemir: Printing in a busy loop may result too many messages, do we really want to do it?

Hazelcast.Net/Hazelcast.Client.Connection/ClientConnectionManager.cs
```diff
@@ -311,10 +311,6 @@ private void Authenticate(ClientConnection connection, bool isOwnerConnection)
             }
             var result = ClientAuthenticationCodec.DecodeResponse(response);
 
-            if (result.address == null)
```
sancar: Why are we removing this ? Address.decode can return null anytime. On java if we get null here we are handling it as failure. Otherwise it is resulting in an unstable intermediate state. I think, We should keep the check and the exception.

Hazelcast.Net/Hazelcast.Client.Spi/ClientPartitionService.cs
sancar: I think instead of this we can reuse the same code that we use in GetPartitions. 
```
if (_partitionCount == 0)
{
         GetPartitionsBlocking();
}
_partitions.TryGetValue(partitionId, out partitionOwner)  <=== this is guaranteed to work
```

asimarslan: GetPartitionsBlocking(); blocks for 10 sec which is too much for here. I'll just add a sleep

asimarslan:
run-net

asimarslan:
run-net

Hazelcast.Test/Hazelcast.Client.Test/ClientRetryTest.cs
asimarslan: This is a delay measurement that the theoretical max is 10 sec and we want it as fast as 100 ms or so. I'm assign a comparibly a large number. I don't know a better solution for that.

asimarslan:
verify-net

asimarslan:
run-net

asimarslan:
run-net

PR #226, opened 2019-03-11
Copyright year update to 2019

stalek71:


asimarslan:
There are missing copyright updates on non cs files

PR #228, opened 2019-03-14
[Backport]GetAsync call removed redundant internal task creation

asimarslan:
Backport of #223
the partition update wait time for the first time is updated to 100ms from 10sec.

### DIFF
--- a/Hazelcast.Net/Hazelcast.Client.Spi/ClientPartitionService.cs
+++ b/Hazelcast.Net/Hazelcast.Client.Spi/ClientPartitionService.cs
@@ -190,7 +190,7 @@ using Hazelcast.Util;
         {
             while (!GetPartitions() && _live.Get())
             {
-                Thread.Sleep(PartitionRefreshPeriod);
+                Thread.Sleep(100);
             }
         }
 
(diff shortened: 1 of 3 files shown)


Hazelcast.Net/Hazelcast.Client.Spi/ClientPartitionService.cs
```diff
@@ -190,7 +190,7 @@ private void GetPartitionsBlocking()
         {
             while (!GetPartitions() && _live.Get())
             {
-                Thread.Sleep(PartitionRefreshPeriod);
+                Thread.Sleep(100);
```
stalek71: I will change it to constant in future commits

stalek71: Could we use some meaningful constant name here instead of the value?

asimarslan: Can you do it on master 3.12 branch

stalek71: ok.

PR #230, opened 2019-03-18
PN Counter sample and doc updates

stalek71:


README.md
```diff
@@ -1554,6 +1555,24 @@ A Semaphore usage example is shown below.
     Console.WriteLine( "Number of available permits: " + semaphore.availablePermits());
 ```
 
+### 7.4.11 Using PN Counter
+
+Hazelcast `PNCounter` (Positive-Negative Counter) is a CRDT positive-negative counter implementation. It is an eventually consistent counter given there is no member failure. For details, see the [PN Counter section](https://docs.hazelcast.org/docs/latest/manual/html-single/index.html#pn-counter) in the Hazelcast IMDG Reference Manual.
+
+A PN Counter usage example has been shown below.
```
Serdaro: Maybe using simple present tense is better here? 
"is" instead of "has been"

stalek71: Replaced

stalek71:
The sample location and namespace have been changed

PR #247, opened 2019-07-16
Less bytes for strings

Scooletz:
This PR removes `byte[]` allocations in `SafeBuffer` whenever a UTF8 encoded string is being written or read. 

I've found it when reading through the `SafeBuffer` codebase. It's a simple removal, I didn't found that much improvement but it reduces coping encoded bytes.


Scooletz:
Replaced by #251

Issue #240, opened 2019-06-06
Reconnect to Hazelcast cluster when network connection was interrupted


arkadiuszdr:
Hazelcast 3.11.1 and Csharp client 3.11
We have .NET HazelcastClient which establish connection to Hazelcast cluster In some stage during application lifetime we switch off network connection. After approximately 20 seconds we got exception Hazelcast Instance is not active. Then we restore network connection but it does not solve problem. Each try to obtain any data from hazelcast results with the same exception message. 
We try to set NetworkConfig parameters: ConnectionTimeout, ConnectionAttemptLimit, ConnectionAttemptPeriod but it does not change hazelcast behaviour. We also perform the same test with Java HazelcastClient with usage of ClientConnectionStrategyConfig and with this settings HazelcastClient was able to reconnect after lost of network connection.
Is there any way to solve that problem?

jgardiner68:
The issue has been identified: when the network connection is down, the .net client is unable to resolve hostnames which generates an exception.

A workaround is to use IP addresses within the .net client config rather than hostnames but we will be issuing an updated .net client that resolves this issue.

Thank you for reporting it.

asergeev95:
@jgardiner68 could you please provide an estimation of when new version with this fix will be available? thanks

jgardiner68:
@asergeev95  the fix will be included in the next update of the .net client which is expected in the 2nd half of July

rjso:
This has been causing some issues on our side of things as well with the exact same symptoms as @arkadiuszdr. 

@jgardiner68 do you think your 2nd half of July is still valid? Would you require some help from my side?

jgardiner68:
@rjso This is on schedule for a release by the end of July.  I will update this ticket when it is available and will be great to get your feedback.

Scooletz:
Will be shipped in 3.12.1

rjso:
Thanks @jgardiner68 and @Scooletz I will test it as soon as 3.12.1 becomes available. thanks for your help :)

rjso:
Hi @jgardiner68, @Scooletz we have tried to confirm this issue has been resolved, and I can confirm it isn't, at least from my perspective.

[...]

PR #233, opened 2019-03-21
When throws "No such host is known" it stops trying to reconnect

serdarsert:
In ConnectAsOwner method, when the GetInetSocketAddress method receives the "No such host is known" error, it stops trying to connect again. 
_In this case, the network has been down for a short time_

System.Net.Sockets.SocketException (0x80004005): No such host is known
   at System.Net.Dns.GetAddrInfo(String name)
   at System.Net.Dns.InternalGetHostByName(String hostName, Boolean includeIPv6)
   at System.Net.Dns.GetHostAddresses(String hostNameOrAddress)
   at Hazelcast.IO.Address.GetAddressByName(String name)
   at Hazelcast.IO.Address.GetInetSocketAddress()
   at Hazelcast.Client.Spi.ClientClusterService.ConnectAsOwner(ICollection`1 triedAddresses, IList`1 exceptions)
   at Hazelcast.Client.Spi.ClientClusterService.ConnectToOne()
   at Hazelcast.Client.Spi.ClientClusterService.ConnectToCluster()
   at Hazelcast.Client.Spi.ClientClusterService.<ConnectionRemoved>b__27_0()

my log file:
[hazelcast-nosuchhostisknown-error.txt](https://github.com/hazelcast/hazelcast-csharp-client/files/2991062/hazelcast-nosuchhostisknown-error.txt)

Please review this change, thanks.

PR #249, opened 2019-07-18
Handling DNS resolution failures gracefully

Scooletz:
This PR addresses #240 
The fix is based on #233 

### Description

This PR introduces an additional layer for DNS probing that enables injecting faults to the resolution mechanism. With this it enables simulating DNS occasional problems.

`ConnectAsOwner` is fixed on basis of #233, by moving the `address.GetInetSocketAddress` to the inside of the `try` block.

### Remarks

I think that `GetPossibleIpAddressesFor` also is prone to throwing, but I'm not able to find the execution path that would fall into this category. Would be great if somebody could hive me a hint how to setup the configuration to make it fall. Now with DNS fault injection, it's easy when right params are known.


Scooletz:
Replaced by #250

PR #256, opened 2019-08-09
Target framework moved to net462

Scooletz:
This PR moves the whole solution to target `net462`. This including using `async-await` and `Tasks`. This includes:

- turning `async Task` all the tests that tests `async` behavior but previously were `void`
- preliminary cleaning up the tests by
  - removing `Test` prefix of every method
  - reviewing async/sync paths
  - extracting common methods
  - applying same naming (it's up to latter debate, which should be chosen)
  - moving classes to nested whenever needed
  - making visibility coherent (it's up to latter debate, which should be chosen)
  - trying to replace `new Thread` with tasks
- `ClientExecutionService` simplified a lot with async-await code
- changing some API:
  - making `IHCollection` accepting `IEnumerable<T>` instead of `ICollection<T>` for `*All` methods
  - removing an unused parameter in `IRingBuffer.AddAsync`

Some work, like adding all the async overloads will be left for future for sure.


Scooletz:
It will be massive.

Scooletz:
verify

Scooletz:
This PR will be split into several smaller ones.

PR #257, opened 2019-08-20
Spi internalized

Scooletz:
This PR internalizes API that was made public.

Requires rebase after #256 is merged


Hazelcast.Net/Hazelcast.Client.Spi/Exception.cs
```diff
@@ -41,7 +40,7 @@ public RetryableHazelcastException(string message) : base(message)
     /// A <see cref="RetryableHazelcastException"/> that indicates operation is sent to a machine that isn't member of the cluster
     /// </summary>
     [Serializable]
-    public class TargetNotMemberException : RetryableHazelcastException
+    class TargetNotMemberException : RetryableHazelcastException
```
asimarslan: This exception is defined in the protocol list so we need to keep it public.

Hazelcast.Net/Hazelcast.Client.Spi/Exception.cs
```diff
@@ -54,7 +53,7 @@ public TargetNotMemberException(string message) : base(message)
     /// whether the invocation runs or not is indeterminate
     /// </summary>
     [Serializable]
-    public class TargetDisconnectedException : RetryableHazelcastException
+    class TargetDisconnectedException : RetryableHazelcastException
```
asimarslan: This exception is defined in the protocol list so we need to keep it public.

asimarslan:
I think we can remove all SPI naming too.

zpqrtbnk:

PR #258, opened 2019-08-22
Simplified proxy creation

Scooletz:
This PR changes a little how the proxy is obtained and created, with no need of capturing the parameters in a closure

This PR should be rebased after #257


zpqrtbnk:

(we are creating proxies in an entirely different way)

PR #259, opened 2019-08-22
Disposable transactions

Scooletz:
This PR introduces a new interface for transactions, idiomatic for .NET:

- uses `IDisposable` pattern 
- naming convention similar to `IDbTransaction`
- no async API yet

```csharp
public interface ITransaction : IDisposable
{
    string Id { get; }
    void Commit();
    void Rollback();
}
```

Should me rebased/cherry-picked and merged after #256 #257 

Questions:
1. Why there's `[ThreadStatic] static bool? _threadFlag;` in `TransactionProxy` and why it's prohibited to run in from different thread? Not that I want to, but what's this check for?


zpqrtbnk:

BTW, we are moving to something akin to:
```
using (var context = client.BeginTransactionAsync())
{
  // do stuff...
  context.Complete();
}
```

PR #263, opened 2019-08-28
Moving to net45

Scooletz:
This PR moves the solution to `net45` to enable newer features of the runtime, packages and the C# language. Additionally:

- it makes `BufferPool` tests independent from the `GC.Collect` by spinning a test in a loop and checking with `GC.GetTotalMemory(true)` is the memory size breached
- it changes the implementation of the `BufferPoolThreadLocal` simpler, by switching to `ThreadLocal` only as in .NET `ThreadLocal` will not leak no matter if they are disposed or finalized (they have a proper `Dispose(bool)` implementation.
- it changes the behavior of the disposed `BufferPool`. Previously, it was possible to still obtain a buffer when the pool is disposed and the `WeakReference`'s target is not released yet. Currently, it will always throw an exception. This changes the type of exception being thrown when the client is shut down and an action requires a serialization service to be used. To preserver the exception that was being thrown before, the `SerializationService` checks whether it's active.

### DIFF
--- a/Hazelcast.Examples/Hazelcast.Examples.csproj
+++ b/Hazelcast.Examples/Hazelcast.Examples.csproj
@@ -1,12 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\shared.project.props" />
   <PropertyGroup Label="build">
-    <TargetFrameworks>net40;netcoreapp2.0</TargetFrameworks>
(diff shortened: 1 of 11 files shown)


Scooletz:
verify

Scooletz:
verify

Scooletz:
verify

Scooletz:
verify

PR #264, opened 2019-09-05
Partial test clean up

Scooletz:
This PR changes a bunch of tests files, doing the following cleanup:

- removing `Test` prefix before all the test methods (all tests were strated with `Test)
- removing of unneeded `virtual` keyword
- using higher version of language features like readonly properties etc.
- replacing a static member for an implementation of `IDistributedObject` and a static `TearDown` with regular instance based approach
- in a few places: properly order `Assert` parameters (expected, actual)
- unifying style for field, agreeing on `_` prefix


Hazelcast.Test/Hazelcast.Client.Test/AbstractLazyDictionaryTest.cs
```diff
@@ -130,10 +127,9 @@ public void Test_Remove_key()
         [Test]
         public void Test_TryGetValue()
```
asimarslan: is this `Test_` is leftover or do we keep some of them?

Hazelcast.Test/Hazelcast.Client.Test.Serialization/PortableSerializationTest.cs
```diff
@@ -166,7 +166,7 @@ public IPortable Create(int classId)
 
         //https://github.com/hazelcast/hazelcast/issues/1096
         [Test]
-        public void Test_1096_ByteArrayContentSame()
+        public void Issue_1096_ByteArrayContentSame()
```
asimarslan: Can we get rid of issue id's here? we need actual test name in here

PR #268, opened 2019-09-17
ClientExecutionService simplified with async-await

Scooletz:
This PR changes the implementation of `ClientExecutionService` from using a `Timer` bases approach to a direct async-await usage.


zpqrtbnk:

PR #276, opened 2020-03-10
Make network code more robust

zpqrtbnk:
Makes network code more robust by improving how the socket reading handles closing connections and sockets. With the current code, the socket may remain used by a pending read although the stream, etc. are all gone, and then the socket throws a `TaskCanceledException` that remains unhandled.

This happens very rarely (but it happens) on .NET Core on Windows. It happens regularly on Linux.

This PR ensure that these exceptions are properly handled. With this code in place, I have not seen unhandled exceptions on Linux.


zpqrtbnk:
won't merge, new 4.0 client code is completely different

PR #277, opened 2020-03-11
Update README for Kerberos

zpqrtbnk:
Keep this PR draft for now - we don't want to update the doc right now.

In addition, this is only the client-side documentation. The server-side documentation needs to be updated too.


zpqrtbnk:
Won't merge, been superseded by more recent work.

PR #280, opened 2020-03-23
client configuration refactoring

asimarslan:
Client configuration refactoring to match .net style.


Hazelcast.Net/Hazelcast.Client.Network/Connection.cs
```diff
@@ -75,31 +75,30 @@ internal sealed class Connection
...
+                    var lingerOption = new LingerOption(true, socketOptions.LingerSeconds);
                     _clientSocket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.Linger, lingerOption);
                 }
                 else
                 {
                     var lingerOption = new LingerOption(true, 0);
                     _clientSocket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.DontLinger, lingerOption);
                 }
-                _clientSocket.NoDelay = socketOptions.IsTcpNoDelay();
-                _clientSocket.ReceiveTimeout = socketOptions.GetTimeout() > 0 ? socketOptions.GetTimeout() : -1;
```
zpqrtbnk: So I think you answered already, but just wanted to make sure removing `ReceiveTimeout` is intentional - I'm ok with it.

Hazelcast.Net/Hazelcast.Config/SecurityConfig.cs
```diff
@@ -0,0 +1,73 @@
...
+            return this;
+        }
+        
+        public SecurityConfig ConfigureTokenIdentity(string encodedToken, TokenEncoding encoding)
+        {
+            TokenIdentityConfig = new TokenIdentityConfig(encodedToken, encoding);
+            return this;
+        }
+
+        public SecurityConfig ConfigureKerberosIdentity(string spn, string username, string password, string domain)
```
zpqrtbnk: Passing a username, password and domain is for tests. In normal usage, only a SPN would be passed. So we need two methods, one accepting only the SPN, and one accepting alll 4 parameters.

asimarslan: Should we mark the 4 param version as `internal` if it is only for tests?

asimarslan:
verify

PR #291, opened 2020-06-02
fix address list shuffle default to true

### LINKED ISSUE #290: Shuffle members by default
Unfortunately, the .NET client does not shuffle the list by default whereas java does by default. 

PS: Related prop is hazelcast.client.shuffle.member.list which is currently false in .NET client.

asimarslan:
fixes #290


asimarslan:
verify

asimarslan:
verify

Issue #243, opened 2019-06-18
Missing "LOADED" event


neilstevenson:
See https://github.com/hazelcast/hazelcast-go-client/issues/464

IMDG 3.11 added LOADED event - https://docs.hazelcast.org/docs/3.12.1/javadoc/com/hazelcast/core/EntryEventType.html#LOADED

This is not handled by this client

zpqrtbnk:
This is a 3.x issue, and 3.x is now in maintenance mode, as we work on the upcoming 4.x version. In 4.x, the *Loaded* event is supported (see [HDictionaryEventTypes](https://github.com/hazelcast/hazelcast-csharp-client/blob/master/src/Hazelcast.Net/Data/HDictionaryEventTypes.cs)) so I am going to close this issue. Please raise a new issue in case of concerns.

PR #293, opened 2020-06-24
Adding loaded event to the EntryAdapter

mogatti:
Looking for some feedback, related to for https://github.com/hazelcast/hazelcast-csharp-client/issues/243

I copied the unit test method from the nodejs client. The entry is added and recorded by the map loader, then expires from the map, and then loaded from the map loader.


zpqrtbnk:
This PR targets the 3.x codebase which is now in maintenance mode, as we work on the new 4.x codebase. Therefore, I am going to close the PR. Sorry no one has been able to pay more attention to PRs for a long time :(

As for the actual issue #243, I am going to comment there.

PR #303, opened 2020-09-17
Add tests for list, set, topic, ringBuffer, and some transactional objects

zpqrtbnk:


zpqrtbnk:
verify

src/Hazelcast.Net/DistributedObjects/Impl/HRingBuffer.cs
```diff
@@ -45,9 +44,12 @@ internal class HRingBuffer<TItem> : DistributedObjectBase, IHRingBuffer<TItem>
         public HRingBuffer(string name, DistributedObjectFactory factory, Cluster cluster, int maxBatchSize, ISerializationService serializationService, ILoggerFactory loggerFactory)
             : base(HRingBuffer.ServiceName, name, factory, cluster, serializationService, loggerFactory)
         {
-            _maxBatchSize = maxBatchSize;
+            MaxBatchSize = maxBatchSize;
```
asimarslan, line 47: This is not a configurable option parameter. It's an internal const value.
On Java:
```
    /**
     * The maximum number of items that can be retrieved in 1 go using the
     * {@link #readManyAsync(long, int, int, IFunction)} method.
     */
    public static final int MAX_BATCH_SIZE = 1000;
```

src/Hazelcast.Net/DistributedObjects/Impl/HSet.Removing.cs
```diff
@@ -25,30 +25,30 @@ public override async Task<bool> RemoveAsync(T item)
         {
             var itemData = ToSafeData(item);
             var requestMessage = SetRemoveCodec.EncodeRequest(Name, itemData);
-            var responseMessage = await Cluster.Messaging.SendAsync(requestMessage).CAF();
+            var responseMessage = await Cluster.Messaging.SendToKeyPartitionOwnerAsync(requestMessage, PartitionKeyData).CAF();
```
asimarslan, line 28: For fixed partition types, List, Set, Queue etc. Invocation's use a fixed value derived from name.
So using `SendToPartitionOwnerAsync` and caching a `PartitionId` instead of `PartitionKeyData`  could be better to avoid partitionId calculation on hot path.

zpqrtbnk: Implemented in new commits below. Now ensuring it build in Jenkins. Want to review again?

src/Hazelcast.Net/DistributedObjects/Impl/HRingBuffer.cs
zpqrtbnk: yes... guess I don't like magic constants and just converted it ;) will revert

asimarslan (approved):

PR #313, opened 2020-10-07
introduce shared IHCollection tests for IHList, IHSet, and IHQueue.

asimarslan:
API doc improvement
fix API methods

### DIFF
--- a/src/Hazelcast.Net.Testing/SingleMemberClientRemoteTestBase.cs
+++ b/src/Hazelcast.Net.Testing/SingleMemberClientRemoteTestBase.cs
@@ -40,7 +40,7 @@ namespace Hazelcast.Testing
         /// <summary>
         /// Gets the Hazelcast client.
         /// </summary>
-        public IHazelcastClient Client { get; private set; }
+        public static IHazelcastClient Client { get; private set; }
 
         /// <summary>
         /// Gets a disposable object that will destroy and dispose a distributed
--- a/src/Hazelcast.Net.Testing/SingleMemberClientRemoteTestBase.cs
+++ b/src/Hazelcast.Net.Testing/SingleMemberClientRemoteTestBase.cs
@@ -40,7 +40,7 @@ namespace Hazelcast.Testing
         /// <summary>
         /// Gets the Hazelcast client.
         /// </summary>
-        public static IHazelcastClient Client { get; private set; }
+        public IHazelcastClient Client { get; private set; }
 
         /// <summary>
         /// Gets a disposable object that will destroy and dispose a distributed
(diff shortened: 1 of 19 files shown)


src/Hazelcast.Net.Testing/SingleMemberClientRemoteTestBase.cs
```diff
@@ -40,7 +40,7 @@ public async Task ClientOneTimeTearDown()
         /// <summary>
         /// Gets the Hazelcast client.
         /// </summary>
-        public IHazelcastClient Client { get; private set; }
+        public static IHazelcastClient Client { get; private set; }
```
zpqrtbnk, line 43: Not sure I like this. Anything static in test classes and especially in test *base* classes can lead to collisions between tests. Can it remain an instance variable?

src/Hazelcast.Net/DistributedObjects/IHList.cs
```diff
@@ -34,28 +41,29 @@ public interface IHList<T> : IHCollection<T>
...
         /// <summary>
-        /// Replaces the element at the specified position in this list with the
-        /// specified element.
+        /// Replaces the element at the specified position in this list with the specified element.
         /// </summary>
         /// <param name="index">index index of the element to replace</param>
         /// <param name="item">element to be stored at the specified position</param>
-        /// <returns></returns>
+        /// <returns>The element previously at the specified position</returns>
         Task<T> SetAsync(int index, T item);
```
zpqrtbnk, line 67: If we return a value should this be `GetAndSetAsync` just to be consistent?

src/Hazelcast.Net.Tests/Remote/ClientBaseCollectionTest.cs
```diff
@@ -0,0 +1,288 @@
...
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Hazelcast.DistributedObjects;
+using Hazelcast.Testing;
+using NUnit.Framework;
+
+namespace Hazelcast.Tests.Remote
+{
+    public abstract class ClientBaseCollectionTest : SingleMemberClientRemoteTestBase
```
zpqrtbnk, line 25: This is a base, abstract class → should be something like `ClientCollectionTestBase`.

zpqrtbnk (commented):
A few comments, if you don't mind.

src/Hazelcast.Net.Testing/SingleMemberClientRemoteTestBase.cs
asimarslan: ok

src/Hazelcast.Net.Tests/Remote/ClientBaseCollectionTest.cs
asimarslan: ok

src/Hazelcast.Net/DistributedObjects/IHList.cs
asimarslan: ok changing the RemoveAt too

zpqrtbnk (approved):
all good

PR #316, opened 2020-10-09
Import and refactor more utilities

zpqrtbnk:

PR #317, opened 2020-10-09
Refactor yet more utilities

zpqrtbnk:
Must merge #316 first


src/Hazelcast.Net/NetStandard/StreamExtensions.cs
```diff
@@ -47,41 +48,53 @@ public static async ValueTask<int> ReadAsync(this Stream stream, Memory<byte> me
             if (stream == null) throw new ArgumentNullException(nameof(stream));
             if (memory.Length == 0) throw new ArgumentException("Empty memory.", nameof(memory));
 
-            byte[] bytes = null;
+            // this is based upon what 2.1 does
+
+            if (MemoryMarshal.TryGetArray(memory, out ArraySegment<byte> array))
+            {
+                //return await stream.ReadAsync(array.Array, array.Offset, array.Count, cancellationToken).CAF();
```
asimarslan, line 55: is this a comment to show below workaround replacement? if so, can you make it no look like comment out code.

src/Hazelcast.Net/NetStandard/StreamExtensions.cs
```diff
@@ -47,41 +48,53 @@ public static async ValueTask<int> ReadAsync(this Stream stream, Memory<byte> me
...
+                return result;
+            }
+
+            var bytes = ArrayPool<byte>.Shared.Rent(memory.Length);
             try
             {
-                // TODO: optimize
-                // of course this is sub-optimal :(
-                bytes = ArrayPool<byte>.Shared.Rent(memory.Length);
+                //var result = await stream.ReadAsync(bytes, 0, memory.Length, cancellationToken).CAF();
```
asimarslan, line 74: same here as above comment

asimarslan (approved):
minor: two line of commented out code can be removed

src/Hazelcast.Net/NetStandard/StreamExtensions.cs
zpqrtbnk: same - this is what is replaced by the code under it - ok will try to make it more explicit

src/Hazelcast.Net/NetStandard/StreamExtensions.cs
zpqrtbnk: same, same

PR #324, opened 2020-10-13
Handle orphaned subscriptions

zpqrtbnk:
Add the ability to handled orphaned subscriptions: when a user unsubscribes from events, we try to remove the corresponding listener on each member. If that succeeds then all is well. Otherwise, we still return `true` but we keep the subscription around, and mark it as orphaned. If we receive an event from a member, corresponding to an orphaned subscription, we try again to remove the listener for that member. Eventually, the subscription should be removed.

There is a corresponding test that uses our testing server to verify that it works.

Now, events are triggered by actions... if a listener remains active on a member and we never trigger an action on that member, no event may be triggered and both the listener on the member, and the subscription on the client, will leak.

A next feature (should we create an issue / feature request for it?) would be to periodically scan the subscriptions, and try to remove listeners for orphaned subscriptions, without waiting for an event? And maybe, after a (long enough) while, simply get rid of the subscription?

### DIFF
--- a/src/Hazelcast.Net/Clustering/DistributedEventScheduler.cs
+++ b/src/Hazelcast.Net/Clustering/DistributedEventScheduler.cs
@@ -28,20 +28,22 @@ namespace Hazelcast.Clustering
     internal class DistributedEventScheduler : IAsyncDisposable
     {
         private readonly Dictionary<int, Task> _partitionTasks = new Dictionary<int, Task>();
-        //private readonly ReaderWriterLockSlim _lock = new ReaderWriterLockSlim();
+        private readonly Func<Task, object, Task> _continueWithHandler;
+        private readonly Action<Task, object> _removeAfterUse;
+        private readonly Func<ClusterSubscription, Guid, long, ValueTask> _onOrphan;
         private readonly object _mutex = new object();
         private readonly ILogger _logger;
         private bool _disposed;
         private int _exceptionCount, _unhandledExceptionCount;
-        private readonly Func<Task, object, Task> _continueWithHandler;
-        private readonly Action<Task, object> _removeAfterUse;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DistributedEventScheduler"/> class.
         /// </summary>
+        /// <param name="onOrphan">A method to handle orphan events.</param>
         /// <param name="loggerFactory">A logger factory.</param>
-        public DistributedEventScheduler(ILoggerFactory loggerFactory)
+        public DistributedEventScheduler(Func<ClusterSubscription, Guid, long, ValueTask>  onOrphan, ILoggerFactory loggerFactory)
         {
+            _onOrphan = onOrphan ?? throw new ArgumentNullException(nameof(onOrphan));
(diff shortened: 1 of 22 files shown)


src/Hazelcast.Net/Clustering/DistributedEventScheduler.cs
```diff
@@ -28,17 +28,17 @@ namespace Hazelcast.Clustering
...
         private readonly ILogger _logger;
         private bool _disposed;
         private int _exceptionCount, _unhandledExceptionCount;
-        private readonly Func<Task, object, Task> _continueWithHandler;
-        private readonly Action<Task, object> _removeAfterUse;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DistributedEventScheduler"/> class.
         /// </summary>
+        /// <param name="onOrphan">A method to handle orphan events.</param>
```
asimarslan, line 41: leftover doc.

src/Hazelcast.Net/Events/EventsOptions.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            SubscriptionCollectDelay = other.SubscriptionCollectDelay;
+            SubscriptionCollectPeriod = other.SubscriptionCollectPeriod;
+            SubscriptionCollectTimeout = other.SubscriptionCollectTimeout;
+        }
+
+        /// <summary>
+        /// Gets or sets the delay before collecting subscriptions starts.
+        /// </summary>
+        public TimeSpan SubscriptionCollectDelay { get; set; } = TimeSpan.FromSeconds(1);
```
asimarslan, line 43: I think default values can be much larger to decrees server load, ~10sec

src/Hazelcast.Net/Events/EventsOptions.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        /// <summary>
+        /// Gets or sets the period of the subscription collection.
+        /// </summary>
+        public TimeSpan SubscriptionCollectPeriod { get; set; } = TimeSpan.FromSeconds(1);
```
asimarslan, line 48: same here, ~10sec maybe or even larger

src/Hazelcast.Net/Clustering/DistributedEventScheduler.cs
zpqrtbnk: ok

src/Hazelcast.Net/Events/EventsOptions.cs
zpqrtbnk: ok for 10s

src/Hazelcast.Net/Messaging/ClientMessage.cs
```diff
@@ -54,6 +54,11 @@ public ClientMessage(Frame firstFrame = null)
             }
         }
 
+        /// <summary>
+        /// Gets or sets the unique identifier of the member that sent the message, if any.
+        /// </summary>
+        public Guid Sender { get; set; }
```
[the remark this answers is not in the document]
zpqrtbnk: not anymore, forgot to remove, removing

asimarslan (approved):

PR #329, opened 2020-10-15
Refactor timespans, timeouts etc

zpqrtbnk:
depends on #343 

Refactor how we handle timespans and timeouts, as discussed removed some of the confusing extension methods.

The idea is to use `TimeSpan` everywhere but with aliases, so it becomes `TimeToLive.Infinite` or `LeaseTime.Default` or even `TimeToWait.FromSeconds(33)` so it's super obvious what the user's calls are doing + internally we try to be cleaner when translating these timespans to whatever our codecs expect (because... they sort of each have their own ideas of what zero, default, infinite etc mean).


src/Hazelcast.Net/Core/LeaseTime.cs
```diff
@@ -12,24 +12,25 @@
...
     /// </summary>
-    public static class LeaseTime
+    public static partial class LeaseTime
     {
         /// <summary>
-        /// A constants used to specify an infinite lease time.
+        /// The max lease-time (according to server configuration).
         /// </summary>
-        public static readonly TimeSpan InfiniteTimeSpan = Timeout.InfiniteTimeSpan;
+        public static TimeSpan MaxValue => TimeSpanExtensions.MinusOneMillisecond;
```
mdumandag, line 29: Why do we have `MaxValue` here, but have `Infinite` for others?

src/Hazelcast.Net/Core/IdleTime.cs
```diff
@@ -12,18 +12,25 @@
...
-        /// A constants used to specify an infinite idle-time (never idles).
+        /// An infinite idle-time (never idle).
+        /// </summary>
+        public static TimeSpan Infinite => TimeSpanExtensions.MinusOneMillisecond;
+
+        /// <summary>
+        /// The default idle-time (use the value configured on the server).
         /// </summary>
-        public static readonly TimeSpan Infinite = TimeSpan.FromMilliseconds(-1);
+        public static TimeSpan Default => TimeSpan.Zero;
```
mdumandag, line 34: Why do we name this `Default`, but some others as `Zero`?

src/Hazelcast.Net/Core/TimeToWait.cs
```diff
@@ -12,24 +12,25 @@
...
         /// </summary>
-        public static readonly TimeSpan InfiniteTimeSpan = Timeout.InfiniteTimeSpan;
+        public static TimeSpan Infinite => TimeSpanExtensions.MinusOneMillisecond;
 
         /// <summary>
-        /// A constants used to specify a zero time-to-wait (do not wait).
+        /// A zero time-to-wait (do not wait).
         /// </summary>
-        public static readonly TimeSpan Zero = TimeSpan.Zero;
+        public static TimeSpan Zero => TimeSpan.Zero;
```
mdumandag, line 34: Why do we name this `Default`, but some others as `Zero`?

src/Hazelcast.Net/Core/LeaseTime.cs
zpqrtbnk: Because... on the server it seems that you cannot specify an infinite lease time - it is either a specified value, or the default value which is configured on the server, and trying to use a value greater than that one returns an error. So... `MaxValue` tries to convey the idea that you're getting the maximum allowed value, which is not infinite.

src/Hazelcast.Net/Core/IdleTime.cs
zpqrtbnk: Because... an idle time of zero does not make sense, but for the server passing zero means "use the default, configured value". So `Default` carries that notion.

src/Hazelcast.Net/Core/TimeToWait.cs
zpqrtbnk: See other anwser.

zpqrtbnk:
@mdumandag A general comment here... The `LeaseTime`, `IdleTime` classes are syntactic sugar, trying to expose the *meaning* of these values. For some of them, you can specify "zero", "infinite", or a value. But for some of them you can also specify "default" which means "use the configuration value on the server". For others, there is no "infinite" but a "max value" configured on the server. My idea was that ppl should not have to think about what happens if they want this or that behavior.

But I'm not sure this is 100% the right way of doing this, we have been discussing it back and forth with Asim. Any idea is appreciated :)

mdumandag (approved):

mdumandag (approved):

src/Hazelcast.Net/Core/TimeOut.cs
```diff
@@ -0,0 +1,36 @@
...
+#nullable enable
+
+using System;
+
+namespace Hazelcast.Core
+{
+    /// <summary>
+    /// Provides time-out constants.
+    /// </summary>
+    public static partial class TimeOut
```
asimarslan, line 24: Can we write it as `Timeout`  ?

zpqrtbnk: There's already a `Timeout` class in .NET so... no.

src/Hazelcast.Net/DistributedObjects/Impl/HReplicatedMap.cs
```diff
@@ -39,13 +39,16 @@ public HReplicatedMap(string name, DistributedObjectFactory factory, Cluster clu
         }
 
         public Task<TValue> PutAsync(TKey key, TValue value)
-            => PutAsync(key, value, TimeToLive.InfiniteTimeSpan);
+            => PutAsync(key, value, TimeToLive.Infinite);
 
         public async Task<TValue> PutAsync(TKey key, TValue value, TimeSpan timeToLive)
```
asimarslan, line 44: The default value is zero so this can be simplified as 
`public async Task<TValue> PutAsync(TKey key, TValue value, TimeSpan timeToLive = default)`
and remove the above call 
```
       public Task<TValue> PutAsync(TKey key, TValue value)
            => PutAsync(key, value, TimeToLive.InfiniteTimeSpan);	            => PutAsync(key, value, TimeToLive.Infinite);

```

src/Hazelcast.Net/DistributedObjects/Impl/HMap.Setting.cs
```diff
@@ -26,11 +26,11 @@ internal partial class HMap<TKey, TValue> // Setting
     {
         /// <inheritdoc />
         public Task SetAsync(TKey key, TValue value)
-            => SetAsync(key, value, TimeToLive.InfiniteTimeSpan, IdleTime.Infinite);
+            => SetAsync(key, value, TimeToLive.Infinite, IdleTime.Infinite);
```
asimarslan, line 29: Should be `TimeToLive.Default`

src/Hazelcast.Net/DistributedObjects/Impl/HMap.Setting.cs
```diff
@@ -26,11 +26,11 @@ internal partial class HMap<TKey, TValue> // Setting
...
 
         /// <inheritdoc />
         public Task<TValue> PutAsync(TKey key, TValue value)
-            => PutAsync(key, value, TimeToLive.InfiniteTimeSpan, IdleTime.Infinite);
+            => PutAsync(key, value, TimeToLive.Infinite, IdleTime.Infinite);
```
asimarslan, line 33: Should be `PutAsync(key, value, TimeToLive.Default, IdleTime.Default);`

zpqrtbnk:
verify

src/Hazelcast.Net/DistributedObjects/IHMap.Setting.cs
```diff
@@ -26,19 +27,25 @@ public partial interface IHMap<TKey, TValue> // Setting
...
-        /// <param name="timeToLive">A time to live.</param>
+        /// <param name="timeToLive">A time to live (<see cref="TimeToLive.Infinite"/> i.e. 0ms to live
+        /// forever; <see cref="TimeToLive.Default"/> i.e. -1ms to use the server-configured value).</param>
         /// <returns>A task that will complete when the entry has been added or updated.</returns>
         /// <remarks>
         /// <para>The value is automatically expired, evicted and removed after the
-        /// <paramref name="timeToLive"/> has elapsed.</para>
-        /// TODO: document zero & infinite
+        /// <paramref name="timeToLive"/> has elapsed. If <paramref name="timeToLive"/> is <see cref="TimeToLive.Infinite"/>
+        /// i.e. -0ms, the value is retained indefinitely. If it is <see cref="TimeToLive.Default"/> i.e. -1ms,
```
mdumandag, line 46: Minor: `-0ms` -> `0ms`. There are also a couple of such places below

src/Hazelcast.Net/DistributedObjects/Impl/HMultiMap.cs
```diff
@@ -289,22 +289,22 @@ public async Task ClearAsync()
 
         /// <inheritdoc />
         public Task LockAsync(TKey key)
-            => LockAsync(key, Timeout.InfiniteTimeSpan);
+            => LockAsync(key, LeaseTime.MaxValue);
 
         /// <inheritdoc />
         public Task<bool> TryLockAsync(TKey key)
-            => TryLockAsync(key, TimeSpan.Zero, Timeout.InfiniteTimeSpan);
+            => TryLockAsync(key, TimeToWait.Zero, LeaseTime.MaxValue);
```
mdumandag, line 296: Java uses `Long.MAX_VALUE` instead of -1 (`LeaseTime.MaxValue`)

src/Hazelcast.Net/DistributedObjects/Impl/HMultiMap.cs
```diff
@@ -289,22 +289,22 @@ public async Task ClearAsync()
...
 
         /// <inheritdoc />
         public Task<bool> TryLockAsync(TKey key, TimeSpan timeToWait)
-            => TryLockAsync(key, timeToWait, Timeout.InfiniteTimeSpan);
+            => TryLockAsync(key, timeToWait, LeaseTime.MaxValue);
```
mdumandag, line 300: Java uses `Long.MAX_VALUE` instead of -1 (`LeaseTime.MaxValue`)

src/Hazelcast.Net/DistributedObjects/Impl/HTxMap.cs
```diff
@@ -91,28 +91,29 @@ public async Task<IReadOnlyList<TKey>> GetKeysAsync(IPredicate predicate)
...
-        public Task<TValue> PutAsync(TKey key, TValue value)
-            => PutAsync(key, value, TimeToLive.InfiniteTimeSpan);
+            => SetAsync(key, value, TimeToLive.Infinite);
 
         public async Task SetAsync(TKey key, TValue value, TimeSpan timeToLive)
         {
             var (keyData, valueData) = ToSafeData(key, value);
-            var timeToLiveMilliseconds = timeToLive.CodecMilliseconds(-1);
+            // FIXME MEH TTL?
+            var timeToLiveMs = timeToLive.RoundedMilliseconds(false);
```
mdumandag, line 100: I think we don't support TTL for TxnMap, so the codec does not have this parameter. I think we can remove it from this client too

src/Hazelcast.Net/DistributedObjects/Impl/HTxMap.cs
```diff
@@ -91,28 +91,29 @@ public async Task<IReadOnlyList<TKey>> GetKeysAsync(IPredicate predicate)
...
 
-            // FIXME timeToLive is ignored?
             var requestMessage = TransactionalMapSetCodec.EncodeRequest(Name, TransactionId, ContextId, keyData, valueData);
             var responseMessage = await Cluster.Messaging.SendToMemberAsync(requestMessage, TransactionClientConnection).CAF();
             _ = TransactionalMapSetCodec.DecodeResponse(responseMessage);
         }
 
+        public Task<TValue> PutAsync(TKey key, TValue value)
+            => PutAsync(key, value, TimeToLive.Infinite);
```
mdumandag, line 108: Java uses `-1` instead of `0`

src/Hazelcast.Net/DistributedObjects/Impl/HMultiMap.cs
mdumandag: Never mind, after looking at the implementation, `-1` seems like the correct value

src/Hazelcast.Net/DistributedObjects/Impl/HMultiMap.cs
mdumandag: Never mind, after looking at the implementation, `-1` seems like the correct value

src/Hazelcast.Net/DistributedObjects/IHMap.Removing.cs
```diff
@@ -25,12 +26,15 @@ public partial interface IHMap<TKey, TValue> // Removing
         /// within the specified time to wait value.
         /// </summary>
         /// <param name="key">A key.</param>
-        /// <param name="timeToWait">The time to wait for a lock on the key.</param>
+        /// <param name="timeToWait">How long to wait for a lock on the key (<see cref="TimeToWait.Infinite"/> i.e. -1ms to wait forever;
```
mdumandag, line 29: We need to update the documentation. It is mentioning deleted files.

src/Hazelcast.Net/DistributedObjects/Impl/HTxMap.cs
```diff
@@ -91,28 +91,29 @@ public async Task<IReadOnlyList<TKey>> GetKeysAsync(IPredicate predicate)
...
+            var timeToLiveMs = timeToLive.RoundedMilliseconds(false);
 
-            // FIXME timeToLive is ignored?
             var requestMessage = TransactionalMapSetCodec.EncodeRequest(Name, TransactionId, ContextId, keyData, valueData);
             var responseMessage = await Cluster.Messaging.SendToMemberAsync(requestMessage, TransactionClientConnection).CAF();
             _ = TransactionalMapSetCodec.DecodeResponse(responseMessage);
         }
 
+        public Task<TValue> PutAsync(TKey key, TValue value)
+            => PutAsync(key, value, TimeSpan.Zero);
```
mdumandag, line 108: Java client uses -1 ms

src/Hazelcast.Net/Core/TimeOut.cs
zpqrtbnk: class has been removed

src/Hazelcast.Net/DistributedObjects/Impl/HMap.Setting.cs
zpqrtbnk: fixed

src/Hazelcast.Net/DistributedObjects/Impl/HMap.Setting.cs
zpqrtbnk: fixed

src/Hazelcast.Net/DistributedObjects/IHMap.Setting.cs
zpqrtbnk: fixing

src/Hazelcast.Net/DistributedObjects/Impl/HTxMap.cs
zpqrtbnk: great, updating

src/Hazelcast.Net/DistributedObjects/Impl/HReplicatedMap.cs
zpqrtbnk: ok

src/Hazelcast.Net/DistributedObjects/IHMap.Removing.cs
zpqrtbnk: ok

mdumandag (approved):

asimarslan (approved):

PR #331, opened 2020-10-29
Troubleshoot heartbeat

zpqrtbnk:
Heartbeat had various problems. First, we were using `DateTime.Now` in sockets and `DateTime.UtcNow` in heartbeat, resulting in false timeouts. Second, values seem small + we end up having timeouts when reading a very big message and a heartbeat ping answer comes in (but waits).

~This PR fixes the most obvious problem, still we should discuss the second point (receiving messages in a pure sequential way can be a problem, at least for some messages).~ ignore this, it's not a thing, we already process responses in parallel - it was only a timeout issue


src/Hazelcast.Net/Clustering/MemberConnection.cs
```diff
@@ -379,7 +397,26 @@ public async Task<ClientMessage> SendAsync(ClientMessage message, CancellationTo
...
+        /// <param name="timeoutMilliseconds">A timeout, in milliseconds.</param>
+        /// <param name="cancellationToken">A cancellation token.</param>
+        /// <returns>A task that will complete when the response has been received, and represents the response.</returns>
+        /// <remarks>
+        /// <para>The operation must complete within the specified operation timeout.</para>
+        /// </remarks>
+        public Task<ClientMessage> SendAsync(Invocation invocation, int timeoutMilliseconds, CancellationToken cancellationToken)
+            => TaskEx.RunWithTimeout(SendAsync2, invocation, timeoutMilliseconds, cancellationToken);
+
+        private async Task<ClientMessage> SendAsync2(Invocation invocation, CancellationToken cancellationToken)
```
asimarslan, line 419: `SendAsyncInternal` ?

src/Hazelcast.Net/Clustering/HeartbeatOptions.cs
```diff
@@ -42,7 +41,7 @@ private HeartbeatOptions(HeartbeatOptions other)
         /// <para>Heartbeat will run periodically, and send a ping request to connections
         /// that have not been written to over the previous period.</para>
         /// </remarks>
-        public int PeriodMilliseconds { get; set; } = 5_000;
+        public int PeriodMilliseconds { get; set; } = 10_000;
```
asimarslan, line 44: java client default is 5000 ms, can we use the same default value here?

src/Hazelcast.Net/Clustering/Heartbeat.cs
```diff
@@ -122,7 +112,7 @@ private async Task RunAsync(CancellationToken cancellationToken)
         {
             _logger.LogDebug("Run heartbeat");
 
-            var now = DateTime.UtcNow;
+            var now = DateTime.Now; // now, or utcNow, but *must* be same as what is used in socket connection base!
```
asimarslan, line 115: Can we refactor out the `DataTime.UtcNow` usage altogether?
(Maybe just the static EPOC date)

asimarslan: What about just using the internal Clock class to make them same everywhere?

asimarslan (approved):
minor detail.

src/Hazelcast.Net/Clustering/MemberConnection.cs
zpqrtbnk: ok

src/Hazelcast.Net/Clustering/HeartbeatOptions.cs
zpqrtbnk: ok

src/Hazelcast.Net/Clustering/Heartbeat.cs
zpqrtbnk: looking into it

PR #335, opened 2020-11-16
Misc fixes

zpqrtbnk:
Includes:
* `MemberInfo.IsLiteMember` is public, no need for `IsLite` property
* `IExecutor` + implementation have been removed, will re-add later
* `NetworkAddress` is more of a DTO (will be further refined in other PRs)
* `IHazelcastClient.Members` is a `IReadOnlyCollection<MemberInfo>`


src/Hazelcast.Net/Events/ClientLifecycleState.cs
```diff
@@ -16,6 +16,13 @@ namespace Hazelcast.Events
...
     public enum ClientLifecycleState
     {
+        // Active    Connecting
+        // Active    Connected
+        // NotActive Disconnecting
+        // Active    Reconnecting (?)
+        // NotActive Disconnected
+        // 
+
         Starting,
```
asimarslan, line 26: Are we firing all of these State events or is it in another PR?

src/Hazelcast.Net/Networking/NetworkingOptions.cs
```diff
@@ -47,6 +47,16 @@ private NetworkingOptions(NetworkingOptions other)
             ConnectionRetry = other.ConnectionRetry.Clone();
         }
 
+        /// <summary>
+        /// Gets the default Hazelcast server port.
+        /// </summary>
+        public int DefaultPort { get; } = 5701;
```
asimarslan, line 53: These two are not options. They are just internal consts.

src/Hazelcast.Net/Events/ClientLifecycleState.cs
zpqrtbnk: Another PR.

src/Hazelcast.Net/Networking/NetworkingOptions.cs
zpqrtbnk: Fixed and pushed.

asimarslan (approved):

PR #336, opened 2020-11-23
Make more things internal + cleaner

zpqrtbnk:
ConnectionAdded/Removed events - internal 
HazelcastOptions#DefaultClientNamePrefix - internal
HazelcastOptions#Hazelcast - internal
HazelcastOptions#Messaging#MaxFastInvocationCount - internal
NearCacheOptions#CleanupPeriodSeconds and EvictionPercentage - internal
CloudOptions#UrlBase - internal
SocketOptions#BufferSizeKiB - documented
NetworkingOptions#RetryOnTargetDisconnected - renamed RedoOperations
NetworkingOptions#WaitForClientMilliseconds - documented + renamed + moved to cluster options
HazelcastOptions#ClientNamePrefix - internal
CloudOptions#Enabled - automatically enable if there is a token
ConnectionLifecycleEventArgs - internal
DistributedObjectLifecycleEventType - "Nothing" enumeration value removed
MemberLifecycleEventType - "Nothing" enumeration value removed
HashCode - internal

### DIFF
--- a/src/Hazelcast.Net/Networking/CloudOptions.cs
+++ b/src/Hazelcast.Net/Networking/CloudOptions.cs
@@ -32,7 +32,6 @@ namespace Hazelcast.Networking
         /// </summary>
         private CloudOptions(CloudOptions other)
         {
-            Enabled = other.Enabled;
             DiscoveryToken = other.DiscoveryToken;
             UrlBase = other.UrlBase;
         }
@@ -40,7 +39,7 @@ namespace Hazelcast.Networking
         /// <summary>
         /// Whether Hazelcast Cloud is enabled.
         /// </summary>
-        public bool Enabled { get; set; }
+        public bool Enabled => !string.IsNullOrWhiteSpace(DiscoveryToken);
 
         /// <summary>
         /// Gets or sets the discovery token of the cluster.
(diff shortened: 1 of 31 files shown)


src/Hazelcast.Net/Networking/CloudOptions.cs
```diff
@@ -32,15 +32,14 @@ public CloudOptions()
...
-            Enabled = other.Enabled;
             DiscoveryToken = other.DiscoveryToken;
             UrlBase = other.UrlBase;
         }
 
         /// <summary>
         /// Whether Hazelcast Cloud is enabled.
         /// </summary>
-        public bool Enabled { get; set; }
+        public bool Enabled => !string.IsNullOrWhiteSpace(DiscoveryToken);
```
asimarslan, line 42: I think we should remove the Enable field. The token field is used for enabling. node.js uses the same idea.

zpqrtbnk: ah, will make it internal all right - it's nice to have because in various places we want to know whether it's "enabled" and it's better to have a bool property - but I get it that users don't need to see it

asimarslan (approved):

PR #338, opened 2020-11-24
Refactoring following reviews (2)

zpqrtbnk:
AggregatorBase<TResult>  -  internal
IAggregator implementations, I couldn't find a reason to make them public - internal
Why do we have IAggregator and IAggregator<TResult>? Couldn't IAggregator<TResult> be enough? - yes, fixed
Predicate - all the implementations should be internal and we return IPredicate


src/Hazelcast.Net/Predicates/Predicate.cs
```diff
@@ -12,119 +12,86 @@
...
+        public static IPredicate Or(params IPredicate[] predicates)
+            => new OrPredicate(predicates);
+
+        public static IPredicate Sql(string sql)
+            => new SqlPredicate(sql);
+
+        public static IPredicate True()
+            => new TruePredicate();
+
+        public static IPredicate Partition(object partitionKey, IPredicate predicate)
```
mdumandag, line 91: This should return something like `IPartitionPredicate`. It has some extra methods defined. See https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/query/PartitionPredicate.java

src/Hazelcast.Net/Predicates/Predicate.cs
```diff
@@ -12,119 +12,86 @@
...
+            => new PartitionPredicate(partitionKey, predicate);
+
+        public static IPredicate Page(int pageSize, IPredicate predicate, IComparer<KeyValuePair<object, object>> comparer)
```
mdumandag, line 94: This should return something like `IPagingPredicate`. It has some extra methods defined. See https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/query/PagingPredicate.java

src/Hazelcast.Net/Predicates/Predicate.cs
```diff
@@ -12,119 +12,86 @@
...
-            return new SqlPredicate(sql);
-        }
-
-        public static TruePredicate True()
-        {
-            return new TruePredicate();
-        }
+            => new PredicateProperty(ThisConst);
+
+        public static IPredicate InstanceOf(string fullJavaClassName)
```
mdumandag, line 37: I am not sure about that, but if we have `IsXX` methods below, should we also rename this to `IsInstanceOf` for consistency? 

By the way, I prefer shorter names like this. For example, I would also change  `IsGreaterThanOrEqual` to `GreaterOrEqual`, but I guess explicit names are a more common thing in C#. So, the current state might be more appropriate for C#

zpqrtbnk: there's a later PR I'm working on that normalizes these names, so I won't change this here

src/Hazelcast.Net/Predicates/Predicate.cs
zpqrtbnk: I can see it in Java yes, now I am wondering whether all properties need to be exposed in `IPartitionPredicate`? Java would have `PartitionKey` and `Target`, are these two properties useful for users?

src/Hazelcast.Net/Predicates/Predicate.cs
zpqrtbnk: I can see it in Java yes, now I am wondering whether all properties need to be exposed in `IPagingPredicate`? Java would have `NextPage`, `PreviousPage`, `Page`, `PageSize`, `Anchor` and a `Reset()` method, are they all useful for users?

Meh - I also note that in Java these two predicates are typed ie it's `PartitionPredicate<K,V>` but totally not the case in C#... need to look into it...

src/Hazelcast.Net/Predicates/Predicate.cs
zpqrtbnk: ok, that one is easy, will add it and update the PR

src/Hazelcast.Net/Predicates/Predicate.cs
zpqrtbnk: Right, I have pushed a fix to the PR that implements the two interfaces.

I still have concerns about the partition predicate being typed (`<K,V>`) in Java vs being `<object,object>` in C# but I have put a `// FIXME` comment in the code, so if it looks good enough for you I propose we merge this PR, and fix later.

src/Hazelcast.Net/Predicates/IPagingPredicate.cs
```diff
@@ -0,0 +1,75 @@
...
+        // note that being typed would require ... the query itself to be typed?
+
+        /// <summary>
+        /// Gets the comparer that is used to sort the results on the client.
+        /// </summary>
+        /// <remarks>
+        /// <para>The comparer implementation should be serializable by Hazelcast,
+        /// and should have a corresponding identical implementation on the server.</para>
+        /// </remarks>
+        IComparer<KeyValuePair<object, object>> Comparer { get; }
```
mdumandag, line 68: I guess we can remove this method. For a client, this does not offer anything valuable. Maybe it makes sense for Java since server and the client are using the same interface

src/Hazelcast.Net/Predicates/IPagingPredicate.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /// <summary>
+        /// Gets the anchor entry, i.e. the last entry on the previous page.
+        /// </summary>
+        KeyValuePair<object, object> Anchor { get; }
```
mdumandag, line 73: I couldn't find a use for this method also. In Python client, I removed this. 

Maybe we can do the same here and add it only if some user requests it(highly unlikely).

src/Hazelcast.Net/Predicates/Predicate.cs
mdumandag: Hmm, looking it again, these in fact seems unnecessary to expose. But since you did, it can stay like this if you want

src/Hazelcast.Net/Predicates/IPagingPredicate.cs
zpqrtbnk: Made it internal

src/Hazelcast.Net/Predicates/IPagingPredicate.cs
zpqrtbnk: Made it internal

mdumandag (approved):

PR #339, opened 2020-11-24
Refactoring following reviews (3)

zpqrtbnk:
IndexOptions - Equivalent of ValidateAttribute(IndexOptions, String) is private on Java.
ProtocolType - Related to advanced network configuration, can also be made private.
IndexType - There is no enum variant for Bitmap indexes - added
BytesExtensions - I believe we don't expose buffer objects we operate on - internal
EndpointQualifier - I believe this class is added to make protocol code generator happy. It is only useful for advanced network configuration(which Java has, but I believe other clients don't), so it should be good to make it private. - internal
MemberInfo - AddressMap this is also related to advanced network config, I believe this one can be made private too - internal
BytesExtensions#ReadUtf8Char(Byte[], ref Int32) (and write too) - The documentation says that it operates on 1,2 or 3 bytes chars. What about 4 bytes chars like 😁? - fixed, chars are only for 1, 2 or 3 but added support for strings with support for surrogate pairs (4 bytes)
BitmapIndexOptions - There is no documentation for the class and its fields. - documented
IndexOptions - There are some missing documentation (documentation for class, BitmapIndexOptions, addAttributes method) - added
NearCacheOptions#EvictionPolicy - Java also have Random eviction option - implemented

### DIFF
--- a/src/Hazelcast.Net/Core/BytesExtensions.ReadFromByteArray.cs
+++ b/src/Hazelcast.Net/Core/BytesExtensions.ReadFromByteArray.cs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 using System;
+using System.Buffers;
 
 namespace Hazelcast.Core
 {
@@ -263,12 +264,16 @@ namespace Hazelcast.Core
                 var b = bytes[position];
                 var x = b >> 4;
 
+                // 1 byte, 7 bits, represented as 0vvvvvvv
+                //                                ^^^^
                 if (x >= 0 && x <= 7)
                 {
                     position += 1;
                     return (char)b;
                 }
 
+                // 2 bytes, 11 bits, represented as 110vvvvv 10vvvvvv
+                //                                  ^^^^
                 if (x == 12 || x == 13)
                 {
                     if (bytes.Length < position + 2 * SizeOfByte)
@@ -280,6 +285,8 @@ namespace Hazelcast.Core
                     return (char)(first | second);
                 }
 
+                // 3 bytes, 16 bits, represented as 1110vvvv 10vvvvvv 10vvvvvv
+                //                                  ^^^^
                 if (x == 14)
                 {
                     if (bytes.Length < position + 3 * SizeOfByte)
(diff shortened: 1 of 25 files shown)


src/Hazelcast.Net/Serialization/ByteArrayObjectDataOutput.cs
```diff
@@ -363,13 +363,18 @@ public void Write(string value, Endianness endianness = Endianness.Unspecified)
         {
             endianness = endianness.Resolve(DefaultEndianness);
 
+            // length is the length of the string, in chars
+            // each char can be 1, 2 or 3 bytes - surrogate pairs are reported as 2 chars
+            // note: this is consistent with Java
+
             var length = value?.Length ?? ArraySerializer.NullArrayLength;
```
mdumandag, line 370: Is this the UTF-8 length of the string or some sort of character count? For example, in Python `len("😀")` is equal to 1, but `len("😀".encode("utf-8"))` is equal to 4. We should be writing 4 as the length here because the encoded string will require this much space. Can you check this is writing the correct length?

src/Hazelcast.Net/NearCaching/LfuComparer.cs
```diff
@@ -40,4 +40,19 @@ public int Compare(NearCacheEntry x, NearCacheEntry y)
             return cx.CompareTo(cy);
         }
     }
+
+    /// <summary>
+    /// Compares <see cref="NearCacheEntry"/> randomly.
+    /// </summary>
+    internal class RandomComparer : IComparer<NearCacheEntry>
```
mdumandag, line 47: Is it okay to put this class under the `LfuComparer.cs` filename?

src/Hazelcast.Net/Core/BytesExtensions.ReadFromByteArray.cs
```diff
@@ -296,6 +303,69 @@ public static char ReadUtf8Char(this byte[] bytes, ref int position)
...
+            var buffer = ArrayPool<char>.Shared.Rent(length);
+            try
+            {
+                bytes.ReadUtf8Chars(ref position, buffer, length);
+                return new string(buffer, 0, length);
+            }
+            finally
+            {
+                ArrayPool<char>.Shared.Return(buffer);
+            }
```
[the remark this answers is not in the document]
zpqrtbnk: We could use `Encoding.UTF8` as well... it's not as flexible as using our code + more generic and therefore slightly slower - but I get your point, would it be OK to merge "as is" + create a task to revisit this later on?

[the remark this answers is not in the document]
mdumandag: I see. I don't have any problems with the current code as long as the way we do encoding is compatible with other clients. Did you test that?

src/Hazelcast.Net/Serialization/ByteArrayObjectDataOutput.cs
zpqrtbnk: The length is the number of `char` items in the string, bearing in mind that every symbol that encodes on 1, 2 or 3 bytes is one `char` for C#, whereas symbols that encode on 4 bytes are two `char` (aka surrogate pair) - bit weird, but that is also what Java seems to do so AFAIK we would be compatible with Java here?

src/Hazelcast.Net/NearCaching/LfuComparer.cs
zpqrtbnk: Totally not, fixing

src/Hazelcast.Net/Serialization/ByteArrayObjectDataOutput.cs
mdumandag: I believe that could lead to some compatibility problems with other clients. I see that Java writes the length of the  UTF-8 encoded byte array (Python and Node.js also do that, not sure about C++) https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/ByteArrayObjectDataOutput.java#L261

src/Hazelcast.Net/Core/BytesExtensions.ReadFromByteArray.cs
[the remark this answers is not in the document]
zpqrtbnk: No, so going to test now

src/Hazelcast.Net/Serialization/ByteArrayObjectDataOutput.cs
zpqrtbnk: Then I mis-read the Java code, let me check

mdumandag (approved):
After an internal discussion, approving the PR as it is. Serialization related fixes will be a part of another PR

asimarslan (commented):
Ok

PR #340, opened 2020-11-26
Refactoring following reviews (4)

zpqrtbnk:
depends on #339 

RemoteError - Java added some new errors to this enum. These five new errors could be added to C# also.  - added
HazelcastOptions#NearCaching#PatternMatcher - Is this only applies to NearCaches? It should be applicable to other configs such as FlakeIdGeneratorConfig, ReliableTopicConfig etc. (We might not have them right now, but they could be added later) - moved to HazelcastOptions
The documentation of IPatternMatcher is stale, it is mentioning ClientConfig etc. - cleaned
TaskEx - Maybe it would be good to document that, when you cancel the task/task gets canceled due to timeout, it may be too late (request is sent to the server), so it is just a client-side cancellation, the operation may still take place on the server-side. -- documented, although this is internal
HazelcastOptions#Messaging - The names are different from the Java, MinRetryDelay is intuitive, but I think it might be good to rename OpertionTimeout to InvocationTimeout. -- renamed
SingleAttributeProjection - This implementation can be made private, and we could provide some factory class called Projections(like we did for aggregators and predicates) with a method like SingleAttribute -- internal + created Project static class
IProjection - Documentation can be improved, there are typos like `Use an to perform` -> `Use an Aggregator to perform` -- cleaned up
Maybe rename Aggregator to Aggregators. It's thin factory class for other aggregators, when I saw it's name, I thought it was the interface of the aggregators or some sort of abstract class for users to write client counterpart of their own aggregators. -- renamed Aggregate.
If we create Project and Aggregate static classes then what about predicates? Query -- done
Predicates - can we have a better syntax? - yes, see PredicatesTests.QuerySyntaxes
MultiAttributeProjection is missing, I believe this couldn't be implemented on 3.x series since it was returning Java serializable, but as far as I know, this has changed now, it should be possible to implement this on non-Java clients. - implemented


src/Hazelcast.Net/Projection/Projections.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// <param name="attributePath">The attribute.</param>
+        /// <returns>A projection.</returns>
+        public static IProjection SingleAttribute(string attributePath)
+            => new SingleAttributeProjection(attributePath);
+
+        /// <summary>
+        /// Projects multiple attributes.
+        /// </summary>
+        /// <returns>A projection.</returns>
+        public static IProjection MultipleAttribute()
```
mdumandag, line 34: Shouldn't this take multiple attribute paths and pass them into actual implementation?

zpqrtbnk: doh yes fixing

zpqrtbnk: pushed the fix

mdumandag (approved):

PR #341, opened 2020-11-27
Refactoring following reviews (5)

zpqrtbnk:
depends on #340 

Hazelcast.Data? Maybe it would be good idea to change the name of this namespace? We can move some of the items to core or configuration. When I saw the name Data, I am automatically thinking of serialization, but that may not be the case for end users. - renamed
Hazelcast.Protocol.Data: same - renamed
Fixed mutual dependencies that were introduced, now clean
Cleaned files up, ensure we have headers, etc.
Do we need IExecutable and IExecutable<TResult>? Also, they should be marker interfaces on the client-side, there is no need for the execute method. -- not implemented nor supported at all now, so, removed these two interfaces
ExecutionResult - Although not implement now, C# ExecutorService returns this (basically list of member id, result tuples) whereas Java returns Map of member id to Future<Result>, so there is a slight difference. -- not implemented nor supported at all now, so, removed this class


mdumandag (approved):

asimarslan (commented):
OK

PR #344, opened 2020-12-01
Serialization service refactoring

asimarslan:


zpqrtbnk:
Quick, general questions while looking at the code apart from the actual serialization code:
* `SerializationService` is now internal + interface is gone, which means we won't expose the serialization service to users, ever. Just want to validate that this is OK.
* Should `ReadUTF` and `WriteUTF` be `ReadUTF8` and `WriteUTF8` to be super explicit?
* Removing the null checks (for instance in `AndPredicate`) certainly makes sense but looks scary + we risk re-introducing the check. Could we mark the parameter with the `NotNull` attribute from `System.Diagnostics.CodeAnalysis` so it's clear it cannot be null: `public void ReadData([NotNull] IObjectDataInput input) {...}`?
* Have you re-introduced the `Executor` class and is this intentional?

Now reviewing the actual serialization code.

zpqrtbnk:
More questions on the actual serialization code:
* So `IDataInput` is merged into `IObjectDataInput` - do we still need that `Object` there or could it be named `IDataInput`? + `IDataInput` had nice `///` documentation that maybe we should keep as it's a public interface? Same for everything `...Output`.
* In `ByteExtensions` again I understand why we remove the checks (for `null` or bad `position` etc) but then I'd like to have a big blob of documentation at the top of each file explaining *where* the checks are expected to take place (data input & output should in theory validate everything, right?)
* `Endianness` now is only `BigEndian` or `LittleEndian`, OK with this. Do we still use the native thing?
* Rest seems OK

src/Hazelcast.Net/Serialization/IObjectDataInput.cs
```diff
@@ -12,24 +12,60 @@
...
-        IData ReadData();
-
-        /// <summary>
-        /// Reads an object.
-        /// </summary>
-        /// <typeparam name="T">The type of the object.</typeparam>
-        /// <returns>The object.</returns>
+        Endianness Endianness { get; }
+
+        bool ReadBoolean();
```
mdumandag, line 23: Will the documentation be added later?

src/Hazelcast.Net/Serialization/IObjectDataOutput.cs
```diff
@@ -12,29 +12,64 @@
...
-        /// Writes an <see cref="IData"/> instance.
-        /// </summary>
-        /// <param name="value">The <see cref="IData"/> instance.</param>
-        void WriteData(IData value);
-
-        /// <summary>
-        /// Writes an object.
-        /// </summary>
-        /// <param name="value">The object.</param>
+        Endianness Endianness { get; }
```
mdumandag, line 21: Will the documentation be added later?

mdumandag (approved):
Approving with minor comments related to missing documentation.

Also, we should make the `IDataSerializerHook` internal. It is the only thing missing from the spreadsheet

zpqrtbnk:
Could you remove all the custom code for UTF8 or do you want to do that later? I see the reading code is gone but the writing code seems to be there?

src/Hazelcast.Net/Serialization/IObjectDataInput.cs
asimarslan: done

src/Hazelcast.Net/Serialization/IObjectDataOutput.cs
asimarslan: done

zpqrtbnk (approved):
ok

PR #346, opened 2020-12-04
Refactor client state management

zpqrtbnk:
This is a general refactoring of the client state management (connect, re-connect, events, etc). Most changes are internal. 

The public changes are:
* `ClientLifecycleEvent` is replaced by `StateChanged` and the args for this event contains the new state (see below)
* `ClientNotConnectedException` now exposes the state the client was in when the exception was thrown
* `ClusterMemberLifecycleEvent` is replaced by `MembersUpdatedEvent` and the args for this event contains the removed & added members, along with the new list of members
* `DistributedObjectLifecycleEvent` is replaced by two events, `DistributedObjectCreated` and `DistributedObjectDestroyed`

The various enums that represented the state of the client and the cluster are merged into one `ConnectionState` enum with the following possible values:
* `NotConnected` : the client is not connected
* `Connecting` : the client is connecting (will transition to `Connected` if ok, to `NotConnected` otherwise)
* `Connected` : the client is connected (at least 1 connection to a member) (will transition to `Disconnected` if the last connection is lost, to `Disconnecting` if requested to shutdown)
* `Disconnected` : the client just lost its last connection (this is a transitory state, will transition to `Reconnecting` if configured to reconnect, to `NotConnected` otherwise)
* `Reconnecting` : the client, after having been disconnected, is trying to reconnect (will transition to `Connected` if ok, to `NotConnected` otherwise)
* `Disconnecting` : the client has been requested to shutdown (will transition to `NotConnected` when done)

Internally, lots of cleanup to make the relationship between the different part (services) of the cluster more clear, with no circular dependencies, clearly identified background tasks, etc. I have a draft document that details all this, which will be part of the documentation for devs to understand it all.


zpqrtbnk:
killing this PR as the  branch has changed

PR #347, opened 2020-12-04
Refactor client state & connection

zpqrtbnk:
This is a general refactoring of the client state management (connect, re-connect, events, etc). Most changes are internal. 

The public changes are:
* `ClientLifecycleEvent` is replaced by `StateChanged` and the args for this event contains the new state (see below)
* `ClientNotConnectedException` now exposes the state the client was in when the exception was thrown
* `ClusterMemberLifecycleEvent` is replaced by `MembersUpdatedEvent` and the args for this event contains the removed & added members, along with the new list of members
* `DistributedObjectLifecycleEvent` is replaced by two events, `DistributedObjectCreated` and `DistributedObjectDestroyed`

The various enums that represented the state of the client and the cluster are merged into one `ConnectionState` enum with the following possible values:
* `NotConnected` : the client is not connected
* `Connecting` : the client is connecting (will transition to `Connected` if ok, to `NotConnected` otherwise)
* `Connected` : the client is connected (at least 1 connection to a member) (will transition to `Disconnected` if the last connection is lost, to `Disconnecting` if requested to shutdown)
* `Disconnected` : the client just lost its last connection (this is a transitory state, will transition to `Reconnecting` if configured to reconnect, to `NotConnected` otherwise)
* `Reconnecting` : the client, after having been disconnected, is trying to reconnect (will transition to `Connected` if ok, to `NotConnected` otherwise)
* `Disconnecting` : the client has been requested to shutdown (will transition to `NotConnected` when done)

Internally, lots of cleanup to make the relationship between the different part (services) of the cluster more clear, with no circular dependencies, clearly identified background tasks, etc. I have a draft document that details all this, which will be part of the documentation for devs to understand it all.


zpqrtbnk:
Java states in `LifeCycleEvent.java`:
```
STARTING (complete) ➝ STARTED (shut down) ➝ SHUTTING_DOWN (down) ➝ SHUTDOWN
(connected) ➝ CLIENT_CONNECTED (disconnected) ➝ CLIENT_DISCONNECTED
```
Others event there are for merging cluster members, connecting and disconnecting a client to the member - these are *member* events, not client events, ignoring.

This is the proposed C# diagram. Labels such as `Started` or `Connected` are *states*, and each `➝` represents a *transition*. On each transition, we raise a `StateChanged` event (which I can rename to *lifecycle something* if important?) and provide the new state in the event arguments.
```
Starting (start) ➝ Started (connect) ➝ (success) ➝ Connected
                                         (failure) ➝ Shutdown

Connected (last connection closed) ➝ Disconnected (up & reconnect) (success) ➝ Connected
                                                                    (failure) ➝ Shutdown
                                                   (down | !reconnect) ➝ Shutdown

anything but ShuttingDown or Shutdown (dispose) ➝ ShuttingDown (down) ➝ Shutdown

```
I think this now mimics the other clients' events closely?

asimarslan (approved):
"System.Threading.Tasks.Dataflow" would be removed.

src/Hazelcast.Net/Clustering/ClusterState.cs
```diff
@@ -180,10 +175,9 @@ public async ValueTask TransitionAsync(ConnectionState to, ConnectionState from
         /// connected. It may make sense to retry operations that fail, because they
         /// should succeed when the cluster is eventually connected.</para>
         /// </remarks>
-        public bool IsActive => ConnectionState == ConnectionState.Connected ||
-                                ConnectionState == ConnectionState.Connecting ||
-                                ConnectionState == ConnectionState.Disconnected ||
-                                ConnectionState == ConnectionState.Reconnecting;
+        public bool IsActive => ConnectionState == ConnectionState.Starting ||
```
mdumandag: Isn't `Started` also an active state?

mdumandag (approved):

PR #367, opened 2020-12-21
Rename .CAF() into .CfAwait()

zpqrtbnk:
The `.CAF()` method is a shortcut for `.ConfigureAwait(false)` but the name is not very nice + we want to implement more "configure await" methods with timeouts, etc. So, renaming it to `.CfAwait()`.


asimarslan (approved):

zpqrtbnk:
verify

PR #418, opened 2021-01-28
Infinite invocation timeout

### LINKED ISSUE #414: Infinite Invocation Timeout
Remove the invocation timeout - let invocations run forever, assume that the server will either respond eventually, or the connection will go down.

zpqrtbnk:
Fixes #414


zpqrtbnk:
Won't be merge as-is, will be re-submitted.

PR #419, opened 2021-02-01
Refactor invocation & connection timeouts

### LINKED ISSUE #415: Infinite Connect Timeout
Change the default timeout when connecting a new client to a server to be infinite.

zpqrtbnk:
Fixes #415 #414


src/Hazelcast.Net.Examples/Client/ContainerExample.cs
```diff
@@ -83,7 +83,7 @@ public static async Task Main(string[] args)
             // configure hazelcast (can do it multiple times..)
             services.Configure<HazelcastOptions>(options =>
             {
-                options.Networking.ConnectionTimeoutMilliseconds = 45_000;
+                options.Networking.TimeoutMilliseconds = 45_000;
```
asimarslan, line 86: the default value on java is `5_000`

src/Hazelcast.Net/Networking/NetworkingOptions.cs
```diff
@@ -130,8 +119,17 @@ private NetworkingOptions(NetworkingOptions other)
...
+        /// <remarks>
+        /// <para>Specifies the Hazelcast client connection parameters, including the timeout, i.e. the maximum
+        /// amount of time that the Hazelcast client can spend trying to establish a connection to the cluster
+        /// before failing. See <see cref="SocketOptions"/> for specifying the individual socket parameters,
+        /// including the individual socket connection timeout.</para>
+        /// </remarks>
         public RetryOptions ConnectionRetry { get; } = new RetryOptions();
 
+        // FIXME document
+        public int TimeoutMilliseconds { get; set; } = 30_000;
```
asimarslan, line 131: This relates to only connection-related timeout (although shared by others, all should be connection timeout)
so a `ConnectionTimeoutMilliseconds ` can be a better name.

Other clients use 5_000 ms as the default value.

src/Hazelcast.Net/Core/RetryOptions.cs
```diff
@@ -12,12 +12,10 @@
...
 
-using Hazelcast.Clustering;
-
 namespace Hazelcast.Core
 {
     /// <summary>
-    /// Represents a <see cref="RetryStrategy"/> configuration.
+    /// Represents the configuration for the retry strategy.
     /// </summary>
     public class RetryOptions
```
asimarslan, line 20: It should be renamed as `ConnectionRetryOptions`

src/Hazelcast.Net/Core/RetryOptions.cs
```diff
@@ -55,9 +53,9 @@ private RetryOptions(RetryOptions other)
         public double Multiplier { get; set; } = 1;
 
         /// <summary>
-        /// Gets or sets the cluster connection timeout in milliseconds.
+        /// Gets or sets the timeout in milliseconds.
         /// </summary>
-        public long ClusterConnectionTimeoutMilliseconds { get; set; } = 20*1000;
+        public long TimeoutMilliseconds { get; set; } = 20*1000;
```
asimarslan, line 58: Likewise, this is not a generic timeout. It's connection timeout, so I think the old name or `ConnectionTimeoutMilliseconds` is a better choice. 

The default value should be infinite, not 20 sec.

src/Hazelcast.Net/Networking/NetworkingOptions.cs
zpqrtbnk: oops forgot the FIXME, going to document now - and use proper name & default value

src/Hazelcast.Net/Core/RetryOptions.cs
zpqrtbnk: ok, I get it, I renamed it because IRetryOptions seemed generic enough that we may want to reuse it in places where it's not about connections... but... never mind, renaming

src/Hazelcast.Net/Core/RetryOptions.cs
zpqrtbnk: ok

src/Hazelcast.Net/Clustering/Invocation.cs
```diff
@@ -233,14 +231,20 @@ public async ValueTask<bool> CanRetryAsync(Func<long> correlationIdProvider)
...
             if (correlationIdProvider == null) throw new ArgumentNullException(nameof(correlationIdProvider));
 
-            if (_cancellationToken.IsCancellationRequested) return false;
+            // fast fail on cancel
+            if (_cancellationToken.IsCancellationRequested) 
+                return false;
+
+            // fast fail on timeout
+            if (Clock.Milliseconds - StartTime > _messagingOptions.RetryTimeoutSeconds * 1000)
+                return false;
```
mdumandag, line 240: Minor: On timeout, Java client completes the invocation with the special `OperationTimeoutException`, I guess here, we are completing it with whatever exception that caused this check

zpqrtbnk: true - fixing

src/Hazelcast.Net/Core/ConnectionRetryOptions.cs
```diff
@@ -55,9 +53,9 @@ private RetryOptions(RetryOptions other)
         public double Multiplier { get; set; } = 1;
 
         /// <summary>
-        /// Gets or sets the cluster connection timeout in milliseconds.
+        /// Gets or sets the timeout in milliseconds.
         /// </summary>
-        public long ClusterConnectionTimeoutMilliseconds { get; set; } = 20*1000;
+        public long ConnectionTimeoutMilliseconds { get; set; } = -1; // infinite
```
mdumandag, line 58: I think the ClusterConnectionTimeoutMiliseconds name (and its documentation) was better than the current state

src/Hazelcast.Net/Clustering/RetryStrategy.cs
```diff
@@ -93,14 +88,14 @@ public async ValueTask<bool> WaitAsync(CancellationToken cancellationToken)
 
             var elapsed = (int) (DateTime.UtcNow - _begin).TotalMilliseconds;
 
-            if (elapsed > _timeout)
+            if (elapsed > _timeoutMilliseconds)
```
mdumandag, line 91: Does that check work for the new default value (-1)? In Java, we set this to Long.MAX_VALUE if it is equal to -1 before constructing something similar to this class

mdumandag (approved):

asimarslan (approved):

PR #423, opened 2021-02-17
Implement IAtomicLong

zpqrtbnk:
*WARN!* This PR cannot build until this [client protocol PR](https://github.com/hazelcast/hazelcast-client-protocol/pull/374) has been reviewed and merged, as it depends on this protocol code.

This PR implements `IAtomicLong` and the very basic elements of the CP subsystem that are required for `IAtomicLong`.

Tests validate that it works.

Note that:
* The rest of the CP subsystem remains to be implemented. 
* None of the `Apply` and `Alter` methods (which accept a `IFunction`) are implemented.


src/Hazelcast.Net/HazelcastClient.Objects.cs
```diff
@@ -25,6 +26,12 @@ namespace Hazelcast
     internal partial class HazelcastClient // Distributed Objects
     {
         private readonly ISequence<long> _lockReferenceIdSequence = new Int64Sequence();
+        private ICP _cp;
+
+        /// <summary>
+        /// Gets the CP subsystem.
+        /// </summary>
+        public ICP CP => _cp ??= new CP.CP(Cluster, _distributedOjects);
```
asimarslan, line 34: All other clients name this as `CPSubsystem`, can we use the same naming here as well?

src/Hazelcast.Net/CP/ICP.cs
```diff
@@ -0,0 +1,35 @@
...
+// limitations under the License.
+
+using System.Threading.Tasks;
+
+namespace Hazelcast.CP
+{
+    /// <summary>
+    /// Defines the CP subsystem.
+    /// </summary>
+    public interface ICP
```
asimarslan, line 22: All other clients name this as `ICPSubsystem`, can we use the same naming here as well?

src/Hazelcast.Net/HazelcastClient.Objects.cs
zpqrtbnk: yes, renaming everywhere.

zpqrtbnk:
verify

hz.ps1
```diff
@@ -981,6 +1038,40 @@ if ($doTagVersion)
...
+    # the testing projects, which defines SERVER_CODEC.
+
+    # wipe existing codecs from the C# and protocol repositories
+    Write-Output "Clear codecs"
+    rm -force $srcDir/Hazelcast.Net/Protocol/Codecs/*.cs
+    rm -force $srcDir/Hazelcast.Net/Protocol/CustomCodecs/*.cs
+    rm -force -recurse $slnRoot/protocol/output/cs
+
+    Write-Output "Generate codecs"
+    python $slnRoot/protocol/generator.py -l cs --no-binary
```
asimarslan, line 1056: there is a `-r` option in the generator so that you won't need the below copy stage.
something like;
```
python $slnRoot/protocol/generator.py -l cs --no-binary -r $srcDir
```

hz.ps1
```diff
@@ -981,6 +1038,40 @@ if ($doTagVersion)
     }
 }
 
+# generate codecs
+if ($doCodecs) {
```
asimarslan, line 1042: the idea of codec generator is to not modify the codecs after generation.
This block of script still borrow some of the code generator responsibility.
I see a normalization stage below, if there is such need, it should be in the generator.

hz.ps1
```diff
@@ -981,6 +1038,40 @@ if ($doTagVersion)
...
+
+    # copy generated codecs to the C# repository
+    Write-Output "Copy codecs"
+    cp $slnRoot/protocol/output/cs/src/Hazelcast.Net/Protocol/Codecs/*.cs $srcDir/Hazelcast.Net/Protocol/Codecs/
+    cp $slnRoot/protocol/output/cs/src/Hazelcast.Net/Protocol/CustomCodecs/*.cs $srcDir/Hazelcast.Net/Protocol/CustomCodecs/
+
+    # normalize codecs FIXME platform?
+    Write-Output "Normalize codecs"
+    foreach ($file in $(ls $srcDir/Hazelcast.Net/Protocol/Codecs/*.cs)) {
+        clrSrc $file
```
asimarslan, line 1066: clrSrc function belongs to client-protocol repo's code generator.

hz.ps1
zpqrtbnk: excellent, will look into it

hz.ps1
zpqrtbnk: I get it

hz.ps1
zpqrtbnk: I get it

zpqrtbnk:
Have used your inputs to prepare this PR: https://github.com/hazelcast/hazelcast-client-protocol/pull/376
Which does everything at the Python level so all we have to do now is:
```
    Write-Output "Clear codecs"
    rm -force $srcDir/Hazelcast.Net/Protocol/Codecs/*.cs
    rm -force $srcDir/Hazelcast.Net/Protocol/CustomCodecs/*.cs

    Write-Output "Generate codecs"
    python $slnRoot/protocol/generator.py -l cs --no-binary -r $slnRoot
```
Can you have a look?

If OK I'll update this PR `hz.ps1` script to use it

hz.ps1
```diff
@@ -660,9 +705,16 @@ function ensureJar($jar, $repo, $artifact) {
     $script:classpath = $classpath
 }
 
-# say hello
-Write-Output "Hazelcast .NET Command Line"
-Write-Output ""
+# clear source file
+function clrSrc($file) {
```
asimarslan, line 709: dead code?

src/Hazelcast.Net.Tests/CP/AtomicLongTests.cs
```diff
@@ -0,0 +1,52 @@
...
+using Hazelcast.Testing;
+using NUnit.Framework;
+
+namespace Hazelcast.Tests.CP
+{
+    [TestFixture]
+    public class AtomicLongTests : SingleMemberClientRemoteTestBase
+    {
+        [Test]
+        public async Task Test()
```
asimarslan, line 25: Can you split the test cases?
see: https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/test/java/com/hazelcast/cp/internal/datastructures/atomiclong/AbstractAtomicLongBasicTest.java

src/Hazelcast.Net/CP/CPSubsystem.cs
```diff
@@ -0,0 +1,118 @@
...
+        //
+        // These objects are therefore IDistributedObject but *not* DistributedObjectBase, and *not*
+        // managed by the DistributedObjectFactory.
+        //
+        // The are destroyed via ClientProxy.destroy, which is getContext().getProxyManager().destroyProxy(this),
+        // which means they are destroyed by ProxyManager aka DistributedObjectFactory, which would try to
+        // remove them from cache (always missing) and end up doing proxy.destroyLocally() which eventually
+        // calls into the object's onDestroy() method.
+        //
+        // But... this is convoluted? For now, our objects inherit from CPObjectBase which is simpler than
```
mdumandag, line 54: +1
We also have separate base classes for the Python and Node.js clients to avoid this problem

src/Hazelcast.Net/HazelcastClient.Objects.cs
```diff
@@ -25,6 +26,12 @@ namespace Hazelcast
     internal partial class HazelcastClient // Distributed Objects
     {
         private readonly ISequence<long> _lockReferenceIdSequence = new Int64Sequence();
+        private ICPSubsystem _cpSubsystem;
+
+        /// <summary>
+        /// Gets the CP subsystem.
+        /// </summary>
+        public ICPSubsystem CPSubsystem => _cpSubsystem ??= new CPSubsystem(Cluster);
```
mdumandag, line 34: I am not familiar with C#, but is `??=` safe to use from multiple tasks? I mean, can I end up with multiple `CPSubsystem` objects from different tasks? Maybe this is not a problem as of now, since we don't have any state on CP subsystem, but once you implement FencedLock, it might be problematic

mdumandag (commented):
Looks good to me as of now with small comments except the tests.

Also, will you send documentation and code samples in a later PR?

hz.ps1
zpqrtbnk: not really - from time to time I normalize the whole codebase - so I'd rather keep the method around and one day I'll update the hz script with some sort of "normalize" function

src/Hazelcast.Net.Tests/CP/AtomicLongTests.cs
zpqrtbnk: yes, splitting

hz.ps1
```diff
@@ -981,6 +1045,25 @@ if ($doTagVersion)
     }
 }
 
+# generate codecs
+if ($doCodecs) {
+
+    # note: codec files contain the client-side methods, along with the
```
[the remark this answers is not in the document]
zpqrtbnk: certainly not planning to rewrite Hazelcast in C# - it's just that for some tests, I simulate the server and need to be able to generate responses, etc - so I have updated the templates to generate both client and server code - but, there is an `#ifdef` of some sort to ensure that the released DLL only contains the server-side methods

src/Hazelcast.Net.Tests/CP/AtomicLongTests.cs
```diff
@@ -0,0 +1,52 @@
...
+// limitations under the License.
+
+using System.Threading.Tasks;
+using Hazelcast.Testing;
+using NUnit.Framework;
+
+namespace Hazelcast.Tests.CP
+{
+    [TestFixture]
+    public class AtomicLongTests : SingleMemberClientRemoteTestBase
```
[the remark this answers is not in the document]
zpqrtbnk: can we do this when we implement the full CP subsys?

src/Hazelcast.Net/HazelcastClient.Objects.cs
zpqrtbnk: `??=` is supposed to be safe, yes - and that code is internal and can change later on - one question I had though is... do we agree just creating the `ICPSubSystem` instance is *always* a local operation that does not involve the server? I am using a property here so it has to be synchronous. My understanding of the Java code is that it is OK, and it's only when you create a FencedLock for instance, that you talk to the server. Can you confirm this?

src/Hazelcast.Net.Tests/CP/AtomicLongTests.cs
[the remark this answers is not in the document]
mdumandag: I think it would be good to have them even for the atomic long tests

src/Hazelcast.Net/HazelcastClient.Objects.cs
mdumandag: It should be OK, creating CP Subsystem is always local, operations on it might talk with the server as you said

src/Hazelcast.Net.Tests/CP/AtomicLongTests.cs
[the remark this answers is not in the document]
asimarslan: Yes, we need CP enabled.

src/Hazelcast.Net.Tests/CP/AtomicLongTests.cs
zpqrtbnk: pushed a commit that splits tests

zpqrtbnk:
verify

asimarslan (approved):

Issue #422, opened 2021-02-13
Strong name error in ASP.NET application


andy-dev947:
Steps to reproduce

- Create a new ASP.NET Web Application (use WebAPI template and .NET 4.7)
- Install Hazelcast.Net 4.0.0
- Run the application
- You will get the following error:
 `Could not load file or assembly 'Hazelcast.Net' or one of its dependencies. Strong name signature could not be verified.  The assembly may have been tampered with, or it was delay signed but not fully signed with the correct private key. (Exception from HRESULT: 0x80131045)`

Note this does not happen with 3.12.3.

Installing 4.0.0 in other types of applications (Console, etc.) seems to work fine

andy-dev947:
Looks like there has been a similar issue in the past which might help: #195

zpqrtbnk:
Investigating...

When you write "install Hazelcast.Net 4.0.0" is it the NuGet-provided package, or a custom-built version?

andy-dev947:
Hi,

Yes, that's via nuget.

Thanks,

zpqrtbnk:
Just to let you know that I have repro the issue and I am currently working on it.

zpqrtbnk:
just to let you know that version 4.0.1 has been released with a fix for this issue - let me know if it works!

andy-dev947:
Hi,

Thanks - this works now.

PR #428, opened 2021-03-05
Fix and cleanup the build process

zpqrtbnk:
Fixes: #422

This fixes our build process, including the assembly signing process, thus fixing #422 which prevented using the Hazelcast.Net.dll in ASP.NET applications.


zpqrtbnk:
verify

asimarslan (approved):

PR #437, opened 2021-03-24
Adding new referances

alisengul53:
net462 referances added to Hazelcast.Net.Test, Hazelcast.Net.Examples, Hazelcast.Net.Testing


zpqrtbnk:
I *think* there are files in this PR that should not be there, for instance the documentation files. Can you have a look?

PR #441, opened 2021-04-09
Bugfix the Cluster code [API-236]

### LINKED ISSUE #429: Reconnections in 4.0 [API-236]
Hi

I just wanted to check what the expected behaviour is when the member a client is connected to is shutdown.  From my observations is seems that the client gets disconnected and will then never get reconnected again even if there are other members in the cluster still online.  The client just repeatedly throws ClientOfflineException.  The only way to recover seems to be to dispose the old client and then start a completely new connection using StartNewClientAsync.

If that is the case then what is the best way to detect when the client is in an unrecoverrable state? 

Also, I know the Java client can auto reconnect but I don't think this functionality works in the C# client even though the options are there?

Thanks

zpqrtbnk:
This is a general bug-fixing & refactoring of our Cluster code (the code which handles connections, retries, etc).
Fixes various reported and unreported issues.

Fixes #429 

This giant PR will not get merged "as-is", instead it is decomposed in the following PRs, which need to be merged in order:

- [x] #447 Improvements to testing utilitites
- [x] #448 Add Client reconnection test (issue #429)
- [x] #449 Cleanup load-balancing
- [x] #450 Cleanup misc core classes
- [x] #451 Cleanup core exceptions
- [x] #452 Main refactoring of Cluster code 👹 
- [ ] #455 Code Analysis and Quality
- [ ] #457 Logging and Testing
- [ ] #458 Projects Cleanup
- [ ] #459 Cluster Updates for Cloud

And then *this* PR can be closed ***without merging***.


src/Hazelcast.Net.Testing/Protocol/ErrorsServerCodec.cs
```diff
@@ -33,7 +33,7 @@ internal static class ErrorsServerCodec
         public static ClientMessage EncodeResponse(IEnumerable<ErrorHolder> errorHolders)
         {
             var clientMessage = new ClientMessage();
-            var initialFrame = new Frame(new byte[InitialFrameSize]);
+            var initialFrame = new Frame(new byte[InitialFrameSize], (FrameFlags)ClientMessageFlags.Unfragmented);
```
alexb5dh, line 36: Could you please explain what this change do?

zpqrtbnk:
arh, was closed automatically when the branch was deleted - ok

PR #443, opened 2021-04-09
Documentation for CPDistributedObjectBase.DisposeAsync

alexb5dh:


zpqrtbnk:
Can you elaborate on "_as this seems like an expected behavior_"? What led you to that conclusion?

There is no such auto-destroy in the Java code + it would be pretty bad in a multi-client situation if a client that terminates and disposes its instances, globally destroys them?

alexb5dh:
You're right, didn't think about multiple clients,  
In that case - what's the idea behind empty `DisposeAsync`? Is this is intended for future compatibility - so users will wrap calls to it in `using` and later we may change this method to do some cleanup job? In such case, I would suggest to add this to the summary docs and provide synchronous `Dispose` implementation.

zpqrtbnk:
>In that case - what's the idea behind empty DisposeAsync? Is this is intended for future compatibility - so users will wrap calls to it in using and later we may change this method to do some cleanup job? 

Sort-of. The idea is that we can eventually do things in this method, without having to introduce the method (which would be a breaking change).

>In such case, I would suggest to add this to the summary docs and provide synchronous Dispose implementation.

Happy to add a comment saying... this does nothing for now but may do something in the future. And it's async because it may have to do async stuff that a synchronous `Dispose` could not. So... not sure about the synchronous `Dispose`.

What do you think?

alexb5dh:
> And it's async because it may have to do async stuff that a synchronous Dispose could not. So... not sure about the synchronous Dispose.
It's an [MS recommendation](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync) and I'm not sure if it's possible to somehow apply `using` operator if only `IAsyncDisposable` is implemented and method is synchronous for whatever reason (which is a quite possible scenario).

zpqrtbnk:
To answer your question, yes when only `IAsyncDisposable` is implemented you can use the `await using` pattern:

```
// assume GetAsyncDisposable returns an IAsyncDisposable
await using var asyncDisposable = await GetAsyncDisposable().ConfigureAwait(false);
```

So I'd prefer if we stick with `IAsyncDisposable`.

alexb5dh:
I just would point out that a lot of old projects still partially of fully relies on synchronous code and it quite a complex task for them to switch to async. If we leave `IAsyncDisposable` only, they won't be able to use CP objects without a workaround of declaring an async method and synchronously waiting for it, where they won't do this otherwise.

zpqrtbnk:
Well, they just cannot use the new client without going async anyways, can they?
You are right about it being a ... big change ... for old projects, but they cannot really avoid it.
So... async all the things!

alexb5dh:
Ok. In this case this PR only contains basic documentation for `CPDistributedObjectBase.DisposeAsync`. Feel free to check it or close   .

zpqrtbnk (approved):
Approved! So the idea is that _you_ now merge the PR.

PR #445, opened 2021-04-09
`AtomicReference` implementation [API-254] and documentation [API-279]

alexb5dh:


src/Hazelcast.Net/CP/AtomicRef.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        /// <inheritdoc />
+        public override async ValueTask DestroyAsync()
+        {
+            var requestMessage = CPGroupDestroyCPObjectCodec.EncodeRequest(CPGroupId, ServiceName, Name);
+            var responseMessage = await Cluster.Messaging.SendAsync(requestMessage).CfAwait();
+            var response = CPGroupDestroyCPObjectCodec.DecodeResponse(responseMessage);
+        }
+
+        protected IData ToData(T value) => SerializationService.ToData(value);
```
zpqrtbnk, line 87: Maybe we need to move this at `CPDistributedObjectBase` level?

alexb5dh: Thought about this but wanted to postpone `till more CP objects will be implemented as I'm not sure all of them relies on the same destroy API. What do you think?

zpqrtbnk: Ah, was talking about the `ToData` etc thing.

alexb5dh: This will require to move `SerializationService` instance to the `CPDistributedObjectBase`, but it looks like none of the CP objects except `AtomicReference` need it.

doc/dev/doc/atomicref.md
```diff
@@ -0,0 +1,35 @@
+# AtomicReference
+
+> [!NOTE]
+> IAtomicReference is a member of CP Subsystem API. For detailed information, see the [CP SubSystem documentation](cpsubsystem.md).
+
+Hazelcast @Hazelcast.CP.IAtomicReference is the distributed implementation of `java.util.concurrent.atomic.AtomicReference` and offers most of its operations such as Get, Set, GetAndSet and CompareAndSet. You can also think of GetAndSet, CompareAndSet as @System.Interlocked Exchange and CompareExchange methods for distributed values. Since @Hazelcast.CP.IAtomicReference is a distributed implementation, these operations involve remote calls and thus their performances differ from local, in-memory, references.
+
+The following example code adds a dot the end of any string added to the shared reference:
+
+```csharp
```
alexb5dh, line 10: Example looks too complex, so will appreciate any better ideas.

src/Hazelcast.Net/CP/AtomicReference.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.Threading.Tasks;
+using Hazelcast.Clustering;
+using Hazelcast.Core;
+using Hazelcast.DistributedObjects;
+using Hazelcast.Protocol.Codecs;
+using Hazelcast.Serialization;
+
+namespace Hazelcast.CP
+{
+    internal class AtomicReference<T>: CPDistributedObjectBase, IAtomicReference<T>
```
alexb5dh, line 11: Removed `where T: class` requirement as both Java and JS client support simple primitive types like `int`, `float`, etc.  
The downside is that `await aref.ContainsAsync(await aref.GetAsync());` returns false if invoked with default value, as it's `null` on the server but `default(T)` on the client.

zpqrtbnk: That would be consistent with the Java client. It needs to be documented in `GetAsync` and `ContainsAsync` documentation (the `<remarks>` part).

doc/dev/doc/atomicref.md
zpqrtbnk: Let's say our "clients" share a common value (that could be a hash or secret of some sort), which needs to be updated periodically, and the new value *must* derive from the previous one (whatever that means), so from periodically each client would run the following code, and only 1 client would actually update the value:
```public async Task UpdateValue(IAtomicReference<string> aref)
{
    var value = await aref.GetAsync();
    var newValue = DeriveNewValue(value);
    await aref.CompareAndSetAsync(value, newValue);
}```
Not sure it makes much sense, but it's cleaner that the proposed example?

src/Hazelcast.Net/CP/AtomicReference.cs
alexb5dh: Added remarks for `GetAsync` and `GetAndSetAsync`. Not sure if `ContainsAsync` need any changes.

zpqrtbnk (approved):
sounds good now, good for merge

PR #455, opened 2021-04-30
Code Analysis and Quality

zpqrtbnk:
This is a code analysis / quality PR - it should *_not_* change anything, functionally, but only adjust things to get rid of most of the warnings that we get when building code. The goal being to have zero warning, zero message when we build.

It also upgrades `dotCover` (that we use for code coverage) to its latest version, which brings improvements such as properly supporting `switch` expressions.

It also adds a special Roslyn analyzer that converts comments containing the `FIXME` keyword into build-time warnings or errors. This is because I like putting `FIXME` comments in places where I know that we _*must_* come back and fix something, before we ship, but then I lose track of them.

### DIFF
--- a/src/Hazelcast.Net/CP/CPSubsystem.cs
+++ b/src/Hazelcast.Net/CP/CPSubsystem.cs
@@ -59,14 +59,14 @@ namespace Hazelcast.CP
         public async Task<IAtomicLong> GetAtomicLongAsync(string name)
         {
             var (groupName, objectName) = ParseName(name);
-            var groupId = await GetGroupIdAsync(groupName, objectName).CfAwait();
+            var groupId = await GetGroupIdAsync(groupName).CfAwait();
 
             return new AtomicLong(objectName, groupId, _cluster);
         }
 
         // see: ClientRaftProxyFactory.java
 
-        private async Task<CPGroupId> GetGroupIdAsync(string proxyName, string objectName)
+        private async Task<CPGroupId> GetGroupIdAsync(string proxyName)
         {
             var requestMessage = CPGroupCreateCPGroupCodec.EncodeRequest(proxyName);
             var responseMessage = await _cluster.Messaging.SendAsync(requestMessage).CfAwait();
@@ -93,7 +93,7 @@ namespace Hazelcast.CP
             }
             else
             {
-                groupName = name.Substring(pos + 1).Trim();
+                groupName = name[(pos + 1)..].Trim();
                 if (groupName.Equals(DefaultGroupName, StringComparison.OrdinalIgnoreCase))
                     groupName = DefaultGroupName;
             }
@@ -101,7 +101,11 @@ namespace Hazelcast.CP
             if (groupName.Length == 0)
                 throw new ArgumentException("CP group name cannot be an empty string.", nameof(name));
 
+#if NETSTANDARD2_1
+            if (groupName.Contains("@", StringComparison.OrdinalIgnoreCase))
+#else
             if (groupName.Contains("@"))
+#endif
                 throw new ArgumentException("CP group name must be specified at most once.", nameof(name));
 
             if (groupName.Equals(MetaDataGroupName, StringComparison.OrdinalIgnoreCase))
--- a/src/Hazelcast.Net/Clustering/StateChangeQueue.cs
+++ b/src/Hazelcast.Net/Clustering/StateChangeQueue.cs
@@ -93,7 +93,7 @@ namespace Hazelcast.Clustering
             var completion = new TaskCompletionSource<object>();
             do
             {
-                marker = -RandomProvider.Random.Next();
+                marker = -RandomProvider.Next();
             } while (marker < 0 && !_markers.TryAdd(marker, completion));
 
             // queue the state
@@ -162,6 +162,8 @@ namespace Hazelcast.Clustering
 
             // wait until the events queue is drained
             await _raising.CfAwait();
+
+            await _states.DisposeAsync().CfAwait();
         }
     }
 }
--- a/src/Hazelcast.Net/CodeAnalysisSuppressions.cs
+++ b/src/Hazelcast.Net/CodeAnalysisSuppressions.cs
@@ -14,11 +14,14 @@
 
 using System.Diagnostics.CodeAnalysis;
 
-#if CODE_ANALYSIS
-
-// this file contains assembly-level SuppressMessage attributes for NDepend,
+// this file contains assembly-level suppression directives for NDepend,
 // they will be compiled into the assembly only if CODE_ANALYSIS is defined
 // and should not be compiled into the final Release version that ships.
+//
+// this file also contains assembly-level suppression directives for the
+// compiler (to deal with some warnings) that are always compiled.
+
+#if CODE_ANALYSIS
 
 // scope can be: deep module namespace type method field
 // and also: namespaceAndDescendants
@@ -58,7 +61,7 @@ using System.Diagnostics.CodeAnalysis;
 // Unfortunately, NDepend uses e.g. 'method' scope which is not OK
 #pragma warning disable IDE0076 // Invalid global 'SuppressMessageAttribute'
 
-#region General
+#region NDepend / General
 
 // NDepend complains about 'public' methods in an 'internal' class
 // but even NDepend documentation mentions that not everyone agrees
@@ -70,7 +73,7 @@ using System.Diagnostics.CodeAnalysis;
 
 #endregion
 
-#region Accepted Checks
+#region NDepend / Accepted Checks
 
 // We use System.Random in our RandomProvider and don't expect it
 // to be used for anything security-related, so, suppress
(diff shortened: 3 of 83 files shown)


src/Hazelcast.Net/Core/TaskCoreExtensions.cs
```diff
@@ -238,7 +238,7 @@ private static async Task AwaitWithTimeout(Task task, int timeoutMilliseconds, C
         {
             if (timeoutMilliseconds < 0)
             {
-                await task;
+                await task.ConfigureAwait(false);
```
alexb5dh, line 241: Is `CfAwait` not available here?  
+line 280.

src/Hazelcast.Net/Events/DistributedObjectLifecycleEventArgs.cs
```diff
@@ -19,7 +19,9 @@ namespace Hazelcast.Events
     /// <summary>
     /// Provides a base class for cluster object lifecycle events data.
     /// </summary>
+#pragma warning disable CA1711 // Identifiers should not have incorrect suffix - here it is correct
```
alexb5dh, line 22: Too many classes having this, maybe disable on assembly-level or in .csproj?

src/Hazelcast.Net/Serialization/PortableContext.cs
```diff
@@ -47,23 +47,18 @@ public IClassDefinition LookupClassDefinition(int factoryId, int classId, int ve
...
-            var factoryId = @in.ReadInt();
-            var classId = @in.ReadInt();
-            var version = @in.ReadInt();
-            var classDefinition = LookupClassDefinition(factoryId, classId, version);
-            if (classDefinition == null)
-            {
-                classDefinition = ReadClassDefinition(@in, factoryId, classId, version);
-            }
-            return classDefinition;
+            if (!data.IsPortable)throw new ArgumentException("Data is not Portable.", nameof(data));
```
alexb5dh, line 52:
```suggestion
            if (!data.IsPortable) throw new ArgumentException("Data is not Portable.", nameof(data));
```

src/Hazelcast.Net/Exceptions/ClientOfflineException.cs
```diff
@@ -57,7 +57,7 @@ public ClientOfflineException(ClientState state)
         /// </summary>
         /// <param name="message">The message that describes the error.</param>
         [ExcludeFromCodeCoverage]
-        private ClientOfflineException(string message)
+        public ClientOfflineException(string message)
```
alexb5dh, line 60: Why making all public now?

src/Hazelcast.Net/CP/CPSubsystem.cs
```diff
@@ -93,15 +93,19 @@ public static (string groupName, string objectName) ParseName(string name)
...
+                groupName = name[(pos + 1)..].Trim();
                 if (groupName.Equals(DefaultGroupName, StringComparison.OrdinalIgnoreCase))
                     groupName = DefaultGroupName;
             }
 
             if (groupName.Length == 0)
                 throw new ArgumentException("CP group name cannot be an empty string.", nameof(name));
 
+#if NETSTANDARD2_1
+            if (groupName.Contains("@", StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
zpqrtbnk: because 2.1 made it available to specify the comparison, and then the recommendation is to specify it

src/Hazelcast.Net/CodeAnalysisSuppressions.cs
```diff
@@ -191,6 +194,115 @@
...
+// nothing we can do about these
+[assembly: SuppressMessage(Analysis.Naming, Analysis.CA1711, Scope = Analysis.Scope.Type,
+    Target = "~T:Hazelcast.DistributedObjects.ITopicEventHandler`1",
+    Justification = Analysis.Justification.CA1711_EventHandler)]
+[assembly: SuppressMessage(Analysis.Naming, Analysis.CA1711, Scope = Analysis.Scope.Type,
+    Target = "~T:Hazelcast.DistributedObjects.IMapEventHandler`3",
+    Justification = Analysis.Justification.CA1711_EventHandler)]
+[assembly: SuppressMessage(Analysis.Naming, Analysis.CA1711, Scope = Analysis.Scope.Type,
+    Target = "~T:Hazelcast.DistributedObjects.IMapEntryEventHandler`3",
+    Justification = Analysis.Justification.CA1711_EventHandler)]
```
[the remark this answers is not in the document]
zpqrtbnk: i'd rather have them all in the same file and not pollute the code with them

src/Hazelcast.Net/Events/DistributedObjectLifecycleEventArgs.cs
zpqrtbnk: yes - looking into it

src/Hazelcast.Net/Serialization/PortableContext.cs
zpqrtbnk: yes

src/Hazelcast.Net/Exceptions/ClientOfflineException.cs
zpqrtbnk: because we want to let users catch it in some cases

src/Hazelcast.Net/Core/TaskCoreExtensions.cs
zpqrtbnk: yes, looking into it

zpqrtbnk: this file defines `CfAwait` so as a general rule I don't use it in the file itself

src/Hazelcast.Net/Clustering/StateChangeQueue.cs
```diff
@@ -162,6 +162,8 @@ public async ValueTask DisposeAsync()
 
             // wait until the events queue is drained
             await _raising.CfAwait();
+
+            await _states.DisposeAsync().CfAwait();
```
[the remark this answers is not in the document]
zpqrtbnk: because it's the async queue that `StateChangeQueue` owns so it needs to dispose it when it is itself disposed?

src/Hazelcast.Net/Events/DistributedObjectLifecycleEventArgs.cs
zpqrtbnk: ok they _are_ already disabled at assembly level and these pragmas should not be here anymore, cleaning

src/Hazelcast.Net/Clustering/StateChangeQueue.cs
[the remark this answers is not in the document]
alexb5dh: Though it was related do another method because of GitHub highlighting.

src/Hazelcast.Net/CP/CPSubsystem.cs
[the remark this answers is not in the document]
alexb5dh: What if we define `ContainsNoCase` extension method, for 2.1 forward it to framework implementation, otherwise use something [like this](https://stackoverflow.com/a/63371269)?

I think extension methods like `ContainsNoCase`, `IndexOfInvariant`, etc. are simpler than adding `StringComparison` parameter every time. In our case it will also help as of adding this `#if` blocks everywhere.

src/Hazelcast.Net/CodeAnalysisSuppressions.cs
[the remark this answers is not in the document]
alexb5dh: What will happen if type is renamed or moved to another namespace? It doesn't look like Visual Studio or ReSharper will pick that up for `SuppressMessage.Target`.

src/Hazelcast.Net/Exceptions/ClientOfflineException.cs
alexb5dh: But exception was public before, change is about constructors here.

zpqrtbnk: ah, I answered to quickly. when there isn't a `public WhateverException(string message)` we get a warning because the "standard" for exception is to have one such constructor. `private` is not good. so I've made it public, does not hurt and gets rid of a warning.

src/Hazelcast.Net/CP/CPSubsystem.cs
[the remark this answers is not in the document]
zpqrtbnk: check what I've done, it was done for other methods: there is a `StringNetStandardExtensions` class that defines the missing methods for NetStandard 2.0 - so you can code just as if you had 2.1 - so your code is future proof. what do you think?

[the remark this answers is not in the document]
alexb5dh: So you plan add `Contains` to `StringNetStandardExtensions` also?

alexb5dh (approved):

PR #475, opened 2021-06-24
[API-274] SQL for Map ([API-617], [API-616], [API-431], [API-619], [API-620], [API-432], [API-659], [API-660], [API-661])

alexb5dh:
Requires:
- #492 
- #495
- #493


src/Hazelcast.Net/HazelcastClient.Objects.cs
```diff
@@ -28,6 +30,9 @@ internal partial class HazelcastClient // Distributed Objects
         private readonly ISequence<long> _lockReferenceIdSequence = new Int64Sequence();
         private ICPSubsystem _cpSubsystem;
 
+        // FIXME [Oleksii] discuss if lazy is needed
```
alexb5dh, line 33: Want to check if lazy initialization is needed here. `SqlService` is a simple relatively lightweight class, so probaly not. On the other hand `CPSubsystem` is similar but uses (thread-unsafe) lazy initialization in it's accessor.

src/Hazelcast.Net/Protocol/Codecs/SqlExecuteCodec.cs
```diff
@@ -0,0 +1,193 @@
...
+            initialFrame.Bytes.WriteByteL(RequestExpectedResultTypeFieldOffset, expectedResultType);
+            clientMessage.Append(initialFrame);
+            StringCodec.Encode(clientMessage, sql);
+            ListMultiFrameCodec.EncodeContainsNullable(clientMessage, parameters, DataCodec.Encode);
+            CodecUtil.EncodeNullable(clientMessage, schema, StringCodec.Encode);
+            SqlQueryIdCodec.Encode(clientMessage, queryId);
+            return clientMessage;
+        }
+
+#if FALSE // FIXME [Olesii] discuss whether to remove these parts and how to update protocol
```
alexb5dh, line 124: Several members like `DecodeRequest` and `EncodeResponse` are not needed for these codecs and methods they are using are not implemented. Wanted to remove them, but that would make this different from code obtained from generate-codecs operation. At the moment, I don't see any option for C# generator to skip server-side methods for the codec. Should I update the protocol to support this or just remove this code and make it a bit different from auto-generate one?

src/Hazelcast.Net/HazelcastClient.Objects.cs
zpqrtbnk: I think lazy initialization is good just to try to create as few objects as possible when we don't need them. And yes, the `CPSubsystem` initialization is not thread-safe because the class is stateless so it's not that important. But maybe it's cleaner to make it thread-safe with a lock, or a `Lazy<T>`, and do the same for `SqlService`?

src/Hazelcast.Net/Protocol/Codecs/SqlExecuteCodec.cs
zpqrtbnk: For other codecs, `DecodeRequest` and `EncodeResponse` are generated but within `#if SERVER_CODEC` blocks, so that they are *not* compiled in the main DLL but can be compiled in the testing DLL in case we want to simulate a server - which we do in some tests.

I think it is important that the SQL codecs are generated just like the others (via the generator), not custom.

What happens with these methods when the are generated, for SQL? Is the generated code not compilable? not runnable?

src/Hazelcast.Net/Sql/SqlResult.cs
```diff
@@ -0,0 +1,123 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Hazelcast.Core;
+using Hazelcast.Serialization;
+
+namespace Hazelcast.Sql
+{
+    // FIXME [Oleksii] Clarify if should be thread safe
+    public class SqlResult : IAsyncEnumerator<SqlRow>
```
alexb5dh, line 24: I didn't make `SqlResult` implement `IAsyncEnumerable`, but `IAsyncEnumerator` instead. The reason is that it doesn't fully match the idea behind enumerable, which usually can be iterated over any number of times via calling `GetEnumerator` . Unfortunately, `SqlResult` can only be iterated once. To highlight this behavior for users, I implemented it as `IAsyncEnumerator` and added `EnumerateOnce` method that will return one-time use `IAsyncEnumerable`.

alexb5dh: @zpqrtbnk I think `EnumerateOnce` make if more straightforward that result is a one-time use object. In most cases if developer sees an `IEnumerable` and haven't yet checked the documentation he assumes it can be enumerated until it's cleaned up or disposed.

src/Hazelcast.Net/HazelcastClient.Objects.cs
alexb5dh: Well, `Lazy<T>` is an object 😉. And straightforward initialization approach currently used by `CPSubsystem` may lead to several instances be created. By the way, `SqlService` is also stateless, at least the same way as `CPSubsystem` is.

src/Hazelcast.Net/Protocol/Codecs/SqlExecuteCodec.cs
alexb5dh: It's became not compilable, because I didn't provide `SqlPage.Encode` method. I can add it but throw `NotSupportedException` inside, what do you think?

src/Hazelcast.Net/HazelcastClient.Objects.cs
alexb5dh: Agreed on creating them immediately.

src/Hazelcast.Net/Protocol/Codecs/SqlExecuteCodec.cs
alexb5dh: Agreed on approach above.

src/Hazelcast.Net/Sql/SqlResult.cs
alexb5dh: Will go with `EnumerateOnce` for now but leave this for open for discussion for now and/or possible change in future versions.

src/Hazelcast.Net/Sql/SqlResultType.cs
```diff
@@ -0,0 +1,40 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Runtime.Serialization;
+
+namespace Hazelcast.Sql
+{
+    // FIXME [Oleksii] check if [Flags] will work
+    public enum SqlResultType
```
mdumandag, line 20: Do we need this enum anymore, since we have separate methods for row and update set queries?

src/Hazelcast.Net.Tests/Core/ByteExtensionsTests.cs
```diff
@@ -417,6 +417,21 @@ public void ReadGuid()
...
+        }
+
+        [Test]
+        [TestCase(new byte[]{0, 0x0C, 0x11, 0x10, 0x00, 0x00, 0x00, 0x00}, 1, "12:17:16")]
+        [TestCase(new byte[]{0, 0x15, 0x3B, 0x38, 0x4E, 0x61, 0xBC, 0x00}, 1, "21:59:56.012345678")]
+        public void ReadLocalTime(byte[] bytes, int position, string expected)
+        {
+            Assert.AreEqual(expected, bytes.ReadLocalTime(position));
+        }
+
```
mdumandag, line 434: Maybe test LocalDateTime and OffsetDateTime as well

src/Hazelcast.Net/HazelcastClient.Objects.cs
```diff
@@ -189,5 +190,8 @@ Task<IHRingBuffer<T>> GetRingBufferAsync<T>(string name)
             return await task.CfAwait();
 #endif
         }
+
+        /// <inheritdoc />
+        public Task<ISqlService> GetSqlServiceAsync() => Task.FromResult<ISqlService>(_sqlService);
```
mdumandag, line 195: Why is this getter async? We don't perform any invocations while constructing the SQL?

Also a note, this is called just `Sql` in the Java, not `SqlService`.

src/Hazelcast.Net/Sql/ISqlService.cs
```diff
@@ -0,0 +1,24 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Threading.Tasks;
+
+namespace Hazelcast.Sql
+{
+    public interface ISqlService
+    {
+        Task<ISqlQueryResult> ExecuteQueryAsync(string sql, object[] parameters = null, SqlStatementOptions options = null);
```
mdumandag, line 21: I am not familiar with .NET, but why this method is async? I mean, yes we make an invocation to the server, but we are not waiting for it and return an SqlResult immediately, right?

src/Hazelcast.Net/Sql/ISqlService.cs
```diff
@@ -0,0 +1,24 @@
...
+        Task<long> ExecuteCommandAsync(string sql, object[] parameters = null, SqlStatementOptions options = null);
```
mdumandag, line 22: If we return long here, how can the user cancel (close) the query? I think we need to return some kind of SqlResult here as well

src/Hazelcast.Net.Tests/Core/ByteExtensionsTests.cs
alexb5dh: Yes, I plan to add serialization/deserialization test for all supported types.

src/Hazelcast.Net/HazelcastClient.Objects.cs
alexb5dh: Added this in case if future versions will require some async operation to initialize the service. Now not sure it's worth it, can do it similar to `HazelcastClient.CPSubsystem` property. @zpqrtbnk what do you think?

Regarding the naming - will update.

src/Hazelcast.Net/Sql/ISqlService.cs
alexb5dh: In current implementation this will return it immediately, this is also for future compatibility in case we will need do some querying before creating `SqlResult`.

src/Hazelcast.Net/Sql/ISqlService.cs
alexb5dh: Will add `CancellationToken` parameter.

src/Hazelcast.Net/Sql/ISqlService.cs
```diff
@@ -0,0 +1,25 @@
...
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Hazelcast.Sql
+{
+    public interface ISqlService
+    {
+        Task<ISqlQueryResult> ExecuteQueryAsync(string sql, object[] parameters = null, SqlStatementOptions options = null);
+        Task<long> ExecuteCommandAsync(string sql, object[] parameters = null, SqlStatementOptions options = null, CancellationToken cancellationToken = default);
```
mdumandag, line 23: So, how do we cancel the initial request, if it is taking too long?

src/Hazelcast.Net/Sql/HLocalDate.cs
```diff
@@ -0,0 +1,140 @@
...
+
+using System;
+using System.Text.RegularExpressions;
+
+namespace Hazelcast.Sql
+{
+    /// <summary>
+    /// Represents Hazelcast SQL <c>DATE</c> type corresponding to <c>java.time.LocalDate</c> in Java.
+    /// </summary>
+    public readonly struct HLocalDate : IEquatable<HLocalDate>
```
alexb5dh, line 23: @zpqrtbnk 
Added deserialization for date/time types to be returned from SQL queries. Now, regarding serialization - to pass them as query parameters...

For now parameters use default serializers (built-in and user-provided), so simplest approach would be to register `IStreamSerializer`s for new types in `ConstantSerializerDefinitions`. However I'm not sure if that's reasonable as I haven't encountered them being used outside of our new SQL feature.

Another question that follows is regarding built-in serialization for `DateTime` and `DateTimeOffset` (plus `DateOnly`, `TimeOnly` in later versions):
- users would definitely like this to be supported out-of-the-box;
- `HLocalDateTime`, `HOffsetDateTime` (`HLocalDate`, `HLocalTime`) contains all allowed range of values for `DateTime`, `DateTimeOffset` (`DateOnly`, `TimeOnly`).

As such, it probably makes sense to support it and just forward call to corresponding `H*` serializer. But before doing this, I want to ensure that there are no other places where date/time types are used yet serialized in a different way.

alexb5dh, line 23: Another question that follows is regarding built-in serialization for `DateTime` and `DateTimeOffset` (plus `DateOnly`, `TimeOnly` in later versions):
- users would definitely like this to be supported out-of-the-box;
- `HLocalDateTime`, `HOffsetDateTime` (`HLocalDate`, `HLocalTime`) contains all allowed range of values for `DateTime`, `DateTimeOffset` (`DateOnly`, `TimeOnly`).

As such, it probably makes sense to support it and just forward call to corresponding `H*` serializer. But before doing this, I want to ensure that there are no other places where date/time types are used yet serialized in a different way.

src/Hazelcast.Net/Sql/SqlColumnType.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// <c>REAL</c> type, represented by <c>java.lang.Float</c> in Java and by <see cref="float"/> in .NET.
+        /// </summary>
+        Real = 7,
+
+        /// <summary>
+        /// <c>DOUBLE</c> type, represented by <c>java.lang.Double</c> in Java and by <see cref="double"/> in .NET.
+        /// </summary>
+        Double = 8,
+
+        // FIXME [Oleksii] discuss year range in HZ SQL and Java
```
alexb5dh, line 68: @mdumandag 
According to [docs](https://docs.hazelcast.com/imdg/4.2/sql/data-types.html) `DATE` type is backed by `java.lang.LocalDate` which supports [year range of ±999,999,999](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html#MIN). However, it's serialized as short (2 bytes) so real supported values ranges from -32768 to 32767.

For now I limited custom `HLocalDate` with second option to avoid data loss, but this probably need to be clarified in .NET and maybe IMDG documentation.

src/Hazelcast.Net/Sql/SqlQueryResult.cs
```diff
@@ -0,0 +1,135 @@
...
+            EnsureEnumerationNotStarted();
+            return Enumerate();
+        }
+
+        public IEnumerable<SqlRow> EnumerateOnce()
+        {
+            IEnumerable<SqlRow> Enumerate()
+            {
+                // FIXME [Oleksii] discuss synchronous approach
+                while (MoveNextAsync().GetAwaiter().GetResult())
```
alexb5dh, line 114: @zpqrtbnk 
We really need synchronous `Enumerate` to support LINQ without external libraries like [System.Linq.Async](https://github.com/dotnet/reactive#systemlinqasync--systeminteractiveasync--systeminteractive), so wanted to check if you're ok with this sync-over-async approach.

src/Hazelcast.Net/Sql/SqlResultType.cs
alexb5dh: Still need it to pass to Hazelcast servers as expected type parameter, but made it internal now.

src/Hazelcast.Net/Sql/ISqlService.cs
alexb5dh: Via cancelling passed `cancellationToken`. Here's an example where it's cancelled by timeout:
```csharp
using var cancellationSource = new CancellationTokenSource(TimeSpan.FromSeconds(1));
var updated = await client.Sql.ExecuteCommandAsync("UPDATE ...", cancellationSource.Token);
```

src/Hazelcast.Net/Sql/ISqlService.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// </para>
+        /// </param>
+        /// <param name="options">
+        /// <see cref="SqlStatementOptions"/> to use with the query.
+        /// <see cref="SqlStatementOptions.Default"/> is used if not specified.
+        /// </param>
+        /// <returns>
+        /// <see cref="ISqlQueryResult"/> representing set of rows returned from this query.
+        /// </returns>
+        Task<ISqlQueryResult> ExecuteQueryAsync(string sql, object[] parameters = null,
```
alexb5dh, line 40: @zpqrtbnk, need you thoughts on best signature for this method.

For now `ISqlQueryResult` is returned immediately and method can be made fully synchronous. On the other hand, I assume it's possible we will make initial result fetching asynchronous in later versions and upgrade will be simpler if we implement it as `async` initially. Also, adjacent `ExecuteCommandAsync` is completely asynchronous.

In case if we go with `async` way - should we add `CancellationToken` parameter?

src/Hazelcast.Net/Sql/ISqlService.cs
alexb5dh: Continued this discussion in a [new thread](https://github.com/hazelcast/hazelcast-csharp-client/pull/475/files#r675454901).

src/Hazelcast.Net/Sql/HLocalDate.cs
zpqrtbnk: I do believe that date/time types should be supported "by default" for everything, for instance also in maps, not purely in SQL, so I *think* it's reasonable?

zpqrtbnk: Nope, no other place that I am aware of.

src/Hazelcast.Net/Sql/SqlQueryResult.cs
zpqrtbnk: Well, that's a problem. `GetAwaiter().GetResult()` is bad. Is using System.Linq.Async bad?

src/Hazelcast.Net/Sql/ISqlService.cs
zpqrtbnk: Probably more future-proof to make it async yes. And I think we could accept a `CancellationToken` parameter yes.

src/Hazelcast.Net/Sql/SqlColumnType.cs
alexb5dh: Looks like this was fixed in one of the updates and it's serialized as int (4 bytes) everywhere.

src/Hazelcast.Net/Sql/HLocalDate.cs
alexb5dh: Missed that we already have built-in `DateTime` serialization via `DateSerializer`.

alexb5dh: Agreed with @mdumandag  on adding default serializers for them.

srknzl:
@alexb5dh In 5.0, they ditched get random connection to a data member approach. Now we should use https://github.com/hazelcast/hazelcast/blob/bd56b56517984be3925b894db78079646a2188ec/hazelcast/src/main/java/com/hazelcast/client/impl/connection/tcp/TcpClientConnectionManager.java#L840

srknzl:
See the new execute() method in java https://github.com/hazelcast/hazelcast/blob/bd56b56517984be3925b894db78079646a2188ec/hazelcast/src/main/java/com/hazelcast/sql/impl/client/SqlClientService.java#L70

src/Hazelcast.Net/Sql/SqlQueryResult.cs
alexb5dh: I don't see anything bad about using it, but anyone who would want to consume SQL result via LINQ will be forced to install separate package. This includes our tests (or in testing library we can just copy/implement needed extension methods).

alexb5dh:
> @alexb5dh In 5.0, they ditched get random connection to a data member approach. Now we should use https://github.com/hazelcast/hazelcast/blob/bd56b56517984be3925b894db78079646a2188ec/hazelcast/src/main/java/com/hazelcast/client/impl/connection/tcp/TcpClientConnectionManager.java#L840

@srknzl could you please clarify what do you mean by data member approach? Just selecting random/round-robin data member instead of any type of member? Or some more significant change?

srknzl:
I am talking about choosing a connection to send sql execute request. 

We were using `getRandomConnection(dataMember: boolean)` and load balancer methods `canGetNextDataMember` and `nextDataMember` were being used. Now in 5.0 they are deprecated and a new `getConnectionForSql` method is added and used. See my pr for nodejs https://github.com/hazelcast/hazelcast-nodejs-client/pull/988

src/Hazelcast.Net/Sql/ISqlQueryResult.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// <para>
+        /// Invoking this method again after enumeration has started, will throw <see cref="System.InvalidOperationException"/>.
+        /// </para>
+        /// <para>
+        /// Reusing obtained enumerable may lead to unpredictable results.
+        /// </para>
+        /// </summary>
+        /// <exception cref="ObjectDisposedException">SQL result was disposed.</exception>
+        /// <exception cref="InvalidOperationException">Enumeration has already started.</exception>
+        IEnumerable<SqlRow> EnumerateOnce();
```
zpqrtbnk, line 54: Do we need that sync method? Everything in our API is async, can't we simply have the async method?

src/Hazelcast.Net/Sql/ISqlQueryResult.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// <para>
+        /// Invoking this method again after enumeration has started, will throw <see cref="System.InvalidOperationException"/>.
+        /// </para>
+        /// <para>
+        /// Reusing obtained enumerable may lead to unpredictable results.
+        /// </para>
+        /// </summary>
+        /// <exception cref="ObjectDisposedException">SQL result was disposed.</exception>
+        /// <exception cref="InvalidOperationException">Enumeration has already started.</exception>
+        IAsyncEnumerable<SqlRow> EnumerateOnceAsync();
```
zpqrtbnk, line 39: Think we discussed this already so this is just for confirmation: we could make `ISqlQueryResult` an `IAsyncEnumerable<>` but the risk is that users enumerate more than once, whereas here we make it more explicit that you should enumerate only once, correct?

zpqrtbnk: In fact I am not _entirely_ convinced about that method, I find it nicer to do `await foreach (var row in result)` ... but I can understand the idea. Now one thing: we should support cancellations. That is: `await foreach (var row in result.EnumerateOnceAsync().WithCancellation(token)` - not sure this works at the moment?

src/Hazelcast.Net.Examples/Sql/SqlCancellationExample.cs
```diff
@@ -0,0 +1,108 @@
...
+            // To run this example you will need to enable Jet in the cluster configuration via '<jet enabled="true"></jet>'.
+            // Below is a simple example of such setup:
+            // "<?xml version=\"1.0\" encoding=\"UTF-8\"?>" +
+            // <hazelcast xmlns="http://www.hazelcast.com/schema/config
+            //   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance
+            //   xsi:schemaLocation="http://www.hazelcast.com/schema/config http://www.hazelcast.com/schema/config/hazelcast-config-5.0.xsd">
+            //   <jet enabled="true"></jet>" +
+            // </hazelcast>;
+
+            // cancel after timeout
```
zpqrtbnk, line 48: I _think_ this would be way cleaner as:
```
{
    var timeout = TimeSpan.FromSeconds(10);
    var cancellation = new CancellationTokenSource(timeout);
    await using var result = client.Sql.ExecuteQuery("SELECT * FROM TABLE(generate_stream(1))");

    await foreach (var row in result.EnumerateOnceAsync().WithCancellation(cancellation.Token))
    {
        logger.LogInformation("{Value}", row.GetColumn<long>(0));
    }
}
```
but I'm not sure that's supported at the moment?

src/Hazelcast.Net/HazelcastClient.cs
```diff
@@ -42,6 +44,7 @@ internal partial class HazelcastClient : IHazelcastClient
         private readonly DistributedObjectFactory _distributedOjects;
         private readonly NearCacheManager _nearCacheManager;
         private readonly MetricsPublisher _metricsPublisher;
+        private readonly SqlService _sqlService;
```
zpqrtbnk, line 47: don't need that field

src/Hazelcast.Net/Protocol/Codecs/SqlCloseCodec.cs
```diff
@@ -0,0 +1,116 @@
+// Copyright (c) 2008-2021, Hazelcast, Inc. All Rights Reserved.
```
zpqrtbnk, line 1: For this codec, and all other codecs and custom codecs: I assume you generated them via the protocol repository (eg the `hz generate-codecs` command)? So you probably made some changes to the protocol repo? Can you push a PR for that repo, and update the submodule accordingly?
(ping me for infos)

zpqrtbnk (commented):
Have commented.

src/Hazelcast.Net.Testing/Remote/Resources.resx
```diff
@@ -1,17 +1,17 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
alexb5dh: Which resource? I placed _jet-enabled.xml_ where _hazelcast.xml_ used by default is located. Should I also move _hazelcast.xml_?

src/Hazelcast.Net.Testing/CollectionExtensions.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+
+namespace Hazelcast.Testing
+{
+    public static class CollectionExtensions
+    {
+        public static T GetOrDefault<T>(this IList<T> source, int index) => (source.Count > index || index < 0)
+            ? source[index]
```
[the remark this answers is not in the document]
alexb5dh: Will fix and rename.

So in case of negative index you want it to throw regular exception?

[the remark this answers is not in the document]
alexb5dh: Actually, I'm removing this as not needed anymore.

src/Hazelcast.Net/Sql/ISqlQueryResult.cs
alexb5dh: We already have this discussion ongoing, got hidden for some reason:
https://github.com/hazelcast/hazelcast-csharp-client/pull/475#discussion_r675435531

src/Hazelcast.Net/Sql/ISqlQueryResult.cs
alexb5dh: > we could make ISqlQueryResult an IAsyncEnumerable<> but the risk is that users enumerate more than once, whereas here we make it more explicit that you should enumerate only once, correct?

Yes

alexb5dh: Will check for cancellation support

src/Hazelcast.Net/Protocol/Codecs/SqlCloseCodec.cs
alexb5dh: Already have, forgot to mention: https://github.com/hazelcast/hazelcast-client-protocol/pull/382

src/Hazelcast.Net/HazelcastClient.cs
alexb5dh: Removed, thanks

src/Hazelcast.Net.Examples/Sql/SqlCancellationExample.cs
zpqrtbnk: Even better, this would make me happy
```
{
    var timeout = TimeSpan.FromSeconds(10);
    var cancellation = new CancellationTokenSource(timeout);
    await using var result = client.Sql.ExecuteQuery("SELECT * FROM TABLE(generate_stream(1))");

    await foreach (var row in result.WithCancellation(cancellation.Token))
    {
        logger.LogInformation("{Value}", row.GetColumn<long>(0));
    }
}
```

src/Hazelcast.Net.Testing/Remote/Resources.resx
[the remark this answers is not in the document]
zpqrtbnk: The `hazelcast.xml` in Testing is the default one, for when nothing else is specified, but all the specific one lives in the Test project, in the Resources/Cluster directory. The idea was that the Testing project was really just the testing "plumbing" and utilities, but the actual things and data and stuff we use for tests are in the Test project. So I'd leave `hazelcast.xml` where it is, but move `jet-enabled.xml` to the Test project along with the `kerberos.xml`, `ssl.xml` etc that you will find there.

src/Hazelcast.Net/Sql/ISqlQueryResult.cs
zpqrtbnk: I am having a hard time with these discussions getting hidden :( OK, 
* I don't want to keep the `GetAwaiter().GetResult()` thing in our code, that's an anti-pattern that can, and will, cause issues
* I would *love* to entirely remove everything about sync-enumerating
Let's slack-discuss it

src/Hazelcast.Net.Testing/Remote/Resources.resx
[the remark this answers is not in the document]
alexb5dh: Done

src/Hazelcast.Net/Protocol/Codecs/SqlCloseCodec.cs
alexb5dh: Merged and updated submodule ref

src/Hazelcast.Net/Sql/SqlQueryResult.cs
alexb5dh: Decided to remove and add example of LINQ over `IAsyncEnumerable` using external package.

src/Hazelcast.Net/Sql/ISqlQueryResult.cs
alexb5dh: Added support for `WithCancellation`. Token is forwarded successfully, but there are few important notes:

- To properly support it, `SqlQueryResult` execution needs to be delayed. Before first page query started when it was constructed, now it starts with first call of `MoveNextAsync`, so that enumeration cancellation token can be used.

- It seems that server invocation itself can't be cancelled with token. So query may be cancelled in case of invocation retry or at the start/end of `MoveNextAsync`, but if page fetching itself takes too long, cancellation token won't affect it. Reason is `SendAsync` doesn't use token provided, see _MemberConnection.cs:374_. Not sure why it is implemented that way, probably worth a discussion.

src/Hazelcast.Net.Examples/Sql/SqlCancellationExample.cs
alexb5dh: Will update examples and documentation once we finalize result interface.

src/Hazelcast.Net/Sql/SqlResult.cs
alexb5dh: `EnumerateOnce` is removed in favor of `IAsyncEnumerable` and System.Linq.Async package.

src/Hazelcast.Net/Sql/ISqlQueryResult.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// Creates a one-off <see cref="IAsyncEnumerable{T}"/> around <see cref="SqlRow"/>s in row set.
+        /// </para>
+        /// <para>
+        /// Reusing obtained enumerable or invoking this method again
+        /// will continue previous enumeration instead of starting a new one.
+        /// </para>
+        /// </summary>
+        /// <param name="cancellationToken">Optional <see cref="CancellationToken"/> to cancel enumeration in progress.</param>
+        /// <exception cref="ObjectDisposedException">SQL result was disposed.</exception>
+        IAsyncEnumerable<SqlRow> EnumerateOnceAsync(CancellationToken cancellationToken = default);
```
zpqrtbnk, line 38: We don't want the `CancellationToken` here. The .NET way with `IAsyncEnumerable<>` is to do
```
await foreach (var row in result.EnumerateOnceAsync().WithCancellation(cancellationToken)) {...}
```
And the `cancellationToken` is passed to the enumerator.

src/Hazelcast.Net/Sql/ISqlQueryResult.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Collections.Generic;
+using System.Threading;
+
+namespace Hazelcast.Sql
+{
+    /// <summary>
+    /// Represents SQL query (SELECT ...).
+    /// Provides methods to enumerate queried rows or cancel/dispose the query.
+    /// </summary>
+    public interface ISqlQueryResult: IAsyncEnumerator<SqlRow>, IAsyncDisposable
```
zpqrtbnk, line 25: Don't think we want to publicly expose that `ISqlQueryResult` implements `IAsyncEnumerator<>`?

src/Hazelcast.Net.Examples/Sql/SqlCancellationExample.cs
zpqrtbnk: So I still think that this example is too complicated and the syntax above should work.

zpqrtbnk (commented):
Have commented - I still have issues with cancellations

src/Hazelcast.Net/Sql/ISqlQueryResult.cs
alexb5dh: Removed

src/Hazelcast.Net.Examples/Sql/SqlCancellationExample.cs
alexb5dh: Updated cancellation example

src/Hazelcast.Net/Sql/HazelcastSqlException.cs
```diff
@@ -0,0 +1,38 @@
...
+// limitations under the License.
+
+using System;
+using Hazelcast.Exceptions;
+
+namespace Hazelcast.Sql
+{
+    public class HazelcastSqlException : HazelcastException
+    {
+        public Guid ClientId { get; }
```
alexb5dh, line 22: I think we name it's named `ClientId` in other clients, but `MemberId` in C#? @zpqrtbnk, should I use `MemberId` everywhere?

src/Hazelcast.Net/Sql/SqlQueryResult.cs
alexb5dh: Done

PR #476, opened 2021-07-05
Add Kerberos example [API-278]

zpqrtbnk:
Add a minimal Kerberos example to our examples project.

### DIFF
--- a/src/Hazelcast.Net.Examples/Client/KerberosExample.cs
+++ b/src/Hazelcast.Net.Examples/Client/KerberosExample.cs
@@ -1,0 +1,102 @@
+﻿// Copyright (c) 2008-2021, Hazelcast, Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+// http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


alexb5dh:
Is there a simple way to setup Kerberos for testing without paying for some license, renting Azure machines, etc.?

src/Hazelcast.Net.Examples/Client/KerberosExample.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            public Worker(IHazelcastClient client, ILogger<Worker> logger)
+            {
+                _client = client;
+                _logger = logger;
+            }
+
+            public async Task RunAsync()
+            {
+                _logger.LogDebug("debug");
```
alexb5dh, line 85: Why 3 levels of logging?

zpqrtbnk: Inherited from another example, just so you can verify what level is currently enabled. Do you think it might confuse people?

zpqrtbnk:
Re. testing Kerberos: to make it simple, it's a pain. No so much because you need to pay, but you need at least a Active Directory Domain Controller that you can have access to and configure the Kerberos service. I have tested locally with a Hyper-V VM here running Windows Server 2019 as an AD DC and it takes time to configure. All I can say is that it works.

Ideally there would be a simple way to test but I cannot thing of any now.

alexb5dh (approved):

PR #481, opened 2021-07-14
[API-642, API-643] Distributed Query & Computing documentation

alexb5dh:
Checked and updated code from v3 to current implementation, simplified some text, fixed markup & typos, etc.


zpqrtbnk (approved):
awesome job, cannot find anything to change!

PR #483, opened 2021-07-16
External Smart-Client Discovery [API-561]

zpqrtbnk:
This PR implements external Smart-Client discovery, for Kubernetes support, so that the Smart-Client mode described in [this guide](https://guides.hazelcast.org/kubernetes-external-client/) is supported. 

In addition, the PR implements a few changes to options, how they are built and how they are exposed by the client, to make the general experience smoother.

## Note
This PR cannot fully work when the cluster falls down to 1 member, due to an issue in the Java Kubernetes code (see [this PR](https://github.com/hazelcast/hazelcast/pull/19168)) - which will be fixed in v5. So if you want to test by following the guide, just make sure not to bring the cluster down to 1 only member. Or, run the latest v5 snapshot, as the PR has now been merged.

## Review

The PR essentially includes a new `ConnectAddressResolver` class whose role is to detect/determine whether members should be reached via their internal or public address, along with a new `Networking.UsePublicAddress` option which can be used to bypass this detection and force usage of either public or private addresses.

The key changes happen in `Hazelcast.Clustering` changing how the client handles connections. Also, `Hazelcast.Models.MemberInfo` changes to handle the fact that one member can have a public and a private address. Then, the PR also refactors various aspects of configuration, mostly to make things more consistent.

Review should consist in code review and, ideally, testing that the Smart-Client mode described in the guide works. However, that requires a working Kubernetes environment, which can be a relatively painful thing to setup.


zpqrtbnk:
Moving the discussion over to [this PR](https://github.com/hazelcast/hazelcast/pull/19111) on the main Hazelcast repo.

zpqrtbnk:
That was a bad idea, actually. Closed the wrong PR.

src/Hazelcast.Net/Core/EnumerableExtensions.cs
```diff
@@ -32,6 +32,20 @@ internal static class EnumerableExtensions
         public static IEnumerable<T> Shuffle<T>(this IEnumerable<T> source)
             => source.OrderBy(x => RandomProvider.Next());
 
+        /// <summary>
+        /// Shuffles an enumerable.
```
alexb5dh, line 36: Source is not just enumerable, is enumerable mention intended?

src/Hazelcast.Net/Networking/ConnectAddressResolver.cs
```diff
@@ -0,0 +1,186 @@
...
+        /// <returns><c>true</c> if public addresses must be used; otherwise <c>false</c>, indicating that internal addresses can be used.</returns>
+        public async Task<bool> DetermineUsePublicAddresses(IReadOnlyCollection<MemberInfo> members)
+        {
+            // if the user has specified its intention, respect it, otherwise try to decide
+            // automatically whether to use private or public addresses by trying to reach
+            // a few members
+
+            if (_options.UsePublicAddresses.HasValue)
+            {
+                var usePublicAddresses = _options.UsePublicAddresses.Value;
```
alexb5dh, line 44: Can be simplified a bit, but it's a matter of preference:
```suggestion
            if (_options.UsePublicAddresses is {} usePublicAddresses)
            {
```

src/Hazelcast.Net/Networking/NetworkAddress.cs
```diff
@@ -302,6 +305,32 @@ internal static bool TryParse(string s, out NetworkAddress address, int defaultP
         /// <returns>A new instance of the network address with the same host but a different port.</returns>
         internal NetworkAddress WithPort(int port) => new NetworkAddress(Host, port);
 
+        /// <summary>
+        /// Determines whether this address is reachable, by trying to connect to it.
+        /// </summary>
+        /// <param name="timeout">A timeout.</param>
+        /// <returns><c>true</c> if the connection was successful; otherwise false.</returns>
+        /// <remarks>Use a timeout value of -1ms for infinite.</remarks>
```
alexb5dh, line 313: .NET has a constant you can reference in such cases: `System.Threading.Timeout.InfiniteTimeSpan`

alexb5dh (approved):

alexb5dh (approved):

src/Hazelcast.Net/Networking/NetworkAddress.cs
zpqrtbnk: yes but I prefer to be explicit here

src/Hazelcast.Net/Networking/ConnectAddressResolver.cs
zpqrtbnk: nice, going to use it

src/Hazelcast.Net/Core/EnumerableExtensions.cs
zpqrtbnk: bad copy/paste, fixing

PR #492, opened 2021-08-16
[API-274] SQL types

alexb5dh:


zpqrtbnk:
Only 1 comment / question really: are the new types pure-SQL types, or could they be useful for other applications, and then, should they be in a `Hazelcast.Sql` namespace or in a less SQL namespace or even in the root namespace? What I mean is... SQL is a query language, these types could be used without using SQL, couln't they?

mdumandag:
@zpqrtbnk In the Java client, these types (date types + BigDecimal) are used in

- SQL as return types/parameters of queries
- Portable as return types/parameters of readXXX/writeXXX methods
- new Compact Serialization as return types/parameters of readXXX/writeXXX methods

Also, they have default serializers registered for them (in 5.0), so they can also be used in any other data structure

zpqrtbnk:
Suggesting we move them to `Hazelcast.Models` which contains various "models" such as the `MemberInfo` class.
WDYT?

alexb5dh:
@zpqrtbnk @mdumandag 

Moved new types from `Hazelcast.Sql` to `Hazelcast.Models`. 
Corresponding tests file moved and renamed as: `/Sql/SqlDataTypesTests.cs` -> `Models/HazelcastDataTypesTests.cs`

zpqrtbnk (approved):
Looks good with the changes. Approved.

PR #493, opened 2021-08-19
Hz script update

alexb5dh:
- Switch to hazelcast-all package
- Update RC version to 0.8 (optional, can be reverted)


zpqrtbnk (approved):
Looks good to me, makes sense.

PR #494, opened 2021-08-19
Get rid of FIXMEs in code and all warnings

zpqrtbnk:
Just a small PR to get rid of all FIXMEs in the code and of all other warnings.

### DIFF
--- a/src/Hazelcast.Net/Networking/SocketConnectionBase.cs
+++ b/src/Hazelcast.Net/Networking/SocketConnectionBase.cs
@@ -231,7 +231,6 @@ namespace Hazelcast.Networking
             _streamReadCancellationTokenSource.Cancel();
 
             // ensure everything is down by awaiting the other task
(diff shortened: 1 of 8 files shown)


src/Hazelcast.Net/Networking/SocketConnectionBase.cs
```diff
@@ -231,7 +231,6 @@ private async ValueTask ShutdownInternal(Task task)
             _streamReadCancellationTokenSource.Cancel();
 
             // ensure everything is down by awaiting the other task
-            // FIXME why could this be null ?!
```
alexb5dh, line 234: So have you found an answer? :)

zpqrtbnk: No :) I don't think it can ben null & I don't remember why this warning is here, so I remove it.

alexb5dh (approved):

PR #497, opened 2021-08-25
Bugfix networking code

zpqrtbnk:
Depends on #512

Kubernetes tests reveal that in some occasion our networking code does not like connections to member going up and down in unpredictable ways - this PR fixes the problem.


src/Hazelcast.Net/Clustering/ClusterConnections.cs
```diff
@@ -120,10 +127,22 @@ await foreach(var (member, token) in _clusterMembers.MembersToConnect.WithCancel
                 else
                 {
                     var details = wasCanceled ? "canceled" : "failed";
```
alisengul53, line 129: Is it still needed to set `details` as canceled or failed here?

src/Hazelcast.Net.Tests/Clustering/ConnectMembersTests.cs
```diff
@@ -68,53 +70,66 @@ await foreach (var (member, token) in memberConnectionQueue.WithCancellation(can
...
+
+            await mutex.WaitAsync(); // blocks the connections
 
             queue.Add(MemberInfo(NetworkAddress.Parse("127.0.0.1:4")));
             await Task.Delay(500);
 
-            queue.Suspend();
-            mutex.Release(); // resume => should cancel current connect
+            var task = queue.SuspendAsync();
+            Assert.That(task.IsCompleted, Is.False);
```
alisengul53, line 100: Is there any specific reason to use always `Assert.That`
What about `Assert.IsTrue(task.IsCompleated)` for true false checks

src/Hazelcast.Net/Clustering/ClusterMembers.cs
```diff
@@ -167,18 +176,25 @@ public void AddConnection(MemberConnection connection, bool isNewCluster)
...
                     {
-                        _logger.LogDebug($"Added connection {connection.Id.ToShortString()} to member {connection.MemberId.ToShortString()}.");
+                        if (_logger.IsEnabled(LogLevel.Debug))
+                        {
+                            var msg = $"Added connection {connection.Id.ToShortString()} to member {connection.MemberId.ToShortString()} at {connection.Address}";
+                            msg += existingConnection == null
+                                ? "."
+                                : $", replacing connection {existingConnection.Id.ToShortString()} at {existingConnection.Address}.";
+                            _logger.LogDebug(msg);
+                        }
```
alisengul53, line 192: Should we log something here when the LogLevel is not debug?
If it is unneccessary, what about the using `else if` directly

src/Hazelcast.Net/Clustering/ClusterConnections.cs
zpqrtbnk: well, yes, I think so? `exception` could be `null`?

alisengul53: I mean in every block `details` var is set to something else, are we using the values `canceled` or `failed` after setting it?

alisengul53 (approved):
Looks good to me generally!

src/Hazelcast.Net/Clustering/ClusterConnections.cs
zpqrtbnk: Yes but if `exception` is null? None of the blocks run, or am I missing something?

alisengul53: Ah sorry my bad, i missed the `details` at the end of the line

PR #498, opened 2021-08-28
[API-545] Flake ID Generator

alexb5dh:
[Protocol PR](https://github.com/hazelcast/hazelcast-client-protocol/pull/390)


src/Hazelcast.Net/HazelcastClient.Objects.cs
```diff
@@ -183,6 +184,24 @@ Task<IHRingBuffer<T>> GetRingBufferAsync<T>(string name)
...
+        }
+
+        /// <inheritdoc />
+        public
+#if !HZ_OPTIMIZE_ASYNC
+        async
+#endif
+        Task<IFlakeIdGenerator> GetFlakeIdGeneratorAsync(string name)
+        {
+            var task = _distributedOjects.GetOrCreateAsync<IFlakeIdGenerator, FlakeIdGenerator>(ServiceNames.Topic, name, true,
```
zpqrtbnk, line 201: Should not be `ServiceNames.Topic` here I assume?

src/Hazelcast.Net/HazelcastOptions.cs
```diff
@@ -66,6 +67,10 @@ private HazelcastOptions(HazelcastOptions other)
             NearCaches = other.NearCaches.Select(kvp
                 => new KeyValuePair<string, NearCacheOptions>(kvp.Key, kvp.Value.Clone()))
                 .ToDictionary();
+
+            FlakeIdGenerators = other.FlakeIdGenerators.Select(kvp
```
zpqrtbnk, line 71: `other.FlakeIdGenerator.ToDictionary(x => x, x => x.Clone())` would work too?

src/Hazelcast.Net/HazelcastClient.Objects.cs
```diff
@@ -183,6 +184,24 @@ Task<IHRingBuffer<T>> GetRingBufferAsync<T>(string name)
...
+                (n, factory, cluster, serializationService, loggerFactory)
+                    => new FlakeIdGenerator(n, factory, cluster, serializationService, loggerFactory, _options));
```
zpqrtbnk, line 203: See my note on `FlakeIdGenerator` constructor

src/Hazelcast.Net/FlakeId/FlakeIdGenerator.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// </summary>
+        /// <param name="name">The unique name of the object.</param>
+        /// <param name="factory">The factory that owns this object.</param>
+        /// <param name="cluster">A cluster.</param>
+        /// <param name="serializationService">A serialization service.</param>
+        /// <param name="loggerFactory">A logger factory.</param>
+        /// <param name="options">A client options.</param>
+        public FlakeIdGenerator(string name, DistributedObjectFactory factory, Cluster cluster,
+            SerializationService serializationService, ILoggerFactory loggerFactory,
+            HazelcastOptions options
```
zpqrtbnk, line 42: To better isolate things can we avoid passing a full `HazelcastOptions` object here? We can pass the dictionary of options, or the right flake id option directly, but better not create a dependency on the whole thing.

src/Hazelcast.Net.Testing/EnumerableExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Hazelcast.Testing
+{
+    public static class EnumerableExtensions
+    {
+        public static IEnumerable<IList<T>> Batch<T>(this IEnumerable<T> source, int batchSize)
```
zpqrtbnk, line 23: Can we have this return `IEnumerable<IEnumerable<T>>` and avoid allocating lists?

src/Hazelcast.Net/FlakeId/AutoBatcher.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        private Lazy<Task<Batch>> _nextBatchLazyTask;
+
+        public AutoBatcher(Func<Task<Batch>> supplier)
+        {
+            _supplier = supplier;
+            _nextBatchLazyTask = NewBatchLazyTask();
+        }
+
+        public async Task<long> GetNextIdAsync(CancellationToken cancellationToken = default)
```
zpqrtbnk, line 34: I am wondering whether we could make this return `ValueTask<long>` and non-async for perfs purposes. Pseudo-code:
```
public ValueTask<long> GetNextIdAsync(...)
{
  if (we have a batch and a value) return new ValueTask(value); // this is fast
  return GetNextIdAsyncSlow(...);
}
private async ValueTask GetNextIdAsyncSlow(...)
{
  // go fetch a new batch etc
}
```

src/Hazelcast.Net/FlakeId/Batch.cs
```diff
@@ -0,0 +1,58 @@
...
+using System;
+using System.Threading;
+
+namespace Hazelcast.FlakeId
+{
+    internal class Batch
+    {
+        private readonly long _increment;
+        private readonly long _max;
+        private readonly DateTime? _invalidSince;
```
zpqrtbnk, line 24: Can be a non-`null` `DateTime` and use `DateTime.MaxValue` when the `validityPeriod` is infinite?

zpqrtbnk (commented):
See my comments

src/Hazelcast.Net/HazelcastOptions.cs
alexb5dh: Just copied implementation above, should I also update it for `NearCaches`?

src/Hazelcast.Net.Testing/EnumerableExtensions.cs
alexb5dh: In such case `List`s will be allocated implicitly, there's no way to split `IEnumerable` into independent sub`IEnumerable`s.
I don't optimization matters that much here, as this is for tests only.

src/Hazelcast.Net/FlakeId/Batch.cs
alexb5dh: `DateTime.MaxValue` is not "really" infinite. Any reason to avoid `null`?

src/Hazelcast.Net/FlakeId/AutoBatcher.cs
alexb5dh: I think you preferred `Task` for public API?

src/Hazelcast.Net/HazelcastOptions.cs
zpqrtbnk: sure

src/Hazelcast.Net.Testing/EnumerableExtensions.cs
zpqrtbnk: ah yes it's in the `Testing` project not the main project, my mistake.
and... doh - an `IEnumerable<IEnumerable<T>>` without lists requires that each inner enumerable be enumerated in order... ok let's keep it like this.

src/Hazelcast.Net/FlakeId/Batch.cs
zpqrtbnk: code simplicity but I'm ok if we keep it like it is

src/Hazelcast.Net/FlakeId/AutoBatcher.cs
zpqrtbnk: When we are 100% certain that the operation will be async, yes a `Task` is preferred, but here we have to opportunity to optimize by avoiding the `async` state machine entirely. So I think a `ValueTask` is justified here.

src/Hazelcast.Net.Tests/FlakeId/BatchTests.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Hazelcast.DistributedObjects.Impl;
+using Hazelcast.Testing;
+using NUnit.Framework;
+
+namespace Hazelcast.Tests.FlakeId
+{
+    using BatchTests_;
```
alexb5dh, line 26: Is this better than having a separate file?

src/Hazelcast.Net.Tests/Sandbox/SupportTests.cs.not
```diff
@@ -0,0 +1,143 @@
+﻿// Copyright (c) 2008-2021, Hazelcast, Inc. All Rights Reserved.
```
alexb5dh, line 1: What's this file about?

src/Hazelcast.Net/DistributedObjects/FlakeIdGeneratorOptions.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <summary>
+        /// Gets the number of identifiers that are pre-fetched when a new id is requested from the cluster.
+        /// </summary>
+        /// <remarks>
+        /// <para>Allowed values are between <c>1</c> and <c>100000</c> inclusive, default value is <c>100</c>.</para>
+        /// </remarks>
+        public int PrefetchCount { get; set; } = 100;
+
+        /// <summary>
+        /// Gets the
```
alexb5dh, line 50: Added by mistake?

src/Hazelcast.Net/DistributedObjects/FlakeIdGeneratorOptions.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// Default value is <c>10 minutes</c>.
+        /// </para>
+        /// <para>
+        /// The IDs contain timestamp component, which ensures rough global ordering of IDs.
+        /// If an ID is assigned to an object that was created much later, it will be much out of order.
+        /// If you don't care about ordering, set this value to <see cref="Timeout.InfiniteTimeSpan"/> for unlimited ID validity.
+        /// </para>
+        /// </summary>
+        /// <remarks>
+        /// <para>FIXME</para>
```
alexb5dh, line 63: What to fix here?

src/Hazelcast.Net/DistributedObjects/Impl/AutoBatcher.cs
```diff
@@ -0,0 +1,46 @@
...
+        private readonly ClusterMessaging _messaging;
+
+        public AutoBatcher(string name, FlakeIdGeneratorOptions options, ClusterMessaging messaging)
+        {
+            _name = name;
+            _options = options;
+            _messaging = messaging;
+        }
+
+        protected override async Task<Batch> FetchBatch()
```
alexb5dh, line 35: Just a note - initial version took batch provider method externally and was only responsible for concurrently iterating batch and fetching next one as needed. It somehow separated it's code from flake id and simplified testing approach.

alexb5dh:
Just a note - initial  `AutoBatcher` version took batch provider method externally and was only responsible for concurrently iterating batch and fetching next one as needed. It somehow separated it's code from flake id and simplified testing approach.

src/Hazelcast.Net.Tests/Sandbox/SupportTests.cs.not
zpqrtbnk: Should not be part of the PR, can be removed.

src/Hazelcast.Net.Tests/FlakeId/BatchTests.cs
zpqrtbnk: Ah, I am not too sure. Ideally I'd want a test to "pollute" the namespace as little as possible, so my first thought was, if a method or a class is used by one test only, let's keep it private to the test. But then, when it's an extension method, it cannot be nested inside the test class itself, so how do we achieve this? I have the same situation in other tests, where clearly the extension method only makes sense for one test, but it's convenient. This namespace hack is useful. But it's... not pretty. Maybe you can come up with a better idea?

We can:
* Move everything back to their own file.
* Use a better namespace name such as `BatchTestsDependencies` or something, and use the same pattern everywhere.
*  ?

I am not too sure and welcome suggestions / discussions.

src/Hazelcast.Net/DistributedObjects/FlakeIdGeneratorOptions.cs
zpqrtbnk: Silly me. Wanted to cleanup the summary to use the more-or-less standard "Gets the..." pattern and did not fully do it, hence the FIXME comment at the end.

src/Hazelcast.Net/DistributedObjects/FlakeIdGeneratorOptions.cs
zpqrtbnk: See question above.

zpqrtbnk:
Re. `AutoBatcher` with batch provider method: I liked the approach of separating concerns. Yet that could introduce an extra async state machine that had a (very little, sure) impact on perfs. So I tried to sort-of keep the approach via the batcher base class, which allows for testing most of the functionalities independently from FlakeId. WDYT?

src/Hazelcast.Net/DistributedObjects/FlakeIdGeneratorOptions.cs
```diff
@@ -0,0 +1,72 @@
...
+using System;
+using System.Threading;
+
+namespace Hazelcast.DistributedObjects
+{
+    /// <summary>
+    /// Represents options for <see cref="IFlakeIdGenerator"/>.
+    /// </summary>
+    public class FlakeIdGeneratorOptions
+    {
```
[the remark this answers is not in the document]
zpqrtbnk: Was not too sure about why we needed to expose it publicly + none of the other options do it.

src/Hazelcast.Net.Tests/FlakeId/BatchTests.cs
alexb5dh: No ideas at the moment, maybe will figure something out later

src/Hazelcast.Net/DistributedObjects/FlakeIdGeneratorOptions.cs
alexb5dh: Adjuster summary.

zpqrtbnk:
Superseded by #517

PR #500, opened 2021-08-31
Bugfix Predicates.In

### LINKED ISSUE #496: Predicates.In not working for IEnumerable primary types.
Assume:
`
var array = new List<int>{ 1, 2, 3 };
var query = Predicates.In("fieldName", array);
`
This will not work, neither casting it to an array with `.ToArray<T>(IEnumerable<T> list)`

I get some kind of an exception that my array or object does not implement java.lang.Comparable.

Was able to work around is as follows
`
var list = new List<int>{ 1, 2, 3 };
var array = list.Cast<object>().ToArray();
var query = Predicates.In("fieldName", array);
`

I am writing my maps with key type 'string' and value type 'HazelcastJsonValue'.

zpqrtbnk:
This fixes #496 .

### DIFF
--- a/src/Hazelcast.Net.Tests/Remote/ClientMapTest.cs
+++ b/src/Hazelcast.Net.Tests/Remote/ClientMapTest.cs
@@ -246,7 +246,7 @@ namespace Hazelcast.Tests.Remote
         }
 
         [Test]
-        public async Task TestEntrySetPredicate()
+        public async Task EntrySetWithSqlPredicate()
         {
             var dictionary = await Client.GetMapAsync<string, string>(CreateUniqueName());
             await using var _ = DestroyAndDispose(dictionary);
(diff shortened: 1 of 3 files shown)


src/Hazelcast.Net.Tests/Remote/ClientMapTest.cs
```diff
@@ -262,6 +262,41 @@ public async Task TestEntrySetPredicate()
...
 
+        [Test]
+        public async Task EntrySetWithInPredicate()
+        {
+            var map = await Client.GetMapAsync<string, HazelcastJsonValue>(CreateUniqueName());
+            await using var _ = DestroyAndDispose(map);
+
+            for (var i = 1; i < 30; i++)
+            {
+                await map.SetAsync("key-" + i, new HazelcastJsonValue($"{{ \"key\": \"key-{i}\", \"i-value\": {i}, \"s-value\": \"{i}\" }}"));
```
alexb5dh, line 273: Why not `SetAllAsync`?

src/Hazelcast.Net.Tests/Remote/ClientMapTest.cs
```diff
@@ -262,6 +262,41 @@ public async Task TestEntrySetPredicate()
...
+
+            result = await map.GetEntriesAsync(Predicates.In("x-value", "1", "2", "3"));
+            Assert.That(result.Count, Is.EqualTo(0));
+
+            // it must work also if we pass only 1 argument that is an IEnumerable<T>
+            // this used to fail (see issue #496) before we added the proper overload to Predicates.In
+            var values = new List<int> { 1, 2, 3 };
+            var query = Predicates.In("i-value", values);
+            result = await map.GetEntriesAsync(query);
+            Assert.That(result.Count, Is.EqualTo(3));
```
alexb5dh, line 293: Directly matching with expected collection would be more reliable than just comparing number of elements. And we can directly replicate expected `Predicates.In` behavior with a bit of LINQ. What do you think?

zpqrtbnk: updating

src/Hazelcast.Net/Query/Predicates.cs
```diff
@@ -127,6 +129,15 @@ public static IPredicate Between(string name, object lowerBound, object upperBou
         public static IPredicate In(string name, params object[] values)
             => new InPredicate(name, values);
 
+        /// <summary>
+        /// Succeeds if the target value is equal to one of the specified values.
+        /// </summary>
+        /// <param name="name">The name of the target.</param>
+        /// <param name="values">The values.</param>
+        /// <returns>A new predicate.</returns>
+        public static IPredicate In<T>(string name, IEnumerable<T> values)
```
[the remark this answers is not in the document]
zpqrtbnk: Have added a TODO - we cannot do it now in any simple way as deep down under, the serialization service ultimately treats everything as objects, and this is the first thing we need to fix.

alexb5dh (approved):

alexb5dh:
New test is failing, otherwise all good.

Issue #501, opened 2021-09-01
HMap EntryRemoved events do not include the OldValue


osmedd:
When subscribing to EntryRemoved events for an HMap (for example as in the MapEventsExample), I do not see the OldValue being populated in the event args.  This does work for EntryUpdated events and did work in the 3.x .NET client.

I am using the 4.1.0 client and tested against the 4.2.1 server as well as a 5.0 snapshot server.

Help?

zpqrtbnk:
Not 100% sure yet but probably reproduced. Stay tuned.

zpqrtbnk:
Confirmed, going to push a PR.

PR #502, opened 2021-09-01
Fix missing old value in map events

zpqrtbnk:
Fix issue #501

### DIFF
--- a/src/Hazelcast.Net.Tests/Support/Issue501.cs
+++ b/src/Hazelcast.Net.Tests/Support/Issue501.cs
@@ -1,0 +1,129 @@
+﻿// Copyright (c) 2008-2021, Hazelcast, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Hazelcast.Core;
+using Hazelcast.DistributedObjects.Impl;
+using Hazelcast.Testing;
+using Hazelcast.Testing.Logging;
+using Microsoft.Extensions.Logging;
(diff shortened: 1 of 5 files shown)


src/Hazelcast.Net.Tests/Support/Issue501.cs
```diff
@@ -0,0 +1,129 @@
...
+using Hazelcast.Core;
+using Hazelcast.DistributedObjects.Impl;
+using Hazelcast.Testing;
+using Hazelcast.Testing.Logging;
+using Microsoft.Extensions.Logging;
+using NUnit.Framework;
+
+namespace Hazelcast.Tests.Support
+{
+    public class Issue501 : SingleMemberRemoteTestBase
```
alexb5dh, line 27: Maybe we should add a link to GitHub issues here as a comment?

src/Hazelcast.Net.Tests/Support/Issue501.cs
```diff
@@ -0,0 +1,129 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Hazelcast.Core;
+using Hazelcast.DistributedObjects.Impl;
+using Hazelcast.Testing;
+using Hazelcast.Testing.Logging;
+using Microsoft.Extensions.Logging;
+using NUnit.Framework;
+
+namespace Hazelcast.Tests.Support
```
alexb5dh, line 25: What tests should be in Support folder?
For example in [this PR](https://github.com/hazelcast/hazelcast-csharp-client/pull/500/files#diff-03e6972c42687cdc082f791ce2f7a5542a79087d748555c3beca15dfd482c7b4R265) fixing #496, test is added directly to entity tests class. Shouldn't this one also go to `MapTests` or `ClientMapTests`?

src/Hazelcast.Net.Tests/Support/Issue501.cs
zpqrtbnk: added

src/Hazelcast.Net.Tests/Support/Issue501.cs
zpqrtbnk: It goes like this: when we investigate an issue we start creating a test in `Support/Issue123` but that test might be too big, because we don't know yet what's going on, and maybe we should even exclude it from normal testing, so it remains in `Support`, and another smaller, dedicated test should go in the normal tests. Which means... I need to update our normal tests. Working on it now.

zpqrtbnk: Pushed an update, tell me what you think.

alexb5dh (approved):

PR #504, opened 2021-09-01
Provide state (connected or not) of members at client level

### LINKED ISSUE #453: IHazelcastClient.Members shows all members of the cluster
There is no way to find currenct connections of a client to members. IHazelcastClient.Members property shows all members not only connected ones for client. It would be better to show connected ones with this property.
There could be another property like IHazelcastClient.AllMembers and it can return all members of the cluster

zpqrtbnk:
Fixes #453 by exposing members (`MemberInfo`) along with the connection state of the member, at client level.


src/Hazelcast.Net/HazelcastClient.cs
```diff
@@ -135,14 +136,17 @@ private void WireComponents()
         public HazelcastOptions Options => _optionsClone ??= _options.Clone();
         
         /// <inheritdoc />
-        public IReadOnlyCollection<MemberInfo> Members => Cluster.Members.GetMembers().ToList();
+        public IReadOnlyCollection<MemberInfoState> Members => Cluster.Members.GetMembersAndState().ToList();
```
alexb5dh, line 139: Can we leave old property for compatibility and add a new one, like `MembersState`?

alexb5dh (approved):

PR #507, opened 2021-09-05
Alternative flakeId implementation

zpqrtbnk:
Alternative flakeId implementation


zpqrtbnk:
I got a little carried on - let's use the original PR.

PR #509, opened 2021-09-15
general code cleanup, dependencies updates

zpqrtbnk:


src/Hazelcast.Net/Sql/HazelcastSqlException.cs
```diff
@@ -13,26 +13,117 @@
...
+        // the 2nd ctor below is used when the response to a request contains an error,
+        // in which case the error is deserialized by the codec and it contains the
+        // originating member id, ie the id of the member on which the error occurred
+        //
+        // so ClientId can be two totally different things and that is wrong. besides,
+        // these client/member identifiers could be used outside of SQL
+        // - would it make sense for HazelcastException to always include client id?
+        // - would it make sense for some exceptions to always include member id?
+        //   that would be for RemoteException, really, which is the other remote
+        //   error that the server can raise
```
mdumandag, line 81: I totally agree with problems around the ambiguity of this. In other clients, we use originatingMemberUuid as the name, but as you write it could either be a client or member UUID. Maybe a generic name like `origin` would be better.

src/Hazelcast.Net/Sql/HazelcastSqlException.cs
```diff
@@ -13,26 +13,117 @@
...
+using System.Runtime.Serialization;
 using Hazelcast.Exceptions;
+using Hazelcast.Core;
 
 namespace Hazelcast.Sql
 {
+    /// <summary>
+    /// Represents the exception that is thrown by the SQL service in case of an error.
+    /// </summary>
+    [Serializable]
```
mdumandag, line 26: Just a question: What does this annotation(?), or whatever it is called, do?

src/Hazelcast.Net/Core/TaskCoreExtensions.WithCancellation.cs
```diff
@@ -23,8 +23,8 @@ namespace Hazelcast.Core
 {
     internal static partial class TaskCoreExtensions
     {
-        public static ConfiguredCancelableAsyncEnumerable<T> WithCancellation<T>(this IAsyncEnumerable<T> enumerable, CancellationToken cancellationToken, bool throwOnCancel)
-            => new ConfiguredCancelableAsyncEnumerable<T>(enumerable, continueOnCapturedContext: true, cancellationToken, throwOnCancel);
+        public static ConfiguredCancelableAsyncEnumerable<T> WithCancellation<T>(this IAsyncEnumerable<T> enumerable, bool throwOnCancel, CancellationToken cancellationToken)
```
mdumandag, line 26: I am not really familiar with C#, it says internal at the top but I saw that we use `WithCancellation` in tests like `src/Hazelcast.Net.Tests/Sql/SqlQueryResultTests.cs`. So, if it is just an internal thing, please ignore this comment.

But if this API is public, can the change on the parameter order be a problem?

src/Hazelcast.Net/Sql/HazelcastSqlException.cs
zpqrtbnk: Good you caught it - I wanted to update this & forgot, going to update. I *think* I'll go with *two* fields, the client & the member, bearing in mind that one might be emtpy.

src/Hazelcast.Net/Sql/HazelcastSqlException.cs
zpqrtbnk: Such an annotation is called an "attribute" marking the class, in C#. It indicates the framework that the class can be serialized via the built-in binary serializer, if needed. It's usually a good idea to avoid this, but is recommended for exceptions.

src/Hazelcast.Net/Core/TaskCoreExtensions.WithCancellation.cs
zpqrtbnk: It could. But this is a fix to code that we haven't released yet. And, yes, the class is internal so users won't see it.

doc/dev/doc/canceling.md
```diff
@@ -0,0 +1,50 @@
...
+In both cases, the enumeration ends up being controlled by a `CancellationToken` which is either `token` (from `ExecuteQueryAsync`), `enumerationToken`, or a combination of both. If `token` or `enumerationToken` is canceled, the enumeration is canceled and throws an `OperationCancelledException`.
+
+It is important to understand that the `ISqlQueryResult` represents a server-side query which needs to be closed in order to release server resources. This is achieved by async-disposing the `ISqlQueryResult` (which is `IAsyncEnumerable`).
+
+If the result is enumerated with `await foreach`, the pattern guarantees that the enumerator will be disposed when exiting. This will, in turn, dispose the result. Should the enumeration be cancelled and abort with an exception, the result would therefore safely be disposed.
+
+With other patterns, it is important to ensure that, in the event of a cancellation, the result is always disposed. This can be achieved by wrapping the `ISqlQueryResult` in a `using` block:
+
+```csharp
+await using result = await client.Sql.ExecuteQueryAsync(...);
```
alexb5dh, line 42:
```suggestion
await using var result = await client.Sql.ExecuteQueryAsync(...);
```

src/Hazelcast.Net.Tests/Exceptions/RemoteExceptionTests.cs
```diff
@@ -32,13 +32,13 @@ public void Constructors()
...
             _ = new RemoteException("exception");
             _ = new RemoteException("exception", new Exception("bang"));
-            _ = new RemoteException(RemoteError.Undefined, true);
-            _ = new RemoteException(RemoteError.Undefined, "exception");
-            _ = new RemoteException(RemoteError.Undefined, "exception", true);
-            _ = new RemoteException(RemoteError.Undefined, new Exception("bang"));
-            _ = new RemoteException(RemoteError.Undefined, "exception", new Exception("bang"));
-            _ = new RemoteException(RemoteError.Undefined, new Exception("bang"), true);
-            var e = new RemoteException(RemoteError.AccessControl, "exception", new Exception("bang"), true);
+            _ = new RemoteException(Guid.Empty, RemoteError.Undefined, true);
```
alexb5dh, line 35: Maybe we should provide a constructor without `Guid`?

src/Hazelcast.Net/Core/SerializationInfoExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+    public static class SerializationInfoExtensions
+    {
+        /// <summary>
+        /// Retrieves a <see cref="Guid"/> value from the <see cref="SerializationInfo"/> store.
+        /// </summary>
+        /// <param name="info">The <see cref="SerializationInfo"/> store.</param>
+        /// <param name="name">The name of the value to retrieve.</param>
+        /// <returns>A <see cref="Guid"/> value.</returns>
+        public static Guid GetGuid(this SerializationInfo info, string name)
+            => (Guid)(info?.GetValue(name, typeof(Guid)) ?? throw new SerializationException());
```
alexb5dh, line 32: If value stored by `name` is not `Guid`, `InvalidCastException` will be thrown. Is this ok?

src/Hazelcast.Net/Models/HOffsetDateTime.cs
```diff
@@ -82,8 +83,8 @@ public override string ToString()
         {
             return Offset switch
             {
-                var s when s < TimeSpan.Zero => LocalDateTime + "-" + Offset.ToString(@"hh\:mm"),
-                var s when s > TimeSpan.Zero => LocalDateTime + "+" + Offset.ToString(@"hh\:mm"),
+                var s when s < TimeSpan.Zero => LocalDateTime + "-" + Offset.ToString(@"hh\:mm", CultureInfo.InvariantCulture),
```
alexb5dh, line 86: Do you think `CultureInfo.InvariantCulture` is needed? We specify quite a strict format here.

src/Hazelcast.Net/Models/HBigDecimal.cs
```diff
@@ -108,10 +110,14 @@ public string ToString(CultureInfo cultureInfo)
 
         public static bool TryParse(string s, CultureInfo cultureInfo, out HBigDecimal bigDecimal)
         {
+            if (cultureInfo == null) throw new ArgumentNullException(nameof(cultureInfo));
```
alexb5dh, line 113: Should we also check `s` in such case?

src/Hazelcast.Net/Protocol/RemoteException.cs
```diff
@@ -126,8 +154,19 @@ public RemoteException(RemoteError error, string message, Exception innerExcepti
...
+            ServerStackTrace = info.GetString(nameof(ServerStackTrace));
+        }
+
+        /// <summary>
+        /// Gets the unique identifier of the member which threw the exception.
+        /// </summary>
+        public Guid MemberId
+        {
+            get => _memberId;
+            set => Data[nameof(MemberId)] = _memberId = value;
```
alexb5dh, line 169: Why to use both dictionary and private field?

src/Hazelcast.Net.Tests/Exceptions/RemoteExceptionTests.cs
zpqrtbnk: The exception should always have a `memberId` which is a `Guid` but here we are testing (for code coverage) all constructors, so we just pass `Guid.Empty`. If we replaces these lines with a constructor that does not require a `Guid`... we end up not covering the constructors that require a `Guid` - and we want to cover them.

src/Hazelcast.Net/Core/SerializationInfoExtensions.cs
zpqrtbnk: Pushing a fix:
```
            if (info == null) throw new ArgumentNullException(nameof(info));
            return (Guid) info.GetValue(name, typeof (Guid));
```
According to code, `GetValue` either returns the specified type (`typeof(Guid)`) or throws, so the case is safe.

src/Hazelcast.Net/Models/HOffsetDateTime.cs
zpqrtbnk: Not "needed" but this does not hurt and it gets rid of a warning.

src/Hazelcast.Net/Protocol/RemoteException.cs
zpqrtbnk: totally, pushing a simplification

src/Hazelcast.Net/Sql/SqlService.cs
```diff
@@ -90,7 +104,7 @@ public async Task<long> ExecuteCommandAsync(string sql, object[] parameters = nu
             var responseMessage = await _cluster.Messaging.SendAsync(requestMessage, cancellationToken).CfAwait();
             var response = SqlExecuteCodec.DecodeResponse(responseMessage);
 
-            if (response.Error != null) throw new HazelcastSqlException(response.Error);
+            if (response.Error != null) throw new HazelcastSqlException(_cluster.ClientId, response.Error);
```
mdumandag, line 107: So, in this case, shouldn't we set the `clientId` to null or something to denote that the error is originated from the member, not the client?

src/Hazelcast.Net/Sql/HazelcastSqlException.cs
```diff
@@ -13,26 +13,109 @@
...
+            if (info == null) throw new ArgumentNullException(nameof(info));
+
+            info.AddValue(nameof(ClientId), ClientId);
+            info.AddValue(nameof(MemberId), MemberId);
+            info.AddValue(nameof(ErrorCode), ErrorCode);
+            base.GetObjectData(info, context);
+        }
+
+        /// <summary>
+        /// Get the identifier of the <see cref="IHazelcastClient"/> that caused the exception.
```
mdumandag, line 107: Are we doing this, if the error is not originated from the client, this is null, thing?

src/Hazelcast.Net/Sql/SqlService.cs
zpqrtbnk: my idea was: the client is always known, so we always set the client. if member is known, then it comes from the member, and we also set the member. want to know if the error is originated from the member? then the member id is not Guid.Empty.

src/Hazelcast.Net/Sql/HazelcastSqlException.cs
zpqrtbnk: Well, Actually, `Guid.Empty` which is 00000-00000-0000-000000 or something - probably worth documenting, then?

src/Hazelcast.Net/Sql/SqlService.cs
mdumandag: Oh, I see :+1:

src/Hazelcast.Net/Sql/HazelcastSqlException.cs
mdumandag: Yeah, I guess it makes sense to document the memberId behavior

mdumandag (approved):
LGTM

PR #510, opened 2021-09-16
Update targets (remove netcoreapp2.1, add net5.0) and dependencies (simplify Thrift)

zpqrtbnk:
~depends on #509, must rebase once #509 is merged~

Hazelcast .NET Client v5 does not build, ship nor test for .NET Core 2.1 as it has reached [end of life](https://dotnet.microsoft.com/platform/support/policy/dotnet-core). In addition, we were using a kudge for Thrift because we had to used different versions of Thrift generated code for .NET Framework and .NET Core. Latest Thrift versions now are NetStandard, and so support both. This allows us to simplify things.

There are quite a lot of modified files, but mainly because:
* All the Thrift-generated files are re-generated
* Many `#if NETSTANDARD` etc conditional statements need to be updated

There are no actual code changes in the PR.

NOTE: while reviewing this PR it would be good to have a look at [this Remote Controller PR](https://github.com/hazelcast/hazelcast-remote-controller/pull/33) which contains the updated Thrift code generation for .NET.


src/Hazelcast.Net.Tests/Remote/MapTests.cs
```diff
@@ -298,11 +298,11 @@ public async Task SetWithTimeToLive()
             // and does not return anything
             // NOTE: no way to know whether it added or replaced?
 
-            await map.SetAsync("key", 42, TimeSpan.FromSeconds(1)).CfAwait();
+            await map.SetAsync("key", 42, TimeSpan.FromSeconds(2)).CfAwait();
             var value = await map.GetAsync("key").CfAwait();
             Assert.AreEqual(42, value);
 
-            await Task.Delay(1500); // wait for 1.5 second
+            await Task.Delay(3000); // wait for 3 second
```
srknzl, line 305: Why change these timeouts?

src/Hazelcast.Net/Exceptions/StackTraceElement.cs
```diff
@@ -12,7 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#if NETSTANDARD2_1
+#if NETFRAMEWORK || NETSTANDARD2_0
+#else
```
srknzl, line 16: why have a empty if block?

hz.ps1
```diff
@@ -352,17 +352,28 @@ else {
     $docMessage = "Version $($options.version) documentation"
 }
 
-# determine framework(s)
-$frameworks = @( "net462", "netcoreapp2.1", "netcoreapp3.1" )
+# determine framework(s) - for building and running tests
+$frameworks = @( "net462", "net48", "netcoreapp3.1", "net5.0" )
```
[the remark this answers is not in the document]
zpqrtbnk: `net462` is the eldest version we support, `net48` is the latest .NET Framework version, so I thought it'd be good to test both.

src/Hazelcast.Net.Tests/Remote/MapTests.cs
zpqrtbnk: I hate these tests. Testing for TTL takes time and is totally unreliable. What happened was, on GitHub environment, the test could run so slow that the TTL would expire on the server before we have a chance to get the value back. So... I changed them. All those tests that rely on timing are really bad but I don't know how do to it differently?

src/Hazelcast.Net/Exceptions/StackTraceElement.cs
zpqrtbnk: ah... because in the past it was not empty, but yes I could change this to:
```
#if !NETFRAMEWORK && !NETSTANDARD2_0
```
will do

srknzl (approved):

src/Hazelcast.Net.Testing/Hazelcast.Net.Testing.csproj
```diff
@@ -21,14 +17,10 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="ApacheThrift" Version="0.13.0.1" />
+    <PackageReference Include="ApacheThrift" Version="0.15.0" />
```
srknzl, line 20: @zpqrtbnk A question: Did you used 0.15.0 version of thrift compiler in this pr to generate remote controller bindings?

Issue #469, opened 2021-05-27
Dll hell


drieseng:
We're running into an issue consuming the Hazelcast .NET client in an ASP.NET WCF web services application.
I'm still digging to find what the exact cause is, but one of the things we noticed is that different components used by the Hazelcast  client expect different versions of **System.Runtime.CompilerServices.Unsafe**.

All components (assemblies) were compiled against version **4.0.4.1** of **System.Runtime.CompilerServices.Unsafe**, except for **Microsoft.Extensions.Primitives**. The latter was compiled against version **5.0.0.0** of **System.Runtime.CompilerServices.Unsafe**.

A brief summary:

Assembly|Depends on version of ...Unsafe|
----------|---------------------|
|System.Memory|4.0.4.1|
|System.Threading.Tasks.Extensions|4.0.4.1|
|Microsoft.Extensions.Primitives|5.0.0.0|

The dependency on **Microsoft.Extensions.Primitives** comes from **Microsoft.Extensions.Configuration.Json** which in turns references a few other assemblies that in turn depend on version **5.0.0.0** of **System.Runtime.CompilerServices.Unsafe**.

We can of course work around this using assembly binding redirects, but I thought it was interesting information nonetheless as it may explain our issue in the ASP.NET WCF web service application since this is our only application in which we effectively had to add the assembly binding redirect.

All our other applications also target .NET Framework 4.7.2 so they all use the same target framework version of Hazelcast (and its dependencies).

drieseng:
@zpqrtbnk I'll keep digging to find the cause of the problem we have in the ASP.NET WCF web services application (and open a new issue if/when I have more info), but do you think this issue is one of the possible causes?

zpqrtbnk:
I somehow missed this issue. Going to look into it.

zpqrtbnk:
OK, I am not *actually* looking into it. IIC we have

Microsoft.Extensions.Configuration.Json 
🡢 (assemblies) 
🡢 Microsoft.Extensions.Primitives 
🡢 System.Runtime.CompilerServices.Unsafe version **5.0.0**

whereas other dependencies want version 4.x of that library.

And there's a similar situation with System.Buffers.

I *think* we should provide a NuGet package that does not require binding redirects, probably by up/downgrading some of our packages. Working on it.

zpqrtbnk:
>I think we should provide a NuGet package that does not require binding redirects, probably by up/downgrading some of our packages. Working on it.

[...]

PR #512, opened 2021-09-21
Cleanup "DLL Hell"

zpqrtbnk:
Depends on #510

This addresses #469 by revisiting our dependencies and/or documenting the required binding redirects for .NET Framework.

### DIFF
--- a/doc/dev/doc/getting-started.md
+++ b/doc/dev/doc/getting-started.md
@@ -1,5 +1,30 @@
 # Getting Started
 
+## Download and Install
+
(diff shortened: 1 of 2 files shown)


doc/dev/doc/getting-started.md
```diff
@@ -1,5 +1,30 @@
 # Getting Started
 
+## Download and Install
+
```
srknzl, line 4: Is the reason of adding this information to this page and not to "download and install" document making it more visible? 

If it is related to download and install, we can directly put it there. I left to you

srknzl (approved):

PR #534, opened 2021-11-03
Pws white space fix

emreyigit:
I faced a problem when a java class path contains white spaces in it, controller and server doesn't work. Quoted the all usages of `$script:options.classpath`, and tested with run-server and test commands.

### DIFF
--- a/hz.ps1
+++ b/hz.ps1
@@ -555,6 +555,9 @@ function ensure-jar ( $jar, $repo, $artifact ) {
     $classpath = $script:options.classpath
     if (-not [System.String]::IsNullOrWhiteSpace($classpath)) { $classpath += $s }
     $classpath += "$tmpDir/lib/$jar"
+    # condiser to quote the path to escape from white space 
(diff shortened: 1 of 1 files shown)


hz.ps1
```diff
@@ -555,6 +555,9 @@ function ensure-jar ( $jar, $repo, $artifact ) {
     $classpath = $script:options.classpath
     if (-not [System.String]::IsNullOrWhiteSpace($classpath)) { $classpath += $s }
     $classpath += "$tmpDir/lib/$jar"
+    # condiser to quote the path to escape from white space
```
zpqrtbnk, line 558: You probably want this to be "consider" ;) And, maybe it's even "be sure to..."?

zpqrtbnk:
Have added 1 minor comment for you, rest sounds good.

hz.ps1
emreyigit: You are right, that could better :)

zpqrtbnk (approved):
Looks good to me.

PR #536, opened 2021-11-05
map creating  command added

emreyigit:
Missing part to query over a map is added to SQL examples.


zpqrtbnk (approved):
Sorry I missed these during v5 release. All good for me.

PR #537, opened 2021-11-08
lazy deserialization impl.

emreyigit:
Lazy sql deserialization is implemented.

### DIFF
--- a/src/Hazelcast.Net/Sql/SqlRow.cs
+++ b/src/Hazelcast.Net/Sql/SqlRow.cs
@@ -47,7 +47,14 @@ namespace Hazelcast.Sql
         /// <summary>
         /// Gets the value of the column by identified by index
         /// </summary>
-        /// <typeparam name="T"></typeparam>
+        /// <typeparam name="T">The type of the column value.</typeparam>
+        /// <param name="index">The index.</param>
+        /// <returns>The value of the column.</returns>
+        /// <remarks>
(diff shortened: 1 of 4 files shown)


zpqrtbnk:
Check build failing for some silly reason... investigating.

src/Hazelcast.Net/Sql/SqlRow.cs
```diff
@@ -43,12 +45,22 @@ public SqlRow(IList<object> values, SqlRowMetadata metadata)
         public SqlRowMetadata Metadata { get; }
 
         /// <summary>
-        /// Gets th
+        /// Gets the value of the column by identified by index
         /// </summary>
         /// <typeparam name="T"></typeparam>
```
zpqrtbnk, line 50:
```suggestion
        /// <typeparam name="T">The type of the column value.</typeparam>
        /// <param name="index">The index.</param>
        /// <returns>The value of the column.</returns>
        /// <remarks>
        /// <para>For performance purposes and to reduce allocations, the column value is deserialized 
        /// but NOT cached. Avoid getting the value of a column multiple times. Cache it in a local variable
        /// instead, if needed.</para>
        /// </remarks>
```

zpqrtbnk (changes requested):
See my suggestion for commenting that there may be a performance impact if getting the value of a column multiple times.

zpqrtbnk (approved):

PR #541, opened 2021-11-23
suggestion field implemented to sql exception

emreyigit:
suggestion field implemented to sql exception

### DIFF
--- a/src/Hazelcast.Net.Tests/Sql/SqlQueryResultTests.cs
+++ b/src/Hazelcast.Net.Tests/Sql/SqlQueryResultTests.cs
@@ -149,7 +149,6 @@ namespace Hazelcast.Tests.Sql
         }
 
         [Test]
-
         public async Task TestSqlLazyDeserializationThrowsExceptionAtGetValue()
         {
             var map = await Client.GetMapAsync<int, DummyPortable>(nameof(SqlQueryResultTests));
@@ -186,7 +185,7 @@ namespace Hazelcast.Tests.Sql
 
             try
             {
-                //query the map without creating map to get exception with suggestion in it.
+                //query the map without creating mapping to get exception with suggestion in it.
                 var result = await Client.Sql.ExecuteQueryAsync($"SELECT * FROM {dummyMapName}");
             }
             catch (HazelcastSqlException ex)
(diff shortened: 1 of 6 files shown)


src/Hazelcast.Net.Tests/Sql/SqlQueryResultTests.cs
```diff
@@ -173,5 +175,25 @@ await foreach (var row in result)
...
+
+            try
+            {
+                //query the map without creating map to get exception with suggestion in it.
+                var result = await Client.Sql.ExecuteQueryAsync($"SELECT * FROM {dummyMapName}");
+            }
+            catch (HazelcastSqlException ex)
+            {
+                Assert.IsNotEmpty(ex.Message);
+                Assert.IsFalse(string.IsNullOrEmpty(ex.Suggestion));
```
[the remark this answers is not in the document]
emreyigit: Don't you think that checking the content is kind of testing the server side? I mean I am just trying to be sure that the client got and set the suggestion message.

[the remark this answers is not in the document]
alisengul53: I felt it would be more safe to test content but right it should already be tested in the server side 👍

alisengul53 (approved):
Looks good!

src/Hazelcast.Net.Tests/Sql/SqlQueryResultTests.cs
```diff
@@ -148,6 +149,7 @@ public async Task DisposeDuringQuery()
         }
 
         [Test]
+
```
zpqrtbnk, line 152: we don't need this empty line ;)

src/Hazelcast.Net.Tests/Sql/SqlQueryResultTests.cs
```diff
@@ -173,5 +175,25 @@ await foreach (var row in result)
...
+        [ServerConditionAttribute("5.0")]
+        public async Task SqlErrorHasSuggestion()
+        {
+            string dummyMapName = "testingMap";
+            var map = await Client.GetMapAsync<int, string>(dummyMapName);
+            await map.PutAsync(0, "some value");
+
+            try
+            {
+                //query the map without creating map to get exception with suggestion in it.
```
zpqrtbnk, line 189: without creating _mapping_, right? because we created the map.

zpqrtbnk (changes requested):
Two minor change requests and then I am OK.

src/Hazelcast.Net.Tests/Sql/SqlQueryResultTests.cs
emreyigit: yeah, good catch!

zpqrtbnk (approved):
All good.

Issue #543, opened 2021-11-29
Adding more logging for the first connection attempt


emreyigit:
One of client had trouble during the first connection. In order to ease debugging on the prod, some of the steps may logged more detailed. 

- Sign of generating of ports -> when no port number is supplied.
- During timeout of very first attempt -> client tries to open connection to member agelessly over thousands of ports. (Occurred when ldap response delayed)

Issue #540, opened 2021-11-22
Refactor ILogger usage patterns [API-1008]


zpqrtbnk:
In many places we have code that does:
```
_logger.LogDebug($"this is {xyz} an interpolated string");
```
And this is a bad thing because it interpolates the string, even though it may not be used. The latest Roslyn analyzers detect it and we even have a suppression rule to get rid of the warnings. This is really bad.

Either move to log message templates:
```
_logger.LogDebug("This is {Value} a constant string", xyz);
```
Combined (?) with:
```
_logger.IfDebug()?.LogDebug(...);
```

PR #547, opened 2021-12-01
Networking fixes [API-1020]

zpqrtbnk:
See linked issue for details.

### DIFF
--- a/src/Hazelcast.Net/Clustering/ClusterConnections.cs
+++ b/src/Hazelcast.Net/Clustering/ClusterConnections.cs
@@ -37,7 +37,6 @@ namespace Hazelcast.Clustering
         private readonly ClusterState _clusterState;
         private readonly ClusterMembers _clusterMembers;
         private readonly Authenticator _authenticator;
-        private readonly AddressProvider _addressProvider;
         private readonly IRetryStrategy _connectRetryStrategy;
         private readonly ILogger _logger;
 
@@ -68,7 +67,6 @@ namespace Hazelcast.Clustering
 
             _logger = _clusterState.LoggerFactory.CreateLogger<ClusterConnections>();
             _authenticator = new Authenticator(_clusterState.Options.Authentication, serializationService, _clusterState.LoggerFactory);
-            _addressProvider = new AddressProvider(_clusterState.Options.Networking, _clusterState.LoggerFactory);
             _connectRetryStrategy = new RetryStrategy("connect to cluster", _clusterState.Options.Networking.ConnectionRetry, _clusterState.LoggerFactory);
 
             if (_clusterState.IsSmartRouting)
@@ -453,7 +451,7 @@ namespace Hazelcast.Clustering
                 yield return address;
 
             // get configured addresses that haven't been tried already
-            addresses = _addressProvider.GetAddresses();
+            addresses = _clusterState.AddressProvider.GetAddresses();
             foreach (var address in Distinct(addresses, distinct, shuffle))
                 yield return address;
         }
@@ -693,8 +691,7 @@ namespace Hazelcast.Clustering
         /// <returns>A task that will complete when the connection has been established, and represents the associated client.</returns>
         private async Task<MemberConnection> ConnectAsync(NetworkAddress address, CancellationToken cancellationToken)
         {
-            // map private address to public address
-            address = _addressProvider.Map(address);
+            // directly connect to the specified address (internal/public determination happened beforehand)
 
             // create the connection to the member
             var connection = new MemberConnection(address, _authenticator, _clusterState.Options.Messaging, _clusterState.Options.Networking, _clusterState.Options.Networking.Ssl, _clusterState.CorrelationIdSequence, _clusterState.LoggerFactory)
--- a/src/Hazelcast.Net/Clustering/ClusterMembers.cs
+++ b/src/Hazelcast.Net/Clustering/ClusterMembers.cs
@@ -348,7 +348,7 @@ namespace Hazelcast.Clustering
             // through configuration options, or determined through rules:
             // - if one member responds on its internal address, assume internal addresses are OK
             // - if enough (sample size) members respond only on their public address, use public addresses
-            // for performance reasons (and this is what other clients e.g. Java do) we determine this
+            // for performance reasons (and this is what the Java client does) we determine this
             // once when getting the first members view, and don't change our mind later on, ever.
             if (previous.Count == 0) // first members view
             {
--- a/src/Hazelcast.Net/Clustering/ClusterMembers.cs
+++ b/src/Hazelcast.Net/Clustering/ClusterMembers.cs
@@ -287,12 +287,19 @@ namespace Hazelcast.Clustering
             msg.Append("Members [");
             msg.Append(table.Count);
             msg.AppendLine("] {");
-            foreach (var member in table.Members)
+            foreach (var (m, d) in diff)
             {
                 msg.Append("    ");
-                msg.Append(member.ToShortString(true));
-                if (diff.TryGetValue(member, out var d) && d == 2)
-                    msg.Append(" - new");
+                msg.Append(m.ToShortString(true));
+                var status = d switch
+                {
+                    1 => "Removing",
+                    2 => "Adding",
+                    3 => "Unchanged",
+                    _ => ""
+                };
+                msg.Append(' ');
+                msg.Append(status);
                 msg.AppendLine();
             }
             msg.Append('}');
@@ -326,8 +333,24 @@ namespace Hazelcast.Clustering
             // anymore, see note in GetMember)
             _loadBalancer.SetMembers(members.Select(x => x.Id));
 
-            // if this is the initial list of members, determine how to connect to members
-            if (previous.Count == 0)
+            // members provided through the members view event always provide their own internal address,
+            // and may (e.g. for Kubernetes clusters) also provide their public address - but in a Cloud
+            // setup for instance, members are not aware of their public address. so, we try to "fix"
+            // the members using the AddressProvider map, if any.
+            if (_clusterState.AddressProvider.HasMap)
+            {
+                foreach (var member in members.Where(x => !x.HasPublicAddress))
+                    member.PublicAddress = _clusterState.AddressProvider.Map(member.Address);
+            }
+
+            // and then we need to determine whether to connect to members through their internal address
+            // or their public address - this is the role of the ConnectAddressResolver - it can be forced
+            // through configuration options, or determined through rules:
+            // - if one member responds on its internal address, assume internal addresses are OK
+            // - if enough (sample size) members respond only on their public address, use public addresses
+            // for performance reasons (and this is what other clients e.g. Java do) we determine this
+            // once when getting the first members view, and don't change our mind later on, ever.
+            if (previous.Count == 0) // first members view
             {
                 var resolver = new ConnectAddressResolver(_clusterState.Options.Networking, _clusterState.LoggerFactory);
                 if (!(members is IReadOnlyCollection<MemberInfo> mro)) throw new HazelcastException("panic"); // TODO: not exactly pretty
@@ -357,8 +380,8 @@ namespace Hazelcast.Clustering
                     else diff[m] = 2;
             }
 
-            // log, if the members have changed (one of them at least is not 3=unchanged)
-            if (_logger.IsEnabled(LogLevel.Information) && diff.Any(d => d.Value != 3))
+            // log
+            if (_logger.IsEnabled(LogLevel.Information))
                 LogDiffs(table, diff);
 
             // process changes, gather events
--- a/src/Hazelcast.Net/Clustering/ClusterMembers.cs
+++ b/src/Hazelcast.Net/Clustering/ClusterMembers.cs
@@ -221,7 +221,7 @@ namespace Hazelcast.Clustering
                 // if we are not disconnecting, we can return - we are done
                 if (!disconnecting)
                 {
-                    _logger.LogDebug($"Removed connection {connection.Id.ToShortString()} to member {connection.MemberId.ToShortString()}, remain connected.");
+                    _logger.LogDebug($"Removed connection {connection.Id.ToShortString()} to member {connection.MemberId.ToShortString()}, remain {(_connected?"":"dis")}connected.");
 
                     // if we are connected,
                     // and the disconnected member is still a member, queue it for reconnection
--- a/src/Hazelcast.Net/Networking/AddressProvider.cs
+++ b/src/Hazelcast.Net/Networking/AddressProvider.cs
@@ -103,9 +103,9 @@ namespace Hazelcast.Networking
         public bool HasMap { get; }
 
         // ensures that we have a map, returns the map + whether it's a new map
-        private (bool NewMap, IDictionary<NetworkAddress, NetworkAddress> Map) EnsureMap()
+        private (bool NewMap, IDictionary<NetworkAddress, NetworkAddress> Map) EnsureMap(bool forceRenew)
         {
-            if (_internalToPublicMap != null) return (false, _internalToPublicMap);
+            if (!forceRenew && _internalToPublicMap != null) return (false, _internalToPublicMap);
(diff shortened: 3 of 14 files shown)


src/Hazelcast.Net/Clustering/ClusterMembers.cs
```diff
@@ -221,7 +221,7 @@ public async Task RemoveConnectionAsync(MemberConnection connection)
                 // if we are not disconnecting, we can return - we are done
                 if (!disconnecting)
                 {
-                    _logger.LogDebug($"Removed connection {connection.Id.ToShortString()} to member {connection.MemberId.ToShortString()}, remain connected.");
+                    _logger.LogDebug($"Removed connection {connection.Id.ToShortString()} to member {connection.MemberId.ToShortString()}, remain {(_connected?"":"dis")}connected.");
```
emreyigit, line 224: I haven't finished the ILogger Refactoring but for further usages, we may get into habit of using `(ifDebug/ifWarning)?.` and not to use string interpolations. fyi. After release, I'll also include these to refactoring.

src/Hazelcast.Net/Networking/AddressProvider.cs
```diff
@@ -169,7 +210,19 @@ internal static IEnumerable<IPAddress> GetV6LocalAddresses()
...
+        /// <param name="address">The address.</param>
+        /// <param name="ipAddress"></param>
+        /// <returns>Addresses with ports.</returns>
+        /// <remarks>
+        /// <para>If the <paramref name="address"/> has a specified port, this yields the address, and
+        /// only the address. But if it does not have a specified port, this yields the address with
+        /// different ports obtained through the <see cref="NetworkingOptions.DefaultPort"/> and
+        /// <see cref="NetworkingOptions.PortRange"/> configuration options.</para>
+        /// </remarks>
+        private IEnumerable<NetworkAddress> ExpandPorts(NetworkAddress address, IPAddress ipAddress = null)
```
emreyigit, line 225: I think that would be great to have log here whether ports are expended or not. We struggled about that production side of a client last week. -> #543 

But we can save it to issue #543. I just wanted to mention when I saw this method.

src/Hazelcast.Net/Clustering/ClusterConnections.cs
```diff
@@ -453,7 +451,7 @@ static IEnumerable<NetworkAddress> Distinct(IEnumerable<NetworkAddress> aa, ISet
                 yield return address;
 
             // get configured addresses that haven't been tried already
-            addresses = _addressProvider.GetAddresses();
+            addresses = _clusterState.AddressProvider.GetAddresses();
```
emreyigit, line 454: I didn't get why `AddressProvider` is moved to `ClusterState`.

zpqrtbnk: Because it's now used both by `ClusterConnections` and `ClusterMembers` and at one point it seemed clever. Now, `ClusterConnections` has a reference to `ClusterMembers` so we *could* have `AddressProvider` owned by `ClusterMembers` instead. I am just not too sure. What do you think?

src/Hazelcast.Net/Networking/AddressProvider.cs
zpqrtbnk: Let's do it in #543 yes. Wasn't aware of the issue.

src/Hazelcast.Net/Clustering/ClusterMembers.cs
zpqrtbnk: Guess we should either use `IfDebug()?.` here, or template strings `_logger.LogDebug("Connection {ConnectionId}...", id);` - want me to change it here, or will you take care of it with your refactoring?

emreyigit: I'll refactor them, np. I'm changing #540 to draft for now. After this is merged, I'll cover it.

src/Hazelcast.Net/Clustering/ClusterConnections.cs
emreyigit: Yeah, it's clever.

emreyigit (approved):
Looks nice!

alisengul53 (approved):
Looks good with a small comment

ihsandemir:
@zpqrtbnk Can you please elaborate the fixes in this PR in the PR description. what are you fixing and how are you fixing? Please some details.

zpqrtbnk:
So, the PR links to an issue, and what we are fixing is usually described in the issue. The problem here is that in some cases, networking can break:
* Can report "remain connected" even though the client hasn't connected yet
* Can freeze, waiting to be connected but doing nothing

Many of the files that have changed have only cosmetic or non-functional changes :adding some logging, some comments, fixing some logged messages that were wrong, etc. The files that have meaningful changes would be:

`ClusterConnections.cs` - `ConnectAsync` used to map the passed addresses using the address provider - it now accepts the addresses unchanged - the map is now applied when we receive the members (see `ClusterMembers.cs`).

`ClusterMembers.cs` - we used to accept the members "as is" when received via members view listener - we now assign each member a public address if (1) the member does not have a public address already and (2) the address provider provides an internal to public map.

When running Cloud Discovery, it will build and provide a map to the address provider - it is Cloud which tells us the internal and public addresses of each members. And then, when we receive the list of members, if running in Cloud, each member will _not_ provide a public address. With these changes in place, we "fix" the situation so that we can treat members "normally" later on instead of having weird exceptions in code for Cloud.

`ConnectAddressResolver.cs` - is the thing that resolves internal / public addresses when running Kubernetes. Mostly cosmetic and logging changes, plus making some constants optional (but internal = not visible to users, it's just cleaner).

src/Hazelcast.Net/Networking/AddressProvider.cs
```diff
@@ -68,49 +85,73 @@ public AddressProvider(NetworkingOptions networkingOptions, ILoggerFactory logge
...
+
         /// <summary>
         /// Gets known possible addresses for a cluster.
         /// </summary>
         /// <returns>All addresses.</returns>
         public IEnumerable<NetworkAddress> GetAddresses()
         {
-            _privateToPublic ??= _createMap() ?? throw new HazelcastException("Failed to obtain an address map.");
-            return _privateToPublic.Keys;
+            _internalToPublicMap ??= _createMap() ?? throw new HazelcastException("Failed to obtain addresses.");
```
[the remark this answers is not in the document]
ihsandemir: I think that there may be another more important problem here. [Java client reloads all the addresses](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/connection/tcp/TcpClientConnectionManager.java#L559) in each call but you just use the cached one here, right? This means that `cloudScanner.Scan()` is only called once, but we expect it to be called each time the client is disconnected from cluster to re-connect to the cluster.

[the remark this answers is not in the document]
zpqrtbnk: `a ??= b` is equivalent to `if (a == null) { a = b; }`
`a = b ?? c` is equivalent to `a = b` if `c` is not null, else use `c` / do `c`

so : if we have a map already, do nothing. if we do not have a map, create it, and if it's created `null`, throw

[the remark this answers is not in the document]
zpqrtbnk: Let's check these things:
* in Java, we reload the address map if we get disconnected.
we should reset the map in .NET too when disconnected - then it will be fetched again.

* re. `?? throw` - it may not happen today but may happen in the future
but we need a way to make this coverable

src/Hazelcast.Net/Clustering/ClusterMembers.cs
```diff
@@ -326,8 +333,24 @@ public async Task<MembersUpdatedEventArgs> SetMembersAsync(int version, ICollect
...
+                foreach (var member in members.Where(x => !x.HasPublicAddress))
+                    member.PublicAddress = _clusterState.AddressProvider.Map(member.Address);
+            }
+
+            // and then we need to determine whether to connect to members through their internal address
+            // or their public address - this is the role of the ConnectAddressResolver - it can be forced
+            // through configuration options, or determined through rules:
+            // - if one member responds on its internal address, assume internal addresses are OK
+            // - if enough (sample size) members respond only on their public address, use public addresses
+            // for performance reasons (and this is what other clients e.g. Java do) we determine this
```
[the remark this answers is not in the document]
zpqrtbnk: Updated the text + `usePublicAddresses` does not have to be `true` or `false` it can be `null` which means: please figure things out. Which is precisely what we do here.

src/Hazelcast.Net/Networking/AddressProvider.cs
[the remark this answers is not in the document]
zpqrtbnk: Update:

I have pushed a commit that deals with the second point (code coverage etc)

As for "reloading the map if disconnected" here what I see in Java:

There is a `RemoteAddressProvider` class that obtains addresses from Cloud Discovery (`privateToPublic = getAddresses.call()`) once and then uses them. If it fails to translate an address, it will run discovery again. But as long as it can translate addresses, regardless of whether the client disconnects, it keeps the same addreses.

.NET works the same.

That address provider is created by `ClusterDiscoveryServiceBuilder.build` which creates the `RemoteAddressProvider` once and stick with it, never changes it. `build` is invoked from `HazelcastClientInstanceImpl` `initClusterDiscoveryService` which itself is invoked from the constructor.

So... unless we create a new client, I fail to see where, should the client disconnect, the address map would be re-fetched. As far as I can see the address map would be re-fetched only if we try and fail to translate an addfress. And .NET implements this, see the `Map` method.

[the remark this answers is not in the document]
ihsandemir: In Java, `privateToPublic` is modified in two places `loadAddresses` and `translate`. .NET is missing what is being done in  `loadAddresses`, right?

src/Hazelcast.Net/Networking/AddressProvider.cs
```diff
@@ -85,20 +85,36 @@ public AddressProvider(NetworkingOptions networkingOptions, ILoggerFactory logge
             }
         }
 
+        // internal constructor for tests (so we can test with a bogus create map method)
+        internal AddressProvider(NetworkingOptions networkingOptions, Func<IDictionary<NetworkAddress, NetworkAddress>> createMap, bool hasMap, ILoggerFactory loggerFactory)
```
ihsandemir, line 89: You are putting extra code into the production code even if it is marked as `internal` which is only needed for testing (I understand that it can not be used by the lib users). Will it not get placed in the produced used dll? Is there no other way such as stubbing a class in .NET you can use and not add extra code to production?

ihsandemir (changes requested):

src/Hazelcast.Net/Networking/AddressProvider.cs
zpqrtbnk: No. If that class is to have that constructor, then... it needs to be here and then yes, the code goes in the production DLL. I don't know how to test otherwise.

src/Hazelcast.Net/Networking/AddressProvider.cs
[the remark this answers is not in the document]
zpqrtbnk: In Java, `privateToPublic` is modified in `loadAddresses` (always re-generated) and in `translate` (re-generated if mapping fails). And, `loadAddresses` is invoked by `getPossibleMemberAddresses` each time we get addresses to try to connect the client to the cluster.

In .NET, `_internalToPublicMap` is modified via `EnsureMap` in `GetAddresses` *if* it does not exist already, and in `Map` (re-generated if mapping fails). 

So... my fix broke it, and now we're not in-line with Java anymore. Or, let's not say "with Java", let's say, with some "specs" that say that (1) get a new map anytime you connect the client and cache it and (2) reuse it to map, unless map fails, then try to get a new map.

Pushing a fix now.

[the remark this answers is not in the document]
zpqrtbnk: Pushed + added tests to verify that :
* the map is updated each time `GetAddresses` is invoked
* the map is then cached and reused when mapping
* the map is created once and cached, when mapping, if it wasn't existing yet
* if mapping fails, and the map existed, the map is re-created (second chance)
* if mapping fails, and the map did not exist and was just created, the map is not re-created

src/Hazelcast.Net/Networking/AddressProvider.cs
ihsandemir: It is usually done via `mock` libraries. I believe there are mock frameworks in .NET, at least when I searched for it on the internet, e.g. [this](https://stackoverflow.com/questions/20400734/how-do-i-mock-a-class-without-an-interface) one. I will question this approach in the team to check how other clients are. I believe it is important not to ship test code for a production library but let's get some ideas on this.

yemreinci: I think this code already does mocking. The tests can "mock" the createMap interface by passing a custom implementation via this ctor. But instead of using interfaces/virtual functions, it is using a function pointer (or whatever .net calls them).

I think the way it works is a bit confusing because the AddressProvider provides a default implementation if you don't want to mock it. Something like this could be clearer:

```
class AddressProvider {
    // have just one ctor:
    AddressProvider(IList<string> addresses, Func<...> createMap);
}

// put the default createMap method somewhere outside, and provide it to AddressProvider during construction
IDictionary<...> defaultCreateMap(cloudConfig) {
   // implement the "if cloud enabled ... else ..." logic here
}

// in tests
IDictionary<...> myMockCreateMap() {
  return {"bogus": "mapping"};
}

var p = new AddressProvider({}, myMockCreateMap)
```

zpqrtbnk: Well, one way or another, the code is there in the production release. We want to be able to inject a `createMap` implementation, we need a constructor to support it. Now we can refactor the existing way to use that constructor and then say, see, it's code that is used, but apart from making things more complex I fail to see how it can help.

I also fail to see how mocking would change anything here. In fact, it is precisely what we are doing: supporting injecting a `createMap` method, including a bogus one. What that new constructor does is, precisely, support mocking.

One way to deal with this is to `#ifdef` the code, and have a "production release" build that is different from a "test release" build - but that sounds quite error prone to me. It means that the production binary we release is not what we test.

At that point, I fail to see how I can test what we are trying to test, without adding that ctor. Ideas welcomed.

zpqrtbnk: Might have an idea actually. Stay tuned.

zpqrtbnk: There, change pushed, now the `AddressProvider` relies on a source, which can be mocked in the test project. So, no more "dead" code in the production dll. Let me know what you think, it's probably cleaner.

yemreinci: > Well, one way or another, the code is there in the production release. We want to be able to inject a createMap implementation, we need a constructor to support it. Now we can refactor the existing way to use that constructor and then say, see, it's code that is used, but apart from making things more complex I fail to see how it can help.

I agree that it makes no practical difference (you still have the code but now it is "used") and I was completely fine with the way it previously was. I was only trying to suggest a compromise that would satisfy both concerns: no dead code + testable via mocking. 

I think everyone should be perfectly fine with the current version.

zpqrtbnk: Hey, let me know what you think of the latest changes.

src/Hazelcast.Net/Networking/AddressProvider.cs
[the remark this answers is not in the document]
ihsandemir: Are these 3 cases for when connecting to cluster? I don't understand how caching relates to the cluster connection process. Caching can relate only when we do a map.put/get kind of operations but those should also be using the addresses from the cluster member list. Is this an extra feature here? Does it exist in Java? Let's talk about it. 

Otherwise, the fix looks good.

ihsandemir (approved):

PR #548, opened 2021-12-01
Metrics / cleanup, fix client type, use CSP code not CSHARP [API-996]

zpqrtbnk:
(else it breaks Management Center and more)

See linked issue for details.


emreyigit (approved):

ihsandemir:
The linked issue does not have any information about the problem and what caused the problem and the details.

I am seeing the PR descriptions not including enough info. Can we force these during PR creation using a template? We need at least the followings:

1.  Problem statement
2. How it is solved.
3. Did you add a test for the new solution?
4. Is there a doc update needed?
5. Should it be included in release notes?

src/Hazelcast.Net/Core/ClientVersion.cs
```diff
@@ -53,5 +57,33 @@ internal static string Version
                 return _clientVersion;
             }
         }
+
+        /// <summary>
+        /// Gets the "pure" version of this assembly.
```
ihsandemir, line 62: How does this change relate to this PR?

zpqrtbnk: Yes this relate to this PR because we send the client version to MC, and a .NET client version is `4.2.6+a3cde44` which is valid semver but MC does not like it and "dies" and we need to send the "pure" version ie `4.2.6`. See changes in `ClientMetricSource`.

zpqrtbnk:
## Problems statement

* Client version does not show properly in MC
* The NearCache section of MC totally breaks (red JavaScript error message)

## Analysis & Resolution

Client version does not show properly in MC because the .NET client version is e.g. `4.2.6+aebc32` which means `4.2.6` built from commit `aebc32`. It's a common versioning scheme in .NET and it is semver-compliant. Unfortunately MC is not, and requires that we send a "pure" version e.g. `4.2.6`.

We fix this by adding a `PureVersion` property to the `ClientVersion` class in addition to the already-existing `Version` property. The `PureVersion` property strips whatever MC does not like. We use that property in `ClientMetricSource` where we do:
```csharp
yield return MetricDescriptors.ClientVersion.WithValue(ClientVersion.PureVersion);
```

The NearCache section of MC breaks because it looks up the client type in a JavaScript map and fails to find it. It fails to find it because we send `CSHARP` instead of `CSP`. There is no detection at any level, on the Java side, or this error, so it bubbles somewhere in MC UI.

We fix this by sending `CSP` instead of `CSHARP`, making it a documented constant in `ClientMetricSource`.

## Impacted Files

In addition to the two fixes mentioned above, this PR cleans metrics so that future maintenance becomes easier. Some files are edited with more comments, better debugging support, etc. Files that have significant changes are:

`MetricCompressor` was not sending "targets" which, although obsolete, still need to be sent as "empty". This was not detected but may cause MC to fail displaying metrics under some circumstances.

`Metric<TValue>` may look like it has changed a lot but it's just C# code reorg.

`IMetricDescriptor` and `MetricDescriptor` have been cleaned up to simplify the creation of metrics + better deal with the dual-formats of metrics (the old text one and the new binary one).

`MetricExtensions` has just been renamed, it provides helper methods for writing metrics to string.

`NearCacheStatistics` has been refactored following changes in `IMetricDescriptor` and `MetricDescriptor` to reduce confusion. `ClientMetricSource` has been refactored for the same reasons. All in all the idea is to have cleaner code.

In `HazelcastOptions.ClusterOptions` the default .NET client name has been changed from `hz.client.N` to `hznet.client.N` so that it's more obvious in MC that a client is .NET. We have verified that not all clients follow the `hz_client` pattern already and that it does not break things.

The `HazelcastOptionsBuilder` and `HazelcastOptions.Build` have been extended to support pre-options. This allows the user to programmatically set options that still can be overriden by, say, command-line arguments or environment variables. Whereas up until now, any change in options via code would always override everything else. This is not purely related to metrics but helped with testing. Some tests have been updated due to this. This _could_ have been another PR, really, as it's more technical debt work that metrics. But... it's here.

The `DumpCoreExtensions` have been slightly refactored - they provide ways to dump arrays of bytes to text, in debugging builds, and are not used in release builds.

`CloudTests` and some java files have been greatly updated to test and validate that what we send for metrics is correct. We now test that we can compress metrics _and_ that we can uncompress them to correct values, both in C# and using the original Java code.

Finally, the `LongRunningCloudClient` has been updated to benefit from all these changes. It's a simple (well...) example that can be run easily against Cloud and MC and will read/write to a map while posting metrics. It has been used extensively to validate that this PR fixes the problems, since ... the only way we have at the moment to do it is via this kind of giant integration test, due to very little error handling at each level, leading to near-impossible unit-testing.

## Tests

See `LongRunningCloudClient` above, that was used extensively to test that everything works + many tests have been added to unit-test metrics. But for most actual problems (`CSHARP` vs `CSP` etc) it is virtually impossible to automate the tests.

## Documentation

The only public-facing change are:

* The default client name is now `hznet.client_N` and this name is documented in the API documentation, which is automatically updated from code comments, which have been updated accordingly.
* The pre-configuration thing has been documented in the configuration documentation.

## Release Notes

Our release notes are built automatically by listing issues on GitHub, so this will be included in the release notes.

src/Hazelcast.Net/Core/ClientVersion.cs
```diff
@@ -53,5 +57,33 @@ internal static string Version
...
+                var version = Version;
+                var pos0 = version.IndexOf('.');
+                var pos1 = version.IndexOf('.', pos0 + 1);
+                if (pos1 >= 0)
+                {
+                    version = version[..pos1];
+                }
+                else
+                {
+                    var pos = version.IndexOf('+');
```
ihsandemir, line 81: I don't get how this works with the above example semver version `5.1.2.preview-0+ab16ec43`? This has 3 dots. Do you have unit tests for this function?

src/Hazelcast.Net/Core/ClientVersion.cs
```diff
@@ -53,5 +57,33 @@ internal static string Version
...
+        /// <remarks>
+        /// <para>Returns the pure version, ie major.minor.patch.</para>
+        /// </remarks>
+        internal static string PureVersion
+        {
+            get
+            {
+                if (_clientVersionPure != null) return _clientVersionPure;
+                var version = Version;
+                var pos0 = version.IndexOf('.');
```
ihsandemir, line 73: What are all these github check actions warning messages ? Do we have a plan to solve them? 
```
Check warning on line 73 in src/Hazelcast.Net/Core/ClientVersion.cs

GitHub Actions
/ Build PR (windows-latest)
src/Hazelcast.Net/Core/ClientVersion.cs#L73
'string.IndexOf(char)' has a method overload that takes a 'StringComparison' parameter. Replace this call in 'Hazelcast.Core.ClientVersion.PureVersion.get' with a call to 'string.IndexOf(char, System.StringComparison)' for clarity of intent.
```

src/Hazelcast.Net/Metrics/MetricsCompressor.cs
```diff
@@ -161,9 +161,12 @@ private void AppendDescriptor(IMetricDescriptor descriptor)
             if (mask.HasNone(DescriptorMask.Unit))
                 _tempOutput.WriteByte((byte) descriptor.Unit);
 
-            //if (mask.HasNone(DescriptorMask.ExcludedTargets))
-            //    tmpDos.WriteByte(MetricTarget.BitSet(descriptor.ExcludedTargets));
+            // include excludeTargets for compatibility purposes (but it's always zero)
+            if (mask.HasNone(DescriptorMask.ExcludedTargets))
+                _tempOutput.WriteByte(0);
 
+            // include excludeTargets for compatibility purposes (but it's always zero)
```
ihsandemir, line 168: same comment two lines above?

src/Hazelcast.Net/Metrics/MetricsCompressor.cs
```diff
@@ -161,9 +161,12 @@ private void AppendDescriptor(IMetricDescriptor descriptor)
...
+            // (the Python client for instance always writes zero here)
```
ihsandemir, line 169: why do we refer to python client here? how is that relevant?

src/Hazelcast.Net/NearCaching/NearCacheStatistics.cs
```diff
@@ -146,24 +147,35 @@ public void NotifyStaleRead()
...
+                Evictions = CreateDescriptor<long>("evictions", MetricUnit.Count);
+                Hits = CreateDescriptor<long>("hits", MetricUnit.Count);
+                Misses = CreateDescriptor<long>("misses", MetricUnit.Count);
+                EntryCount = CreateDescriptor<long>("ownedEntryCount", MetricUnit.Count);
+                Expirations = CreateDescriptor<long>("expirations", MetricUnit.Count);
+                Invalidations = CreateDescriptor<long>("invalidations", MetricUnit.Count);
+                InvalidationRequests = CreateDescriptor<long>("invalidationRequests", MetricUnit.Count);
+                OwnedEntryMemoryCost = CreateDescriptor<long>("ownedEntryMemoryCost", MetricUnit.Bytes);
+
+                // these exist in Java but not in Python?
```
ihsandemir, line 174: Java is the reference, right?

src/Hazelcast.Net/NearCaching/NearCacheStatistics.cs
```diff
@@ -197,12 +209,13 @@ public IEnumerable<Metric> PublishMetrics()
             yield return _metricDescriptors.InvalidationRequests.WithoutValue();
             yield return _metricDescriptors.OwnedEntryMemoryCost.WithoutValue();
 
+            // TODO consider enabling these?
```
ihsandemir, line 212: Why did you disable them?

src/Hazelcast.Net/HazelcastOptions.ClusterOptions.cs
```diff
@@ -39,10 +39,10 @@ public partial class HazelcastOptions : IClusterOptions // ClusterName, Subscrib
         internal const string DefaultClusterName = "dev";
 
         /// <summary>
-        /// Gets the default client name prefix, which is <c>"hz.client_"</c>.
+        /// Gets the default client name prefix, which is <c>"hznet.client_"</c>.
```
ihsandemir, line 42: I don't get the benefits of doing things differently from our reference implementation, Java client. If there is such a benefit, we need to discuss as a group to decide whether we should not follow the reference.

src/Hazelcast.Net/HazelcastOptions.Build.cs
```diff
@@ -30,23 +30,25 @@ public partial class HazelcastOptions // Build
...
 
-            return Build(configuration, configure);
+            return Build(configuration, preConfigure, configure);
         }
 
         /// <summary>
         /// (internal for tests only) Builds Hazelcast options, using an alternate key.
         /// </summary>
         /// <param name="setup">An <see cref="IConfigurationBuilder"/> setup delegate.</param>
+        /// <param name="preConfigure">An <see cref="HazelcastOptions"/> pre-configuration delegate.</param>
```
ihsandemir, line 51:
```suggestion
        /// <param name="preConfigure">A <see cref="HazelcastOptions"/> pre-configuration delegate.</param>
```

src/Hazelcast.Net/HazelcastOptions.Build.cs
```diff
@@ -56,29 +58,32 @@ internal static HazelcastOptions Build(Action<IConfigurationBuilder> setup, Acti
...
-        private static HazelcastOptions Build(IConfiguration configuration, Action<IConfiguration, HazelcastOptions> configure, string altKey)
+        private static HazelcastOptions Build(IConfiguration configuration, Action<HazelcastOptions> preConfigure, Action<IConfiguration, HazelcastOptions> configure, string altKey)
         {
             // must HzBind here and not simply Bind because we use our custom
             // binder which handles more situations such as ignoring and/or
             // renaming properties
 
             var options = new HazelcastOptions();
+
+            preConfigure?.Invoke(options);
```
ihsandemir, line 85: `The HazelcastOptionsBuilder and HazelcastOptions.Build have been extended to support pre-options. This allows the user to programmatically set options that still can be overriden by, say, command-line arguments or environment variables.`
Does this mean that you are changing the behaviour of user api options in a patch release? We usually tend not to change it even for minor releases.

ihsandemir (commented):
Regarding you comment `But for most actual problems (CSHARP vs CSP etc) it is virtually impossible to automate the tests.`, can you not execute simple javascript code at server side to retrieve the client name? We do tests like this using RC.

src/Hazelcast.Net/Core/ClientVersion.cs
```diff
@@ -53,5 +57,33 @@ internal static string Version
...
+                if (pos1 >= 0)
+                {
+                    version = version[..pos1];
+                }
+                else
+                {
+                    var pos = version.IndexOf('+');
+                    if (pos >= 0) version = version[..pos];
+                }
+                _clientVersionPure = version;
```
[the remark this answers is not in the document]
zpqrtbnk: yes, it assigns a reference, and reference assignments are safe

src/Hazelcast.Net/Core/ClientVersion.cs
zpqrtbnk: say the original `IndexOf` method only accepted 1 arg, and then a new version of .NET introduces an overload that accepts 2 args, and they recommend you use 2 args, so now we have a warning - they are harmless but I would die to fix them. This is a TECH_DEBT task that I need to create in Jira. Will do.

src/Hazelcast.Net/Metrics/MetricsCompressor.cs
zpqrtbnk: Because this "excludeTargets" thing is generally undocumented and the only way to figure what to do it do spy other clients, so this explains *why* we are sending zero here: because other clients seem to do it and it seems to work.

src/Hazelcast.Net/NearCaching/NearCacheStatistics.cs
zpqrtbnk: In theory yes, but there are many places in Java where things have become obsolete... Like, I am not too sure what these metrics are, and not all clients send them. Again, the issue here is that there is no place I can look at to learn how these things work, and even though "java is the reference", in many places it would lead us to implement obsolete things in other clients just because Java does it...?

src/Hazelcast.Net/NearCaching/NearCacheStatistics.cs
zpqrtbnk: Because they cause undocumented issues.

src/Hazelcast.Net/HazelcastOptions.ClusterOptions.cs
zpqrtbnk: But then again, "hz.client_" is for Java, but other clients _already_ use a different name, so I aligned with them...

src/Hazelcast.Net/HazelcastOptions.Build.cs
zpqrtbnk: Somehow. I am extending the API so code that works in 5.0.0 cannot break due to 5.0.1 changes. OTOH new methods become available. Can we consider this OK? Else, will remove it 'til 5.1. Let me know.

src/Hazelcast.Net/Metrics/MetricsCompressor.cs
zpqrtbnk: oups. fixing.

src/Hazelcast.Net/Core/ClientVersion.cs
zpqrtbnk: This is not a valid semver version that we use, would be `5.1.2-preview.0+ab16ec43`. Was a mistake in comments. Fixing + pushing unit tests (that never existed) and cleanup.

src/Hazelcast.Net/HazelcastOptions.ClusterOptions.cs
zpqrtbnk: I am pretty sure I have seen another client use a different name on MC, but everyone seems to agree that every client use the `hz.client_<n>` pattern. I am pretty sure that any value could be used here, since this is only used to create a client name when the user does not specify one, but the user could really specify anything. Anyways. Reverting the change.

src/Hazelcast.Net/HazelcastOptions.Build.cs
ihsandemir: Yes, it can not be in 5.0.1. Can it be in 5.1? Let's discuss this. We normally do not do backward incompatible changes. The only way is to deprecate the old but keep the old.

zpqrtbnk: well, that is, I believe what I have described: I am adding a new method, but not removing any old method, nor changing the old behavior in any way. so 5.0.1 would be fully backward-compatible, wrt options, only it would provide an additional method.

zpqrtbnk: OK let me know, I am willing to remove that code, which is only used in an example, if that can help merging the PR.

zpqrtbnk: Code removed.

zpqrtbnk:
OK... now that we don't have 5.0.1 anymore... I have entirely rebased that PR on top of master, re-introduced the code that I just removed, and we just need approval.

ihsandemir (approved):

PR #549, opened 2021-12-01
Force UsePublicAddress option when CloudDiscovery is enabled [API-1024]

zpqrtbnk:
See linked issue for details.

### DIFF
--- a/src/Hazelcast.Net/HazelcastClientFactory.cs
+++ b/src/Hazelcast.Net/HazelcastClientFactory.cs
@@ -17,6 +17,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Hazelcast.Aggregation;
 using Hazelcast.Clustering;
+using Hazelcast.Configuration;
 using Hazelcast.Core;
 using Hazelcast.Partitioning.Strategies;
 using Hazelcast.Projection;
@@ -207,6 +208,25 @@ namespace Hazelcast
 
             // clone the options - we don't want any change to the original options to impact this client
             options = options.Clone();
+
+            if (options.Networking.Cloud.Enabled)
+            {
+                // if cloud is enabled we must force usage of public addresses
+                // otherwise, ConnectAddressResolver may decide to use internal addresses if SSL is enabled
+                // TODO: why?
+
+                if (options.Networking.UsePublicAddresses.HasValue)
+                {
+                    if (!options.Networking.UsePublicAddresses.Value)
+                        throw new ConfigurationException("If CloudDiscovery is enabled, then UsePublicAddress cannot be false.");
+                }
+                else
+                {
+                    var logger = options.LoggerFactory.Service.CreateLogger(typeof (HazelcastClientFactory));
+                    options.Networking.UsePublicAddresses = true;
+                    logger.LogDebug("Options: Networking.Cloud.Enabled is true => set Networking.UsePublicAddress to true.");
+                }
+            }
 
             // this ensures that the clock is correctly configured before anything else
             // happens - remember the clock is static - so we are doing it here - and


emreyigit:
I'm just trying to understand. What will happen cloud+ssl connection?

zpqrtbnk:
The problem is the collision between the Cloud Discovery mechanism and the other (esp. k8s) mechanisms. With Cloud, the members will only report internal addresses, but Cloud itself specifies public addresses during the discovery phase... and for some reason that I don't fully understand, with k8s, you need to use internal addresses if you are doing SSL. These are, somehow, the rules that are used with Java. The end result was that Cloud would specify public addresses but, because of SSL, we would force the usage of internal addresses, which cannot be reached = failure.

Eventually, I would *really* like to restructure this to stop following the fragmented approach Java has, and have only 1 discovery and resolution service.

emreyigit:
Yeah, it looks confusing. We say use public for cloud but if there is SSL, then use internal -k8s wants it- but again cloud forces you to use public. It looks like a contradiction. 

Restructuring the service would be beneficial and more clear.

emreyigit (approved):
No rejection, looks nice!

ihsandemir:
> The problem is the collision between the Cloud Discovery mechanism and the other (esp. k8s) mechanisms. With Cloud, the members will only report internal addresses, but Cloud itself specifies public addresses during the discovery phase... and for some reason that I don't fully understand, with k8s, you need to use internal addresses if you are doing SSL. These are, somehow, the rules that are used with Java. The end result was that Cloud would specify public addresses but, because of SSL, we would force the usage of internal addresses, which cannot be reached = failure.
> 
> Eventually, I would _really_ like to restructure this to stop following the fragmented approach Java has, and have only 1 discovery and resolution service.

@zpqrtbnk  Can you please point me to the Java code which has the behaviour you described? We need to justify this behaviour. What is the reason for this different behaviour even for Java client? Do you agree on above @sancar @mdumandag ?

zpqrtbnk:
We discussed it with @sancar this morning. In Java you have code like this:
```java
String publicIpEnabledProperty = properties.getString(ClientProperty.DISCOVERY_SPI_PUBLIC_IP_ENABLED);
if (publicIpEnabledProperty == null) {
    SSLConfig sslConfig = config.getSSLConfig();
    if (sslConfig != null && sslConfig.isEnabled()) {
        if (logger.isFineEnabled()) {
            logger.fine("SSL is configured. The client will use internal addresses to communicate with the cluster. If "
                           + "members are not reachable via private addresses, "
                           + "please set \"hazelcast.discovery.public.ip.enabled\" to true ");
        }
    }
    return false;
}
```
If this code runs, and SSL is enabled, and you have not forced usage of public addresses, it will select private addresses. Which is an error in the case of Cloud. Sancar asked:
> Are we talking about the behaviour that we force users to set `"hazelcast.discovery.public.ip.enabled"` explicitly when SSL is enabled ?

Well, yes, exactly. And then he wrote:
> I think I got the problem. I also found the reasoning of this check. I am assuming that we are advising setting `hazelcast.discovery.public.ip.enabled` for cloud and also setting that option by default in the config provided via the web UI. [...] This is the reason of the strange behaviour. https://github.com/hazelcast/hazelcast/pull/17895#discussion_r535907860

Waiting for him to provide more details. But for .NET I thought, since it just cannot work without the option being set, better force the option for Cloud rather than fail and then have to tell user and document that they must set the option.

zpqrtbnk:
I believe this PR has been closed by mistake when the `5.0.z` branch was destroyed. The issue remains unsolved. We should re-open this PR. Alas, it targets a branch that does not exist anymore = cannot be re-opened, and the target branch cannot be edited. Going to open a new PR :(

PR #555, opened 2021-12-07
Connection exceptions during first connection

emreyigit:
During first connection, accumulating exceptions were causing memory leakage. It is fixed. Also, parsing method of network address is improved to check empty spaces.


src/Hazelcast.Net/Clustering/ClusterConnections.cs
```diff
@@ -500,26 +501,25 @@ private async Task ConnectFirstAsync(CancellationToken cancellationToken)
...
-                                exceptions.Add(attempt.Exception);
-
-                                _logger.LogDebug(attempt.Exception, $"Failed to connect to address {address}.");
+                                isExceptionThrown = true;
+                                _logger.LogError(attempt.Exception, "Failed to connect to address {address}.", address.ToString());
                             }
                         }
                         else
                         {
+                            //is !attempt.HasException possible here?
```
zpqrtbnk, line 522: I am not entirely sure but better handle it, yes. And... if it cannot happen *today* it may happen in the future.

src/Hazelcast.Net/Clustering/ClusterConnections.cs
```diff
@@ -549,23 +546,24 @@ private async Task ConnectFirstAsync(CancellationToken cancellationToken)
                 }
                 catch (Exception e) // gather exceptions
```
zpqrtbnk, line 547: Well, we're not gathering anymore.

src/Hazelcast.Net/Clustering/ClusterConnections.cs
```diff
@@ -549,23 +546,24 @@ private async Task ConnectFirstAsync(CancellationToken cancellationToken)
...
+                    isExceptionThrown = true;
                 }
 
             } while (canRetry);
 
-            var aggregate = exceptions == null ? null : new AggregateException(exceptions);
+            string msgSomeThingWentWrong = "";
+
+            if (isExceptionThrown)
+                msgSomeThingWentWrong = "Exception(s) is thrown in the past. Please, see the logs.";
```
zpqrtbnk, line 559: "Some exceptions were thrown and have been written to the log. Please refer to the log for details."

src/Hazelcast.Net/Clustering/ClusterConnections.cs
```diff
@@ -549,23 +546,24 @@ private async Task ConnectFirstAsync(CancellationToken cancellationToken)
...
 
             // canceled exception?
             if (cancellationToken.IsCancellationRequested)
-                throw new OperationCanceledException($"The cluster connection operation to \"{_clusterState.ClusterName}\" has been canceled. " +
-                    $"The following addresses where tried: {string.Join(", ", tried)}.", aggregate);
```
zpqrtbnk, line 564: I think we want to keep logging the addresses that were tried. Can you keep the "The following addresses..." message? And verify that the `tried` thing is some sort of `HashSet<>` that will not grow infinitely?

zpqrtbnk (changes requested):
Have commented.

src/Hazelcast.Net/Clustering/ClusterConnections.cs
emreyigit: I undo that part. Tried addresses are already kept in set.

src/Hazelcast.Net/Clustering/ClusterConnections.cs
emreyigit: Yeah :D

zpqrtbnk (approved):
Happy now.

src/Hazelcast.Net/Clustering/ClusterConnections.cs
```diff
@@ -547,25 +543,30 @@ private async Task ConnectFirstAsync(CancellationToken cancellationToken)
...
             // canceled exception?
             if (cancellationToken.IsCancellationRequested)
                 throw new OperationCanceledException($"The cluster connection operation to \"{_clusterState.ClusterName}\" has been canceled. " +
-                    $"The following addresses where tried: {string.Join(", ", tried)}.", aggregate);
+                    $"The following addresses where tried: {string.Join(", ", tried)}." +
+                    $" {msgSomeThingWentWrong}");
 
             // other exception
             throw new ConnectionException($"Unable to connect to the cluster \"{_clusterState.ClusterName}\". " +
-                $"The following addresses where tried: {string.Join(", ", tried)}.", aggregate);
```
ihsandemir, line 568: .NET uses `ConnectionException` when could not connect and Java uses `IllegalStateException`. This is confusing. Why is this difference? (it may not related to this PR)

src/Hazelcast.Net.Tests/Networking/NetworkAddressTests.cs
```diff
@@ -29,6 +29,9 @@ namespace Hazelcast.Tests.Networking
     public class NetworkAddressTests
     {
         [TestCase("127.0.0.1", true, "127.0.0.1:0")]
+        [TestCase(" 127.0.0.1", true, "127.0.0.1:0")]
```
ihsandemir, line 32: Does port 0 means we will use port 5701 for connection? Asking this because that is our default port if none is provided.

emreyigit: This is just to parse network address correctly independently from config. Server ports come up during connection establishment. 
For reference : https://github.com/hazelcast/hazelcast-csharp-client/blob/bbae59bea5cc254d7779affaa0dfb8e50a8704dd/src/Hazelcast.Net/Networking/AddressProvider.cs#L186

src/Hazelcast.Net/Clustering/ClusterConnections.cs
```diff
@@ -500,22 +501,20 @@ private async Task ConnectFirstAsync(CancellationToken cancellationToken)
 
                         HConsole.WriteLine(this, $"Failed to connect to address {address}");
 
-                        if (attempt.HasException) // else gather exceptions
+                        if (attempt.HasException)
                         {
                             if (attempt.Exception is RemoteException { Error: RemoteError.HazelcastInstanceNotActive })
                             {
-                                _logger.LogDebug($"Failed to connect to address {address} (member is not active).");
+                                _logger.LogWarning($"Failed to connect to address {address} (member is not active).");
```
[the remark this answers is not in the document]
emreyigit: I think @zpqrtbnk can give more details.

src/Hazelcast.Net/Clustering/ClusterConnections.cs
emreyigit: Isn't `ConnectionException`  more relevant to the context? WDYT @zpqrtbnk?

zpqrtbnk: Yes, .NET uses different names for a range of exceptions, so if you compare line-to-line to Java, you're going to be confused. The `ConnectionException` name was choosen because it was considered more meaningful than the Java couterpart (illegal state?).

This is not related to this PR, so should not block the PR. OTOH we need to have one discussion re. the differences between the .NET Client and Java Client. They *are* definitively different, the more you look into it the more you'll be confused. And we need to decide whether this is acceptable, or whether we want to start aligning things.

ihsandemir: Yes, it should not stop the PR. My take on the matter is that unless there is a good reason and very obvious benefit to the user, we try keeping things the same so that the users will be more familiar with names, etc. and better documentation for all the clients. Let's talk about this in the group.

src/Hazelcast.Net/Clustering/ClusterConnections.cs
```diff
@@ -529,7 +529,7 @@ private async Task ConnectFirstAsync(CancellationToken cancellationToken)
                     // address is invalid or cannot be resolved via DNS... a DNS problem may
                     // be transient: better retry
                     isExceptionThrown = true;
-                    _logger.LogError(e, "Connection attempt has thrown.");
+                    _logger.LogError(e, "Connection attempt failed due to possible DNS error.");
```
ihsandemir, line 532: The actual exception reason will be printed via `e`, and we can not be 100% sure 't 's DNS problem, right? Then, it may be DNS or something else, hence I would rather say it may be DNS or some other problem,.

ihsandemir (approved):

src/Hazelcast.Net/Clustering/ClusterConnections.cs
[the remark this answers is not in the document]
zpqrtbnk: `HConsole` is an internal debugging tool / console that is not compiled into the `Release` build, so it does not ship to users, whereas everything "logger" is. In some cases yes, messages may be duplicated. Usage is different.

PR #557, opened 2021-12-09
Adjustments to Jira workflow

zpqrtbnk:
Proposing changes for the Jira workflow:
- Rename the label workflow file to `jira-label.yml`
- Get it to trigger on the `Jira` label too, so we can rename the label
- Does not trigger if the issue title already contains `[API-`

That last one condition is added so that we can add the `Jira` label to an issue or PR that we have already manually linked to Jira, and prevent the workflow from triggering. Or, should we remove the label by accident, we can add it back without the workflow triggering. Yes, happened to me, and I ended up with duplicates in Jira.

Note: if/when the PR is merged, the `to-jira` label can (should) be renamed to `Jira`.

### DIFF
--- a/.github/workflows/jira-label.yml
+++ b/.github/workflows/jira-label.yml
@@ -1,7 +1,7 @@
 name: Create Jira issue
 
 # runs: whenever a label is added to an issue
-# if: the issue title does not yet contain '[API-' and
+# if: the issue title does not yet contain 'API-' and
 #     the label is 'to-jira' or 'Jira'
 # does: create corresponding Jira issue and link the two issues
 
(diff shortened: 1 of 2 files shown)


.github/workflows/jira-label.yml
```diff
@@ -1,13 +1,23 @@
...
 on:
   issues:
     types: labeled
 
 jobs:
   jira:
-    if: ${{ github.event.label.name == 'to-jira' }}
     runs-on: ubuntu-latest
+    if: |
+      !contains(github.event.issue.title, '[API-') &&
```
degerhz, line 16: Brackets are actually not necessary. So the title might include (if added manually) Jira issue key without brackets. Maybe we can consider removing the starting bracket.

degerhz:
LGTM. Only minor comment. Thanks @zpqrtbnk

.github/workflows/jira-label.yml
zpqrtbnk: Makes sense, removed the test for the bracket.

degerhz (approved):
Thanks for the improvements @zpqrtbnk

PR #566, opened 2022-01-13
hz.ps1 / Add local-lib and libpath options

zpqrtbnk:
Problem:

In its current version, the `hz.ps1` script insists on validating the server version on Maven (for instance, if it is 5.0, we may actually use 5.0.4 if that is the latest available on Maven), and then downloading JAR files and a default XML configuration file, from Maven.

This is a problem when trying to test the client against server versions that are not yet published on Maven, such as the recent `5.1-BETA-1`.

Proposal:

This PR adds the two following options to the `hz.ps1` script:

* `-libpath` specifies the libs directory, i.e. where to store and find JAR files and the default XML server configuration file. By default, this is `./temp/lib` but it can now be set to point anywhere really.
* `-local-lib` specifies that the libs directory is expected to contain whatever is needed, an no attempt should be made at validating the server version from Maven, nor downloading JAR files, nor downloading XML files.

With these options, it becomes possible to run tests against a completely weird server version that is not available on Maven:
```
./hz.ps1 test -server-version 5.1.0-MEH-44 -local-lib
```

This of course assumes that the expected JARs and XML files are present in `./temp/lib` or at the location indicated by the `-libpath` option. The files we need are:
* `hazelcast-<version>.jar`
* `hazelcast-sql-<version>.jar`
* `hazelcast-<version>-tests.jar`
* `hazelcast-<version>.xml` (default configuration file)
* `hazelcast-remote-controller-<rc-version>.jar`

In addition, when enterprise tests are run, i.e.
```
./hz.ps1 test -enterprise -server-version 5.1.0-MEH-44 -local-lib
```

The two additional files are required:
* `hazelcast-enterprise-<version>.jar`
* `hazelcast-enterprise-<version>-tests.jar`

### DIFF
--- a/hz.ps1
+++ b/hz.ps1
@@ -125,10 +125,6 @@ $params = @(
     @{ name = "verboseTests";   type = [switch];  default = $false;      alias = "verbose-tests";
        desc = "Verbose tests results with errors"
     },
-    @{ name = "localLib" ;      type=[switch];    default=$false;         alias = "local-lib";
-       desc = "Whether to not download JAR files";
-       note = "When this option is set, no attempt is made to verify the server version, nor to download any JAR or configuration files."
-    },
     @{ name = "yolo";            type = [switch]; default = $false;
        desc = "Confirms excution of sensitive actions"
     }
@@ -161,7 +157,7 @@ $actions = @(
     },
     @{ name = "test";
        desc = "runs the tests";
-       need = @( "git", "dotnet-complete", "java", "server-files", "server-version", "build-proj", "enterprise-key" )
+       need = @( "git", "dotnet-complete", "java", "server-files", "build-proj", "enterprise-key" )
     },
     @{ name = "build-docs";
        desc = "builds the documentation";
@@ -184,19 +180,19 @@ $actions = @(
        uniq = $true;
        desc = "runs the remote controller for tests";
        note = "This command downloads the required JARs and configuration file.";
-       need = @( "java", "server-files", "server-version", "enterprise-key" )
+       need = @( "java", "server-files", "enterprise-key" )
     },
     @{ name = "run-server";
        uniq = $true;
        desc = "runs a server for tests";
        note = "This command downloads the required JARs and configuration file.";
-       need = @( "java", "server-files", "server-version", "enterprise-key" )
+       need = @( "java", "server-files", "enterprise-key" )
     },
     @{ name = "get-server";
        uniq = $true;
        desc = "gets a server for tests";
        note = "This command downloads the required JARs and configuration file.";
-       need = @( "java", "server-files", "server-version", "enterprise-key" )
+       need = @( "java", "server-files", "enterprise-key" )
     },
     @{ name = "generate-codecs";
        uniq = $true;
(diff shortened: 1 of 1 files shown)


alisengul53:
Do we really need an extra parameter to handle this issue?

AFAIK script checks the maven first and if the version is there and the checking /tmp/lib folder. If the required jars are in this path, it doesn't dowload the jars and using already existing ones. For example, we were running compatibilites and the version was on maven but we didn't download any jars from there because we already built the jars and put on `/tmp/lib/`. Script detected the jars

I think we can change the order. For example, before checking the version on maven, we can check if it is available on the /tem/lib folder, if the required jars are there, no need to check version on maven. And it will download remote controller and xml itself. Then ready to go

That is just my understanding

emreyigit:
As far as I see, script expects all required arguments should be satisfied. Checking server version is one of them. Work around on this could be tricky without adding an argument.

zpqrtbnk:
Interesting. We check for Maven because maybe the version that is specified (`5.0`) is not the one we should use, but instead we should use the latest `5.0.12`. But... if you specify `5.1.0-MEH-42` and we indeed have all required files there, then the idea would be to skip the check entirely? So normal usage is specify `5.0` and do the Maven magic, but if you put files in the right place + specify a version, then nothing is needed. Right?

So I understand you don't need `-libpath` because it's OK for you to copy the files in `./temp/lib`? Correct?

I'll look into how I can flip the order of things in the script so that no additional option is required. Stay tuned.

alisengul53:
>So normal usage is specify 5.0 and do the Maven magic, but if you put files in the right place + specify a version, then nothing is needed. Right?

Yes, we are specifying the exact version when we trigger workflows.

>So I understand you don't need -libpath because it's OK for you to copy the files in ./temp/lib? Correct?

Yes because our workflows are already putting built jars to the folder `tmp/lib/`

zpqrtbnk:
I have pushed an update and here is what we have now, assuming that `./temp/lib` is empty.

```text
# Detects that we have nothing
# Pings Maven, determine that the latest 5.0 is 5.0.3
# Download 5.0.3-SNAPSHOT file
./hz.ps1 get-server -server-version 5.0-SNAPSHOT

# Detects that we do NOT have 5.0-SNAPSHOT (since we have 5.0.3-SNAPSHOT)
# Pings Maven, determine that the latest 5.0 is 5.0.3
# Detects that we have all files for 5.0.3
./hz.ps1 get-server -server-version 5.0-SNAPSHOT

# Detects that we have all files for 5.0.3
# Does not ping Maven, downloads nothing
./hz.ps1 get-server -server-version 5.0.3-SNAPSHOT
```

That being said, notice that the "pings Maven, determine the version" only triggers for `-SNAPSHOT` versions, so if you specify `5.1-BETA-4` then it will *not* trigger, so I believe that you could essentially use the script as it is today, just drop the appropriate files in `./temp/lib` beforehand (the JARs including the RC, and the XML config file).

Which means that *maybe* we don't even need this PR at all unless we want to be able to build on a laptop which is not connected to the network. Happened to me couple of times on a train. So I'd say we merge it anyways.

alisengul53:
So we need to put rc and xml to the folder in this approach right?

In the current approach if we are able to skip maven check, it is also downloading rc and xml itself. An example;

`Server: found version 5.0.1-SNAPSHOT on Maven, using this version
Prepare server/rc...
Downloading hazelcast-remote-controller-0.8-SNAPSHOT.jar ...
Detected hazelcast-5.0.1-SNAPSHOT-tests.jar
Detected hazelcast-5.0.1-SNAPSHOT.jar
Detected hazelcast-sql-5.0.1-SNAPSHOT.jar
Downloading hazelcast-default.xml -> hazelcast-5.0.1-SNAPSHOT.xml...`

But we can merge it like that because we are also planning to unify these scripts for backward compatibility tests, maybe it is unnecessary to work on this so much.

zpqrtbnk:
yes, in this approach you'd need to put rc, server and xml in the temp/lib folder beforehand - and then it would skip the maven check - ok I am going to kill this PR as I am preparing another one with various improvements to the script as we discussed on Slack.

PR #569, opened 2022-01-19
Fix SQL codecs generation

zpqrtbnk:
This PR fixes the SQL codecs generation. By mistake, we let a PR go through with some hand-writen codecs. This PR re-sync the `protocol` submodule and makes sure that the SQL codecs are generated.

Note that the commit we use for `protocol` is going to generate extra codecs (`MapReplaceAllCodec`, `SqlMappingDdlCodec`, and `HazelcastJsonValueCodec`) and these files should then be manually removed, because we cannot do miracles. It will all fix itself are we merge all our pending PRs.


src/Hazelcast.Net/Protocol/CustomCodecs/SqlErrorCodec.cs
```diff
@@ -66,16 +67,16 @@ public static Hazelcast.Sql.SqlError Decode(IEnumerator<Frame> iterator)
 
             var originatingMemberId = initialFrame.Bytes.ReadGuidL(OriginatingMemberIdFieldOffset);
             var message = CodecUtil.DecodeNullable(iterator, StringCodec.Decode);
-            string suggestion = String.Empty;
-
-            if (!iterator.AtStructEnd())
+            var isSuggestionExists = false;
+            string suggestion = default;
+            if (iterator.NextIsNotTheEnd())
```
emreyigit, line 72: I think this shifts the frames more than needed. `StringCodec.Decode` uses `Frame.Take` which takes current frame and shifts to next. When combined with  `NextIsNotTheEnd+Decode` makes two shift. OTOH, `Frame.AtStructEnd` just peeks to current frame.

emreyigit (approved):
Looks nice👌

src/Hazelcast.Net/Protocol/Codecs/SqlExecuteCodec.cs
```diff
@@ -97,10 +98,21 @@ public sealed class RequestParameters
...
+            /// <summary>
+            /// Flag to skip updating phone home statistics.
+            ///</summary>
+            public bool SkipUpdateStatistics { get; set; }
+
+            /// <summary>
+            /// <c>true</c> if the skipUpdateStatistics is received from the client, <c>false</c> otherwise.
+            /// If this is false, skipUpdateStatistics has the default value for its type.
+            /// </summary>
+            public bool IsSkipUpdateStatisticsExists { get; set; }
```
alisengul53, line 111: Do we really need two parameters here? Don't we have a chance to handle it with only one veriable? AFAIU, they are depended.
Or server side expecting both of them?

zpqrtbnk: No, these 2 parameters don't really make sense, but that is the code that the generator generates. And we don't want to think, we want to be able to run the generator and just accept whatever it generates. So we'd have to fix the generator, and that has impact on *all* languages (the Java code is similar to this too). For the time being... we will have to do with this.

If we fix this code in our repo, next time we generate codecs, these changes would be overriden.

src/Hazelcast.Net/Protocol/CustomCodecs/SqlErrorCodec.cs
zpqrtbnk: `NextIsNotTheEnd` had a bug, weird enough, that I fixed, and now this passes all tests. Or can you think of a test that we could get to fail?

alisengul53:
Looks good to me

alisengul53 (approved):

src/Hazelcast.Net/Protocol/CustomCodecs/SqlErrorCodec.cs
emreyigit: I think that is enough as long as that is not shift the position.

PR #570, opened 2022-01-19
Improve the `hz` script

zpqrtbnk:
This PR improves the `hz` script in various ways.  
Note that each improvement has its own commit.

## Target Frameworks

The .NET **target frameworks** are now determined by inspecting the `Hazelcast.Net.Tests.csproj` file. Before, they were hard-coded in the script. This means that the script will not need to be changed when we add support for .NET 6. This also means that the script may be used on old versions of the code (4.x), though this remains to be tested.

## Remote Controller

We had a `run-remote-controller` command which would start the remote controller, wait for a key to be pressed, and then stop the remote controller. The Remote Controller **can now be started and stopped independently** with the `start-remote-controller` and `stop-remote-controller` commands.

>NOTE: the pid of the running remote controller is stored in `temp/rc/pid`

By default, the `test` command would always start a remote controller, and stop it after the tests execution. The command now **detects a running remote controller and uses it** based upon whether a `temp/rc/pid` file exists. This means that one can start a remote controller with `start-remote-controller` and run tests with `test` multiple times without starting a remote controller each time. Of course, one has to `stop-remote-controller` eventually.

## Java Version

The script now **detects the Java version and fails** if the version is greater than 11, as our tests require the scripting feature of Java, which is supported up until Java 11. This is a fail-fast feature: instead of seeing weird failures in tests, the script does not even run the tests.

## Shell AutoComplete

The script now **supports shell auto-complete**. It must be initialized by running `hz completion-initialize` and then, tab-completion becomes available. For instance, typing `./hz.ps1 remote` and pressing tab cycles through all commands that include "remote" in their name.

## Server Version

As a remplacement for #566 the script now only hit Maven if the version is a snapshot version, and a `major.minor`, for instance `5.0-SNAPSHOT`. If it is `5.0.3-SNAPSHOT` it assumes the version number is correct and does not hit Maven. And then, it downloads the JAR and XML files to `temp/lib` *only* if they are not there already. Which means that if the files are there, and the version is complete, *nothing* will be downloaded.

In addition, the `-server-version` option now accepts `master` as a value, in which case it determines the version from GitHub and downloads snapshot files for that version.


alisengul53:
Even if we started the remote controller and pid file is in `temp/rc/` and then `.\hz.ps1 test` it still download server 5.0.3. As server default value is 5.0-SNAPSHOP. It cannot find it on maven and downloading 5.0.3 version. After it downloads I gues it is checking the `pid` file. We can totally skip downloading jars process if pid file is there I think

zpqrtbnk:
First point: you are right, if the rc is already running we can probably assume that it's the correct server version + we have all the files. So we'd detect PID first and then skip even more checks. Looking into it.

Second point: if we don't check the PID how can we figure out that the RC is not running? It would then forces us to always take care of starting the RC manually?

alisengul53:
>Second point: if we don't check the PID how can we figure out that the RC is not running? It would then forces us to always take care of starting the RC manually?

You are right. Current behaviour of `test` command  is downloading everything and running the tests. It needs to know if rc is running. I just thought that if we use a common script for rc with all clients, a command which assumes rc is already running and runs the tests would be helpful for us. But it is not so important, we can also handle it on the GHA workflow.

zpqrtbnk:
> It needs to know if RC is running.

Yes ;) Either by using the PID file, or I could add a `-rc-is-running` option of some sort... Can we conclude on this PR and approve or reject or discuss? There are some things that can be useful there :)

alisengul53:
The PR solves the general problem, we can have more additions to improve it in the future. For now it is good to go. Thanks for the work on it @zpqrtbnk

alisengul53 (approved):

PR #573, opened 2022-02-01
Remove dependency on ExpectedObjects for tests

zpqrtbnk:
Up until now we used a dependency on an external project in our tests, that had to be built in our code because it was not distributed on NuGet as a signed/unsigned assembly. It was embedded as a Git submodule. All in all, it made things more complex, and we barely used its features. This PR removes the dependency and achieves the same results differently.

This is only part of our tests and has no impact on production code.


emreyigit (approved):
LGTM

alisengul53 (approved):
Thanks for the PR, it is good to remove an external dependency

Issue #550, opened 2021-12-02
Run SSL tests without relying on the JAR certificates [API-1029]


zpqrtbnk:
Currently, the .NET Client SSL tests rely on SSL certificates which are provided by the `hazelcast-enterprise-<version>-tests.jar` JAR file. The core team would like to be able to stop producing and distributing this JAR file. In addition, certificates in this file expire after a while. This forces us to run our tests against `-SNAPSHOT` versions of the server to make use of the latest, valid certificates.

Therefore: we want to run our SSL tests without depending on the `hazelcast-enterprise-<version>-tests.jar` JAR file.

Issue #304, opened 2020-09-18
Some SSL tests randomly fail [API-1246]


zpqrtbnk:
What is wrong with `ClientSslMutualAuthTests`?

emreyigit:
#574 solves the issue.

PR #574, opened 2022-02-01
Stop relying on enterprise test JAR for testing SSL[API-1246][API-1029]

zpqrtbnk:
We want to run our SSL tests without depending on the `hazelcast-enterprise-<version>-tests.jar` JAR file. See #550 for reasons.

> *Before reviewing and merging this PR, or reviewing and merging #583 instead, decisions must be made (see below). I am going to write a TDD for this.*

This PR:
* Updates the resources management (`Resources.resx`, `TestFiles.cs`, etc.) so we stop embedding certificates in the .NET test excutable. They are available on-disk anyways, and that simplifies resources management. It also lets us use files generated in a `temp` directory.
* Updates the SSL test classes to use these on-disk files, and to have cleaner code, better names, etc. Re-enables some SSL tests that were ignored because they used to be unstable. Enables testing of chain & name validation & mutual auth in a more systematic way.
* Provides scripting (via `hz.ps1`) to generate the required, fresh certificates before running tests, if needed (if they are not already available on-disk and we want to run enterprise tests).

The end result is that we have more consistent SSL tests, that do not depend on the `hazelcast-enterprise-<version>-tests.jar` file anymore, but rely on generated certificates. We have scripted the certificates generation.

**Note 1**: because our PR builder cannot have access to the enteprise key, it does *not* run the SSL tests, which means that this PR checks can be green without any SSL tests running. 

**Note 2**: the code supports two modes. In MIXED mode, we keep using the original certificate file from Java that is signed by a trusted CA. This is what this PR does. In CUSTOM mode, we *also* replace that certificate, but that implies that we register our own temporary root CA on the test machine. This is what PR #583 does. That other PR checks are green too, but see note 1.

**Note 3**: I have created a `temp/ssl-tests` branch directly on the Hazelcast repository, running with CUSTOM mode. The branch builder *does* build it and runs the tests successfully. So this validates that the CUSTOM mode can be used on GitHub and we can run on fully generated certificates.

**Note 4**: these PRs should not modify the production code in any way. Two files from the `Hazelcast.Net` project are touched, mostly to document things or handle errors. The actual behavior of the client is unchanged.


emreyigit:
I can verify that generating and signing SSL certificates in test environment is working. It looks legit 💯. As you said, ssl part is not tested during `build-pr` but I think that is okay. We can rely on `build-branch` since there is not much traffic on the repo. We are doing the things mostly on the repo. Even something wrong after merge, we can handle it. 

One thing, after this PR is merged, it's better to do some cleaning on the repo, resource files etc.

Since this PR solves the ssl test problems, I'll close the #304 .

srknzl:
I think this PR's API-1264 link is wrong 🤔

emreyigit (dismissed):

src/Hazelcast.Net/Networking/SslLayer.cs
```diff
@@ -162,7 +167,13 @@ internal bool ValidateCertificate(object sender, X509Certificate cert, X509Chain
             {
                 if (_options.ValidateCertificateName)
                 {
-                    _logger.IfWarning()?.LogWarning("SSL certificate error: {PolicyErrors}.", policyErrors);
+                    var name = "";
+                    try
+                    {
+                        name = $" (cert name: '{cert.Subject}')";
```
emreyigit, line 173: Wondering, what kind of exception do you expect here?

emreyigit (approved):
LGTM.

alisengul53:
Approved

alisengul53 (approved):

PR #578, opened 2022-02-02
Clarify a test following SQL codecs generation fixes

zpqrtbnk:
This is a very minor PR that only affects 1 test and clarifies what it tests.

### DIFF
--- a/src/Hazelcast.Net.Tests/Messaging/ClientMessageTests.cs
+++ b/src/Hazelcast.Net.Tests/Messaging/ClientMessageTests.cs
@@ -152,15 +152,20 @@ namespace Hazelcast.Tests.Messaging
                 Assert.That(!e.NextIsNotTheEnd());
 
                 e.MoveNext();
+                Assert.That(e.Current, Is.Not.Null);
                 while (!e.Current.IsBeginStruct) e.MoveNext();
-                // at 2
-                Assert.That(e.NextIsNotTheEnd());
+                // at 2, is BeginStruct
+                Assert.That(!e.AtStructEnd()); // whether e.Current is EndStruct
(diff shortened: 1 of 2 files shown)


src/Hazelcast.Net.Tests/Messaging/ClientMessageTests.cs
```diff
@@ -152,15 +152,20 @@ public void Enumerator()
                 Assert.That(!e.NextIsNotTheEnd());
 
                 e.MoveNext();
+                Assert.That(e.Current, Is.Not.Null);
                 while (!e.Current.IsBeginStruct) e.MoveNext();
-                // at 2
-                Assert.That(e.NextIsNotTheEnd());
+                // at 2, is BeginStruct
+                Assert.That(!e.AtStructEnd()); // whether e.Current is EndStruct
+                Assert.That(e.NextIsNotTheEnd()); // whether e.Current is !EndStruct
```
emreyigit, line 159: I think comment should be like that;

```suggestion
                Assert.That(e.NextIsNotTheEnd()); // whether e.Current.Next is !EndStruct
```
Because `e.NextIsNotTheEnd()` is just peaking the next one, right?

emreyigit (approved):
Just one small comment.

src/Hazelcast.Net.Tests/Messaging/ClientMessageTests.cs
zpqrtbnk: Yes. But what's tricky is that it's looking that the next one that `e.Take()` would return, and that is the `e.Current` one. It's wacky but we cannot change that enumerator as it must mimick the Java enumerator for the generated codecs to work. This is why we have this test: to help understand the strange enumerator ;)

PR #581, opened 2022-02-05
Fix authentication options binding

zpqrtbnk:
See related issue.


src/Hazelcast.Net.Tests/Configuration/HazelcastOptionsTests.cs
```diff
@@ -235,6 +235,105 @@ public void AuthenticationOptionsFile()
...
 
+        [Test]
+        public void AuthenticationUsernamePassword()
+        {
+            const string json = @"{ ""hazelcast"": {
+""authentication"" : {
+    ""username-password"": { ""username"": ""bob"", ""password"": ""secret"" }
+}
+}}";
+            var stream = new MemoryStream(Encoding.UTF8.GetBytes(json));
```
emreyigit, line 246: Could you consider to use `ReadResource(json)` for name of not to repeat parsing part? We have already a method to do that just above.

emreyigit:
Just left one small comment. Other than that, looks great.

emreyigit (dismissed):

alisengul53 (commented):
Do we have any integration test which is using a real server(maybe with RC) for these authentication types?
Or maybe we also don't have for other clients. I have no idea

src/Hazelcast.Net.Tests/Configuration/HazelcastOptionsTests.cs
zpqrtbnk: Ah sure, have pushed a commit.

zpqrtbnk:
Ali: we don't have such tests as far as I am aware. OTOH, these tests here only validate that we parse the options correctly.

alisengul53 (dismissed):
LGTM

emreyigit (approved):

alisengul53 (approved):

PR #583, opened 2022-02-09
Stop relying on enterprise test JAR for testing SSL (full-custom)[API-1246]

zpqrtbnk:
This is an extension of #574 with full-custom certificates to test whether it would work on GitHub actions, as it require installing a root CA on both Windows and Linux machines... See notes & discussion on #574.


srknzl:
I think this PR's API-1264 link is wrong 🤔

zpqrtbnk:
that one will not get merged, #574 will be merged instead.

PR #585, opened 2022-02-10
Refactor test conditions, stop running SQL tests on 4.0 server

zpqrtbnk:
We realized that our SQL tests run against 4.0 servers, which makes no sense, as SQL is not supported there (only beta) and jet is n/a. Thus, it breaks our QA test suites. We *do* have a condition mechanism to filter tests per server version, but up until now it was relying on an environment variable set by `hz.ps1`, not the actual version of the running server.

With this PR,
* We fetch the server version from the server. Therefore, we are sure it is correct
* We remove the environment variable support as it does not make sense anymore
* If the RC is already running, we validate that the version matches the expected version
* We enforce the `[5.0,)` condition on all SQL tests = run them on 5.0 and above


emreyigit (approved):

alisengul53:
>We realized that our SQL tests run against 4.0 servers, which makes no sense, as SQL is not supported there (only beta)

Minor: SQL comes with 4.1

alisengul53:
The failures for Build PR, are they expected?

alisengul53:
When I try it locally, it seems it is still sending 5.x server config to rc and trying to create 4.x server with this config. So tests are fails immediately instead of skipping because one time set up is failed.
Or did I miss something?

zpqrtbnk:
Failures are *not* expected, investigating.

src/Hazelcast.Net.Tests/Testing/NUnitPropertyAttributeTests.cs
```diff
@@ -0,0 +1,83 @@
...
+            Assert.IsNull(TestContext.CurrentContext.Test.Properties["FIXTURE"]?.FirstOrDefault());
+
+            // test method properties are immediately available
+            Assert.AreEqual("method", TestContext.CurrentContext.Test.Properties["METHOD"]?.FirstOrDefault());
+        }
+
+        [TestCase(0)]
+        [TestCase(1)]
+        [Property("METHOD", "method")]
+        public void CanSetAndGetTestSuiteProperties(int ignored)
```
alisengul53, line 54: Do we have a purpose of this parameter?

src/Hazelcast.Net.Tests/Testing/NUnitTestActionAttributeTests.cs
```diff
@@ -0,0 +1,90 @@
...
+            }
+
+            Assert.That(TestContext.CurrentContext.Test.Properties["FIXTURE"]?.FirstOrDefault(), Is.EqualTo("fixture"));
+            Assert.That(TestContext.CurrentContext.Test.Properties["TEST"]?.FirstOrDefault(), Is.EqualTo("test"));
+        }
+
+        [TestCase(0)]
+        [TestCase(1)]
+        [Some("TEST", "test")]
+        public void TestSuite(int ignored)
```
alisengul53, line 49: Same question here

zpqrtbnk: Same answer here

zpqrtbnk:
> The failures for Build PR, are they expected?

Not happening anymore, were probably temporary glitches.

zpqrtbnk:
> When I try it locally, it seems it is still sending 5.x server config to rc and trying to create 4.x server with this config. So tests are fails immediately instead of skipping because one time set up is failed.

Interesting. With this PR we should *not* run v5 tests against v4 servers. However... At the moment we still use one unique server configuration file for both. I need to discuss it with you to understand the situation better.

alisengul53:
> > When I try it locally, it seems it is still sending 5.x server config to rc and trying to create 4.x server with this config. So tests are fails immediately instead of skipping because one time set up is failed.
> 
> Interesting. With this PR we should _not_ run v5 tests against v4 servers. However... At the moment we still use one unique server configuration file for both. I need to discuss it with you to understand the situation better.

Most probably It was my mistake, a later tried worked expected.

zpqrtbnk:
OK, there *were* issues when running against v4 server due to `FencedLock` tests, I have fixed these tests and now one can run the whole tests against v4 or v5 and it works (e.g. `hz test -server-version 4.0`). Plus, we have `FencedLock` tests against 3 members, which is better.

alisengul53:
Works well now.

alisengul53 (approved):

emreyigit (approved):
👌

PR #590, opened 2022-02-22
Fenced Lock

emreyigit:
### Fenced Lock
This PR represents Fenced Lock implemenattion on .NET clinet. [About Fenced Lock](https://docs.hazelcast.com/hazelcast/5.0/data-structures/fencedlock)

Highlights:

- Implementation follows the Java client. 
- Exceptions received from server as thrown as it is -> `RemoteException`
- There may behavioral differences between Java client due to platform. Such as async context flow instead thread. This is thing that user should be aware of it. Since a `Task` can work on different threads with same context, thread based session locks are done with async context in .NET client. However, this may not result with same behavioral as Java because a context can be used another `Task`s at the parallel. In this case, different tasks with same context can use the same session(remember sessions bound to context not to thread as in Java) on the CP Subsystem, and this may lead to a race condition.

>***NOTE***
>*The codecs used in this .NET PR are generated from the protocol repo as per [this PR](https://github.com/hazelcast/hazelcast-client-protocol/pull/419) which does _not_ exactly generate the correct code, as it would update all copyright year + generate some extra codecs - so if you try to rebuild the codecs it won't exactly match - this is accepted for now, we have some out-of-sync issues with the protocol repo that will be resolved eventually - not now - Stephan*


src/Hazelcast.Net/CP/CpSubSystemSession.Heartbeat.cs
```diff
@@ -0,0 +1,102 @@
...
+                return RunAsync(p.Item1, p.Item2, cancellationToken);
+
+            }, cancellationToken).CfAwait();
+
+        }
+
+        /// <summary>
+        /// Runs for a given session
+        /// </summary>
+        /// <param name="cancellationToken"></param>
```
zpqrtbnk, line 86: comments mismatch

src/Hazelcast.Net/CP/CpSubSystemSession.Requests.cs
```diff
@@ -0,0 +1,82 @@
...
+            var responseMessage = await Cluster.Messaging.SendAsync(requestMessage).CfAwait();
+            var response = CPSessionGenerateThreadIdCodec.DecodeResponse(responseMessage);
+            return response.Response;
+        }
+
+        /// <summary>
+        /// Creates a new session on the server
+        /// </summary>
+        /// <param name="groupId"></param>
+        /// <returns>CPSubsystemSessionState and hearbeat milliseconds</returns>
```
zpqrtbnk, line 46: Can you check the `heartbeat` spelling here and in many other places?

src/Hazelcast.Net/CP/CpSubSystemSession.Requests.cs
```diff
@@ -0,0 +1,82 @@
...
+        protected async Task<(CPSubsystemSessionState, long)> RequestNewSessionAsync(CPGroupId groupId)
+        {
+            var requestMessage = CPSessionCreateSessionCodec.EncodeRequest(groupId, Cluster.ClientId.ToString());
+            var responseMessage = await Cluster.Messaging.SendAsync(requestMessage).CfAwait();
+            var response = CPSessionCreateSessionCodec.DecodeResponse(requestMessage);
```
zpqrtbnk: You want to decode the `responseMessage` not the `requestMessage`.

src/Hazelcast.Net/CP/CpSubSystemSession.Requests.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// Closes the session on the server
+        /// </summary>
+        /// <param name="groupId"></param>
+        /// <param name="sessionId"></param>
+        /// <returns></returns>
+        protected async Task<bool> RequestCloseSessionAsync(CPGroupId groupId, long sessionId)
+        {
+            var requestMessage = CPSessionCloseSessionCodec.EncodeRequest(groupId, sessionId);
+            var responseMessage = await Cluster.Messaging.SendAsync(requestMessage).CfAwait();
+            var response = CPSessionCloseSessionCodec.DecodeResponse(requestMessage);
```
zpqrtbnk: Decoding wrong message.

src/Hazelcast.Net/CP/CpSubSystemSession.Heartbeat.cs
```diff
@@ -0,0 +1,102 @@
...
+        private async Task RunAllAsync(CancellationToken cancellationToken)
+        {
+            _logger.LogDebug("Run CP Session Heartbeat");
+
+            lock (_mutex) if (_sessions.IsEmpty) return;
+
+            IEnumerable<(CPGroupId, CPSubsystemSessionState)> sessions;
+            lock (_mutex) sessions = _sessions.Select(p => (p.Key, p.Value));
+
+            await sessions.ParallelForEachAsync((p, cancellationToken) =>
```
zpqrtbnk, line 74: .NET 6 introduces [Parallel.ForEachAsync](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.parallel.foreachasync?view=net-6.0) and we should follow this syntax to be future-proof. But we cannot add a static method to `Parallel` for pre- .NET 6. In other situations what we've done is create a `ParallelEx` class in the `Polyfills` namespace, and just copy/paste the .NET 6 code there (backporting, in other words). You'll find the code [here](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Linq.Parallel/src/System/Linq/ParallelEnumerable.cs). The benefit is that... we trust that code and don't write tests for it, since it's MS code and supposed to work.

So here it would be... `ParallelEx.ForEachAsync(...)` using the syntax that would be valid in .NET 6.

src/Hazelcast.Net/Core/TaskParallelExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Hazelcast.Core
+{
+    internal static class TaskParallelExtensions
```
zpqrtbnk, line 23: See my comment in the places where this is used. `Parallel.ForEachAsync` has been introduced in .NET 6 and we should just backport / polyfill it but use the same syntax, names and concepts. So... that class probably goes away.

src/Hazelcast.Net/CP/FencedLock.cs
```diff
@@ -0,0 +1,411 @@
...
+
+        public override ValueTask DestroyAsync()
+        {
+            _lockedThreadToSession.Clear();
+            return default;
+        }
+
+        internal class LockOwnershipState
+        {
+            public long Fence { get; private set; }
```
zpqrtbnk, line 364: You don't need the `private` setters for these properties if the fields are readonly and set by the constructor.

src/Hazelcast.Net/CP/FencedLock.cs
```diff
@@ -0,0 +1,411 @@
...
+            {
+                Fence = fence;
+                SessionId = sessionId;
+                ThreadId = threadId;
+                LockCount = lockCount;
+            }
+
+            public bool Locked => FencedLock.InvalidFence != Fence;
+
+            public bool LockedBy(long threadId, long sessionId) => Locked && SessionId == sessionId && ThreadId == threadId;
```
zpqrtbnk, line 379: Can it be "locked by" but with an "invalid fence"?

src/Hazelcast.Net/CP/FencedLock.cs
```diff
@@ -0,0 +1,411 @@
...
+
+            public LockOwnershipState(long fence, long sessionId, long threadId, int lockCount)
+            {
+                Fence = fence;
+                SessionId = sessionId;
+                ThreadId = threadId;
+                LockCount = lockCount;
+            }
+
+            public bool Locked => FencedLock.InvalidFence != Fence;
```
zpqrtbnk, line 377: OK I see the usage of the invalid fence here. Still, if the fence is invalid it's not locked, right? Then `LockedBy` is going to return `false`? maybe a short note to explain it. In fact I'm wondering whether we need 2 ctors: one for "locked" state (with fence, ids, etc) and one for "not-locked" state (with basically no args).

src/Hazelcast.Net/CP/FencedLock.cs
```diff
@@ -0,0 +1,411 @@
...
+
+            var ownership = await RequestLockOwnershipStateAsync().CfAwait();
+
+            if (ownership.LockedBy(threadId, sessionId))
+            {
+                _lockedThreadToSession[threadId] = sessionId;
+                return ownership.Fence;
+            }
+
+            VerifyNoLockedSessionExist(threadId);
```
zpqrtbnk, line 84: Here, and in many places in this class. `VerifyNoLockedSessionExist` ends up reading and writing `_lockedThreadToSession` and this dictionary is not concurrent and I am not sure that this is all thread-safe?

zpqrtbnk (changes requested):
I have commented in various places on the PR. The only thing that is a more general comment is I haven't looked into `FencedLock` and `CpSubsystemSession` in details for thread-safety, will do now.

src/Hazelcast.Net/CP/IFencedLock.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        Task LockAsync();
+
+        Task LockInterruptiblyAsync();
+
+        Task<long> LockAndGetFenceAsync();
+
+        Task<bool> TryLockAsync(TimeSpan timeout);
+
+        Task<long> TryLockAndGetFenceAsync(TimeSpan timeout);
```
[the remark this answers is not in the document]
ihsandemir: I understand your point about magic constants but i would rather keep consistency with Java if this is not a strong .NET difference. Different API, different doc, difficult for support.

src/Hazelcast.Net/CP/FencedLock.cs
```diff
@@ -0,0 +1,398 @@
...
+        private void VerifyNoLockOnThread(long threadId, long sessionId, bool releaseSession = true)
+        {
+            if (_lockedThreadToSession.TryGetValue(threadId, out var lockedSessionId) && lockedSessionId != sessionId)
+            {
+                _lockedThreadToSession.TryRemove(threadId, out var _);
+
+                if (releaseSession)
+                    _subsystemSession.ReleaseSession(CPGroupId, sessionId);
+
+                throw new LockOwnershipLostException(lockedSessionId.ToString("D"));
```
ihsandemir, line 343: Can you match the message in Java? 
`"Current thread is not owner of the Lock[" + proxyName + "] because its Session["
                + sessionId + "] is closed by server!"`

src/Hazelcast.Net/CP/CpSubsystemSession.cs
```diff
@@ -0,0 +1,271 @@
...
+        /// <summary>
+        /// Invalidates the given session after invalidation no more heartbeat will be sent.
+        /// </summary>
+        /// <param name="groupId"></param>
+        /// <param name="sessionId"></param>
+        public void InvalidateSession(CPGroupId groupId, long sessionId)
+        {
+            if (_sessions.TryGetValue(groupId, out var sessionState) && sessionState.Id == sessionId)
+            {
+                _sessions.TryRemove(groupId, out var _);
```
ihsandemir, line 96: java also tries to match the value of the map key to remove, is it possible to do the same?

src/Hazelcast.Net/CP/CPSubsystem.cs
```diff
@@ -76,6 +78,14 @@ public async Task<IAtomicReference<T>> GetAtomicReferenceAsync<T>(string name)
             return new AtomicReference<T>(objectName, groupId, _cluster, _serializationService);
         }
 
+        public async Task<IFencedLock> GetFencedLockAsync(string name)
```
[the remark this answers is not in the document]
emreyigit: This is already fixed but in my local. I pushed after your comment. Side effects of working on a changing staff 😶

src/Hazelcast.Net/CP/CPSubsystemSessionState.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <summary>
+        /// Checks session is expired to given milliseconds
+        /// </summary>
+        /// <param name="timestamp">milliseonds</param>
+        /// <returns>Is session expired</returns>
+        private bool IsSessionExpired(long timestamp)
+        {
+            long expiration = _creationTime + _ttlMiliseconds;
+
+            if (expiration < 0)
```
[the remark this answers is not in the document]
emreyigit: Agreed. Took it as it is. https://github.com/emreyigit/hazelcast/blob/40d555dc1110da9f63fd66f504c83d4382d0384b/hazelcast/src/main/java/com/hazelcast/cp/internal/session/AbstractProxySessionManager.java#L267

src/Hazelcast.Net/CP/CPSubsystemSessionState.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <returns>Session Id</returns>
+        public long Release(int count)
+        {
+            int minusCount = -count;
+            Interlocked.Add(ref _acquireCount, minusCount);
+            return _id;
+        }
+
+        /// <inheritdoc/>     
+        public override bool Equals(object obj)
```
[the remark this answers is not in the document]
emreyigit: We didn't need yet. We are already comparing them by session id like java. What do you mean by state? Do you mean all of the properties?

src/Hazelcast.Net/CP/IFencedLock.cs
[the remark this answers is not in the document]
emreyigit: @zpqrtbnk Feeling same but I think we can't `out` something from an async method.

src/Hazelcast.Net/CP/IFencedLock.cs
```diff
@@ -0,0 +1,47 @@
...
+    //TODO: documentation
+    public interface IFencedLock : ICPDistributedObject
+    {
+        long InvalidFence { get; }
+
+        ICPGroupId CPGroupId { get; }
+
+        Task LockAsync();
+
+        Task LockInterruptiblyAsync();
```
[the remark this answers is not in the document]
emreyigit: Yeah :), removed that method completely since it was heavily java oriented thing.

src/Hazelcast.Net/CP/CpSubSystemSession.Requests.cs
emreyigit: Fixed during testing :)

src/Hazelcast.Net/CP/CpSubSystemSession.Requests.cs
emreyigit: Fixed during testing :)

src/Hazelcast.Net/CP/FencedLock.cs
emreyigit: I don't think so. `LockOwnershipState` represents server side status of the lock. If the fence is already invalid, `LockedBy` will be false. See `Locked` property.

src/Hazelcast.Net/CP/FencedLock.cs
emreyigit: Sorry, you have already seen the `Locked`. I'm reading them from top to down. `LockOwnershipState ` are created in `RequestLockOwnershipStateAsync`, so it means it must be initialized with args from server. We haven't need such "no lock" state yet. If we need, we can put a static `NotLocked` in `LockOwnershipState `.

src/Hazelcast.Net/CP/FencedLock.cs
emreyigit: You are right, it is not. I also realized the same thing during testing and fixed to concurrent.

src/Hazelcast.Net/CP/CPSubsystem.cs
```diff
@@ -76,6 +78,14 @@ public async Task<IAtomicReference<T>> GetAtomicReferenceAsync<T>(string name)
...
+        {
+            var (groupName, objectName) = ParseName(name);
+            var groupId = await GetGroupIdAsync(groupName).CfAwait();
+
+            return new FencedLock(_cpSubsystemSession, objectName, groupId, _cluster);
```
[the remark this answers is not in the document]
emreyigit: Yeap, you are right. I also realized that, and fixed during testing.

src/Hazelcast.Net/CP/CpSubsystemSession.cs
```diff
@@ -0,0 +1,246 @@
...
+        //Cluster Fields
+        private readonly Cluster Cluster;
+        private readonly ClusterState State;
+        #endregion
+
+        #region SessionManagement
+        public CpSubsystemSession(Cluster cluster)
+        {
+            Cluster = cluster ?? throw new ArgumentNullException(nameof(cluster));
+            State = cluster.State ?? throw new ArgumentNullException(nameof(cluster.State));
```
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/615acd74e54cd18425eb94dfeb577ca5512e2a61

src/Hazelcast.Net/Exceptions/LockAcquireLimitReachedException.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Text;
+
+namespace Hazelcast.Exceptions
+{
+    internal class LockAcquireLimitReachedException : HazelcastException
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="LockAcquireLimitReachedException"/> class with specific message.
+        /// </summary>
+        /// <param name="message"></param>
```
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/615acd74e54cd18425eb94dfeb577ca5512e2a61

src/Hazelcast.Net/Exceptions/HazelcastInstanceNotActiveException.cs
```diff
@@ -0,0 +1,53 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+using System.Text;
+
+namespace Hazelcast.Exceptions
+{
+    [Serializable]
+    public class HazelcastInstanceNotActiveException : HazelcastException
```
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/615acd74e54cd18425eb94dfeb577ca5512e2a61

src/Hazelcast.Net/CP/CPSubsystemSessionState.cs
```diff
@@ -0,0 +1,112 @@
...
+
+namespace Hazelcast.CP
+{
+    /// <summary>
+    /// Represents the Session State on CPSubsystem
+    /// </summary>
+    internal class CPSubsystemSessionState
+    {
+        private readonly long _id;
+        private int _acquireCount = 0;
```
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/615acd74e54cd18425eb94dfeb577ca5512e2a61

src/Hazelcast.Net/CP/CPSubsystemSessionState.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        /// <summary>
+        /// Decreases the acquire count
+        /// </summary>
+        /// <param name="count">Acquire count to decrease</param>
+        /// <returns>Session Id</returns>
+        public long Release(int count)
+        {
+            int minusCount = -count;
```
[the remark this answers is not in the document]
emreyigit: It is an old habit copying to local. Changed. Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/615acd74e54cd18425eb94dfeb577ca5512e2a61

src/Hazelcast.Net/CP/CPSubsystemSessionState.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            return timestamp > expiration;
+        }
+
+        /// <summary>
+        /// Increases the acquire count
+        /// </summary>
+        /// <param name="count">Acquire count to increase</param>
+        /// <returns>Session Id</returns>
+        public long Acquire(int count)
```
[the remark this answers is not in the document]
emreyigit: Release shouldn't return something. That was my mistake, Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/615acd74e54cd18425eb94dfeb577ca5512e2a61. [Ref: Java.Acquire](https://github.com/emreyigit/hazelcast/blob/40d555dc1110da9f63fd66f504c83d4382d0384b/hazelcast/src/main/java/com/hazelcast/cp/internal/session/AbstractProxySessionManager.java#L275)

src/Hazelcast.Net/CP/IFencedLock.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        Task UnlockAsync();
+
+        Task<bool> IsLockedAsync();
+
+        Task<long> GetFenceAsync();
+
+        Task<int> GetLockCountAsync();
+
+        Task<bool> IsLockedByCurrentThreadAsync();// todo: check naming the thread
```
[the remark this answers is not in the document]
emreyigit: Agreed. I had doubts about naming as I commented on the code. Changed. Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/615acd74e54cd18425eb94dfeb577ca5512e2a61

src/Hazelcast.Net/CP/FencedLock.cs
```diff
@@ -0,0 +1,411 @@
...
+        {
+            _subsystemSession = subsystemSession;
+            _groupId = groupId;
+        }
+
+        #region IFencedLock Methods
+        /// <inheritdoc/>  
+        public async Task<long> GetFenceAsync()
+        {
+            long threadId = ContextId;
```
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/615acd74e54cd18425eb94dfeb577ca5512e2a61

src/Hazelcast.Net/CP/CpSubSystemSession.Heartbeat.cs
```diff
@@ -0,0 +1,102 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Hazelcast.CP
+{
+    /// <summary>
+    /// Implements Heartbeat side of CP Session
+    /// </summary>
+    internal partial class CpSubsystemSession
+    {
+        private readonly CancellationTokenSource _cancel;
```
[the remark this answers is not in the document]
emreyigit: Initialized in ctor, also commented on code. Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/615acd74e54cd18425eb94dfeb577ca5512e2a61

src/Hazelcast.Net/CP/CpSubSystemSession.Heartbeat.cs
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/615acd74e54cd18425eb94dfeb577ca5512e2a61

src/Hazelcast.Net/CP/CpSubSystemSession.Requests.cs
```diff
@@ -0,0 +1,82 @@
...
+    /// Implements server side requests for CP Session
+    /// </summary>
+    internal partial class CpSubsystemSession
+    {
+        /// <summary>
+        /// Generates a cluster-wide unique thread id for the caller
+        /// </summary>
+        /// <param name="groupId"></param>
+        /// <returns></returns>
+        protected async Task<long> RequestGenerateThreadIdAsync(CPGroupId groupId)
```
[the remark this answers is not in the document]
emreyigit: Made them `internal` for testing purposes. Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/615acd74e54cd18425eb94dfeb577ca5512e2a61

src/Hazelcast.Net/CP/CpSubSystemSession.Requests.cs
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/615acd74e54cd18425eb94dfeb577ca5512e2a61

src/Hazelcast.Net/CP/FencedLock.cs
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/615acd74e54cd18425eb94dfeb577ca5512e2a61

src/Hazelcast.Net/CP/IFencedLock.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Threading.Tasks;
+
+namespace Hazelcast.CP
+{
+    //TODO: documentation
+    public interface IFencedLock : ICPDistributedObject
+    {
+        long InvalidFence { get; }
+
+        ICPGroupId CPGroupId { get; }
```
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/615acd74e54cd18425eb94dfeb577ca5512e2a61

src/Hazelcast.Net/CP/FencedLock.cs
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/615acd74e54cd18425eb94dfeb577ca5512e2a61

src/Hazelcast.Net/CP/CpSubsystemSession.cs
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/615acd74e54cd18425eb94dfeb577ca5512e2a61

src/Hazelcast.Net/CP/CpSubsystemSession.cs
```diff
@@ -0,0 +1,246 @@
...
+
+namespace Hazelcast.CP
+{
+    /// <summary>
+    /// Cp Subsystem Session manages server side session requests and heartbeat
+    /// </summary>
+    internal partial class CpSubsystemSession : IAsyncDisposable
+    {
+        #region Properties
+        private readonly ConcurrentDictionary<CPGroupId, SemaphoreSlim> _groupIdSemaphores = new ConcurrentDictionary<CPGroupId, SemaphoreSlim>();
```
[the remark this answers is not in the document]
emreyigit: Removed the unused ones. Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/615acd74e54cd18425eb94dfeb577ca5512e2a61
About naming, private ones start with  underscore `_`. General usage of `Cluster` is capital among the project. That is why I didn't touch it. If it is okay for you, I can rename it to `_cluster`.

src/Hazelcast.Net/CP/CPSubsystem.cs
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/615acd74e54cd18425eb94dfeb577ca5512e2a61

src/Hazelcast.Net/CP/CPSubsystem.cs
```diff
@@ -76,6 +78,14 @@ public async Task<IAtomicReference<T>> GetAtomicReferenceAsync<T>(string name)
             return new AtomicReference<T>(objectName, groupId, _cluster, _serializationService);
         }
 
+        public async Task<IFencedLock> GetFencedLockAsync(string name)
+        {
+            var (groupName, objectName) = ParseName(name);
```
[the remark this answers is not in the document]
emreyigit: Yeah,  it is a general message. Maybe we can write a sentence which is specific to that block of code. WDYT @zpqrtbnk ?

src/Hazelcast.Net/CP/CpSubsystemSession.cs
```diff
@@ -0,0 +1,271 @@
...
+using Hazelcast.Core;
+using Hazelcast.Exceptions;
+using Microsoft.Extensions.Logging;
+
+namespace Hazelcast.CP
+{
+    /// <summary>
+    /// Cp Subsystem Session manages server side session requests and heartbeat
+    /// </summary>
+    internal partial class CPSubsystemSession : IAsyncDisposable
```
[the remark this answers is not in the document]
emreyigit: We don't use proxy in naming. Also, we don't have much proxy, I think. Session is a kind of service actually. WDYT @zpqrtbnk ?

src/Hazelcast.Net/CP/CPSubsystem.cs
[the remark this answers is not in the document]
zpqrtbnk: That is the general .NET rule, when an argument cannot be null we just say it cannot be null, same for empty strings. And exception messages in .NET don't end with a bang.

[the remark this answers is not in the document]
zpqrtbnk: So this is the general rule we use in .NET and in our client and I think we should follow it here.

src/Hazelcast.Net/CP/FencedLock.cs
```diff
@@ -0,0 +1,394 @@
...
+
+                }
+                catch (RemoteException e)
+                {
+                    if (e is RemoteException { Error: RemoteError.SessionExpiredException })
+                    {
+                        _subsystemSession.InvalidateSession(CPGroupId, sessionId);
+                        VerifyNoLockedSessionExist(threadId);
+                    }
+                    else if (e is RemoteException { Error: RemoteError.DistributedObjectDestroyed })
```
ihsandemir, line 179: why do you do specific to `DistributedObjectDestroyed `? You need to do it for all the rest of the exceptions. And you also need to do `_subsystemSession.ReleaseSession(CPGroupId, sessionId);` before re-throwing.

src/Hazelcast.Net/CP/CpSubsystemSession.cs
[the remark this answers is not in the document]
zpqrtbnk: It's capital `Cluster` whenever it's either a property or a protected field. Everything purely private is always `_cluster`. If that is not the case please point me to the places so we can fix it.

src/Hazelcast.Net/CP/CpSubsystemSession.cs
```diff
@@ -0,0 +1,264 @@
...
+        #region Properties
+        private readonly ConcurrentDictionary<CPGroupId, SemaphoreSlim> _groupIdSemaphores = new ConcurrentDictionary<CPGroupId, SemaphoreSlim>();
+        private readonly ConcurrentDictionary<CPGroupId, CPSubsystemSessionState> _sessions = new ConcurrentDictionary<CPGroupId, CPSubsystemSessionState>();
+        private int _disposed;
+        private readonly object _mutex = new object();
+        private readonly SemaphoreSlim _semaphoreReadWrite = new SemaphoreSlim(1, 1);
+        private readonly ILogger _logger;
+        private readonly Cluster Cluster;
+
+        public const int NoSessionId = -1;
```
ihsandemir, line 40:
```suggestion
        public const long NoSessionId = -1;
```

src/Hazelcast.Net/CP/FencedLock.cs
```diff
@@ -0,0 +1,394 @@
...
+        {
+            var threadId = ContextId;
+            var sessionId = _subsystemSession.GetSessionId(CPGroupId);
+            HConsole.WriteLine(this, $"Session:{sessionId} ThreadId:{threadId} -> GetFence");
+            VerifyNoLockOnThread(threadId, sessionId, false);
+
+            if (sessionId == CPSubsystemSession.NoSessionId)
+            {
+                _lockedThreadToSession.TryRemove(threadId, out var _);
+                throw new SynchronizationLockException();
```
ihsandemir, line 74: why different exception from Java? Also the exception message string.

src/Hazelcast.Net/CP/CpSubsystemSession.cs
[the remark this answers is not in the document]
zpqrtbnk: I am slightly confused by the name. A "session" is just, well, a session, not something that manages sessions. Now I respectfully hate that every single class in Java is prefixed `Client` as in `ClientProxyWhatever` because well, yes, we're writing a client, and I am even tempted to retaliate by naming my classes `DotNetClientProxyWhatever`. So please can we drop the `Client` prefix and use a name that makes sense? Like, we are managing sessions, can it be `SessionManager`? In the cp subsystem namespace? or `CpSessionManager` or something that *tells* what it does? And yes it's not going to be the same name as Java but the Java names in most cases make no sense.

src/Hazelcast.Net/CP/FencedLock.cs
```diff
@@ -0,0 +1,394 @@
...
+            var ownership = await RequestLockOwnershipStateAsync().CfAwait();
+
+            if (ownership.LockedBy(threadId, sessionId))
+            {
+                _lockedThreadToSession[threadId] = sessionId;
+                return ownership.Fence;
+            }
+
+            VerifyNoLockedSessionExist(threadId);
+            throw new SynchronizationLockException();
```
ihsandemir, line 86: newIllegalMonitorStateException?

src/Hazelcast.Net/CP/IFencedLock.cs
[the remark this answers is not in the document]
zpqrtbnk: ah well yes, my mistake, but then... we are doing it in other places, using the `Attempt<TResult>` class. So your method returns a `Task<Attempt<long>>` and you can do something similar to:
```
var attempt = await TryLockAndGetFenceAsync(...);
if (attempt)
{
  var id = attempt.Result;
}
```
this is what we do in many other places of the code when it's async.

src/Hazelcast.Net/Exceptions/DistributedObjectDestroyedException.cs
```diff
@@ -0,0 +1,62 @@
...
+// limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+using System.Text;
+
+namespace Hazelcast.Exceptions
+{
+    internal class DistributedObjectDestroyedException : HazelcastException
```
zpqrtbnk, line 22: we don't want to introduce specialized remote exceptions for now.

src/Hazelcast.Net/Exceptions/LockAcquireLimitReachedException.cs
```diff
@@ -18,12 +18,15 @@
 
 namespace Hazelcast.Exceptions
 {
+    /// <summary>
+    /// Represents exception that defined acquire limit on the lock is exceeded.
```
zpqrtbnk, line 22:
```suggestion
    /// Represents the exception that is thrown when the aquisition count limit on the lock is exceeded.
```

zpqrtbnk (commented):
More comments - and I still need to review the thread-safety.

src/Hazelcast.Net/CP/FencedLock.cs
emreyigit: This is the `IllegalMonitorException `counterpart on .NET.

src/Hazelcast.Net/CP/FencedLock.cs
emreyigit: This is the `IllegalMonitorException `counterpart on .NET.

src/Hazelcast.Net/CP/CpSubSystemSession.Heartbeat.cs
```diff
@@ -0,0 +1,101 @@
...
+        /// <param name="cancellationToken"></param>        
+        private async Task RunAsync(CPGroupId groupId, CPSubsystemSessionState sessionState, CancellationToken cancellationToken)
+        {
+            try
+            {
+                await RequestSessionHeartbeat(groupId, sessionState.Id).CfAwait();
+            }
+            catch (Exception e)
+            {
+                _logger.LogWarning(e, "CP Session Heartbeat has thrown an exception, but will continue.");
```
ihsandemir, line 97: Missing this as exist at Java 
`                        if (cause instanceof SessionExpiredException || cause instanceof CPGroupDestroyedException) {
                            invalidateSession(groupId, session.id);
                        }
`

src/Hazelcast.Net/CP/CpSubSystemSession.Heartbeat.cs
```diff
@@ -0,0 +1,101 @@
...
+        /// </summary>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        private async Task RunAllAsync(CancellationToken cancellationToken)
+        {
+            _logger.LogDebug("Run CP Session Heartbeat");
+
+            lock (_mutex) if (_sessions.IsEmpty) return;
+
+            IEnumerable<(CPGroupId, CPSubsystemSessionState)> sessions;
```
ihsandemir, line 71: Java keeps a list of previous heartbeat tasks and cancels them before starting new heartbeat tasks. Can this be done for C# since you have cancellationtoken mechanism?

src/Hazelcast.Net/CP/CpSubsystemSession.cs
```diff
@@ -0,0 +1,264 @@
...
+                {
+                    // Wait and lock only for the groupId
+                    var semaphore = GetSemaphoreBy(groupId);
+                    await semaphore.WaitAsync().CfAwait();
+
+                    try
+                    {
+                        var session = await RequestNewSessionAsync(groupId).CfAwait();
+                        ScheduleHeartbeat(TimeSpan.FromMilliseconds(session.Item2));
+                        _sessions[groupId] = session.Item1;
```
ihsandemir, line 170: java uses this order:

```suggestion
                        _sessions[groupId] = session.Item1;
                        ScheduleHeartbeat(TimeSpan.FromMilliseconds(session.Item2));
```
puts the session in the map even if the schedule heartbeat throws.

src/Hazelcast.Net/CP/FencedLock.cs
```diff
@@ -0,0 +1,394 @@
...
+        {
+            var threadId = ContextId;
+            var sessionId = _subsystemSession.GetSessionId(CPGroupId);
+
+            VerifyNoLockOnThread(threadId, sessionId, false);
+
+            if (sessionId == CPSubsystemSession.NoSessionId)
+            {
+                _lockedThreadToSession.TryRemove(threadId, out var _);
+                throw new SynchronizationLockException();
```
ihsandemir, line 291: different from java

src/Hazelcast.Net/CP/FencedLock.cs
```diff
@@ -0,0 +1,394 @@
...
+
+                _subsystemSession.ReleaseSession(CPGroupId, sessionId);
+            }
+            catch (RemoteException e)
+            {
+                if (e is RemoteException { Error: RemoteError.SessionExpiredException })
+                {
+                    _subsystemSession.InvalidateSession(CPGroupId, sessionId);
+                    _lockedThreadToSession.TryRemove(threadId, out var _);
+                    throw new SessionExpiredException(e);
```
ihsandemir, line 312: different exception from java: throw newLockOwnershipLostException(sessionId);

src/Hazelcast.Net/CP/CPSubsystemSessionState.cs
```diff
@@ -88,12 +92,9 @@ public long Acquire(int count)
...
         /// </summary>
         /// <param name="count">Acquire count to decrease</param>
-        /// <returns>Session Id</returns>
-        public long Release(int count)
+        public void Release(int count)
         {
-            int minusCount = -count;
-            Interlocked.Add(ref _acquireCount, minusCount);
-            return _id;
+            Interlocked.Add(ref _acquireCount, -count);
```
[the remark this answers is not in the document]
emreyigit: Yeap, it's same as java impl., and there is no check on less than acquired. WDYT @ihsandemir ? 

Also, note that `Release()` is not exposed to public.

src/Hazelcast.Net/CP/FencedLock.cs
```diff
@@ -0,0 +1,394 @@
...
+                }
+                else if (e is RemoteException { Error: RemoteError.IllegalMonitorState })
+                {
+                    _lockedThreadToSession.TryRemove(threadId, out var _);
+                    throw new SynchronizationLockException("IllegalMonitorState exception is thrown by server. Are you trying to unlock without holding?", e);
+                }
+            }
```
ihsandemir, line 319: should you throw for all other `RemoteException` cases?

src/Hazelcast.Net/Exceptions/DistributedObjectDestroyedException.cs
emreyigit: So, should I just throw `RemoteException` ? Most of the tests look for `DistributedObjectDestroyedException` whether is thrown.

src/Hazelcast.Net/CP/CPSubsystem.cs
```diff
@@ -136,5 +150,13 @@ public static (string groupName, string objectName) ParseName(string name)
 
             return (groupName, objectName);
         }
+
+        public async ValueTask DisposeAsync()
+        {
+            foreach (var item in _cpObjectsByName.Values)            
+                await item.DestroyAsync().CfAwait();
```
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/5172b1429ee358627e0192663ae666a29610b29f

src/Hazelcast.Net/CP/CPSubsystemSessionState.cs
```diff
@@ -88,12 +92,9 @@ public long Acquire(int count)
         /// Decreases the acquire count
```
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/5172b1429ee358627e0192663ae666a29610b29f I didn't commit the suggestions not to increase number of commits, and to cause confusion, fyi.

src/Hazelcast.Net/CP/CPSubsystemSessionState.cs
```diff
@@ -88,12 +92,9 @@ public long Acquire(int count)
...
         /// </summary>
         /// <param name="count">Acquire count to decrease</param>
```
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/5172b1429ee358627e0192663ae666a29610b29f

src/Hazelcast.Net/CP/CpSubsystemSession.cs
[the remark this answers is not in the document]
emreyigit: Renamed to `CpSessionManager` , _cpSubsystemSession to `_cpSessionManager`. Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/5172b1429ee358627e0192663ae666a29610b29f

src/Hazelcast.Net/CP/CpSubsystemSession.cs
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/5172b1429ee358627e0192663ae666a29610b29f

src/Hazelcast.Net/CP/CpSubsystemSession.cs
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/5172b1429ee358627e0192663ae666a29610b29f

src/Hazelcast.Net/CP/CpSubsystemSession.cs
```diff
@@ -148,20 +147,18 @@ await _sessions.Keys
         /// <exception cref="HazelcastInstanceNotActiveException"></exception>
         private async Task<CPSubsystemSessionState> GetOrCreateSessionAsync(CPGroupId groupId)
         {
-            readerWriterLock.EnterUpgradeableReadLock();
+            await _semaphoreReadWrite.WaitAsync().CfAwait();
 
             try
             {
-                if (_disposed == 1) throw new HazelcastInstanceNotActiveException("CP Subsystem Session is already disposed!");
+                if (_disposed == 1) throw new ObjectDisposedException("CP Subsystem Session is already disposed!");
```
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/5172b1429ee358627e0192663ae666a29610b29f

src/Hazelcast.Net/CP/CpSubsystemSession.cs
```diff
@@ -173,15 +170,18 @@ private async Task<CPSubsystemSessionState> GetOrCreateSessionAsync(CPGroupId gr
                         _sessions[groupId] = session.Item1;
                         return session.Item1;
                     }
+                    catch (Exception e)
+                    {
+                        throw;
```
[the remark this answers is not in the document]
emreyigit: Left over from debugging, my mistake. Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/5172b1429ee358627e0192663ae666a29610b29f

src/Hazelcast.Net/CP/IFencedLock.cs
```diff
@@ -0,0 +1,47 @@
...
+// limitations under the License.
+
+using System;
+using System.Threading.Tasks;
+
+namespace Hazelcast.CP
+{
+    //TODO: documentation
+    public interface IFencedLock : ICPDistributedObject
+    {
```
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/7412acbc14cfab334e2a91cbdcbc7b7756c606cd

src/Hazelcast.Net/Exceptions/LockAcquireLimitReachedException.cs
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/5172b1429ee358627e0192663ae666a29610b29f

src/Hazelcast.Net/CP/CPSubsystemSessionState.cs
```diff
@@ -0,0 +1,113 @@
...
+
+namespace Hazelcast.CP
+{
+    /// <summary>
+    /// Represents the Session State on CPSubsystem
+    /// </summary>
+    internal class CPSubsystemSessionState
+    {
+        private readonly long _id;
+        private int _acquireCount;
```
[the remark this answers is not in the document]
emreyigit: It is used as atomic, please see line 87
`Interlocked.Add(ref _acquireCount, count);`

src/Hazelcast.Net/CP/FencedLock.cs
```diff
@@ -0,0 +1,394 @@
...
+                        _subsystemSession.InvalidateSession(CPGroupId, sessionId);
+                        VerifyNoLockedSessionExist(threadId);
+
+                        long duration = Clock.Milliseconds - start;
+
+                        if (duration <= 0)
+                            return InvalidFence;
+
+                    }
+                    else if (e is RemoteException { Error: RemoteError.DistributedObjectDestroyed })
```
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/4cac01cbb5360d58415aa9bbe55d6d04f9c848e5

src/Hazelcast.Net/CP/FencedLock.cs
emreyigit: Java client throws only for these cases. I have followed that.

src/Hazelcast.Net/CP/IFencedLock.cs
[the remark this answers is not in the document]
emreyigit: Got it, thanx. I checked the `Attempt` but it is internally accessible. We have to make it public in order to use.

src/Hazelcast.Net/CP/CpSubSystemSession.Heartbeat.cs
emreyigit: I found the `Parallel.ForeachAsync` at [here](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Threading.Tasks.Parallel/src/System/Threading/Tasks/Parallel.ForEachAsync.cs) However, it requires nested things to be ported as well. So, I'm canceling the custom parallel for each for now since it needs more time than we think.

src/Hazelcast.Net/CP/CPSubsystem.cs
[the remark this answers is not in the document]
ihsandemir: Will be updated similar to java putIfAbsent.

src/Hazelcast.Net/CP/FencedLock.cs
ihsandemir: remove DistributedObjectDestroyed? waitkeycancelled case should throw IllegalMonitorStateException as in Java.

src/Hazelcast.Net/CP/CPSubsystemSessionState.cs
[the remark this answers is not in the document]
ihsandemir: Please also check getter `bool IsInUse => _acquireCount > 0`

src/Hazelcast.Net/CP/CpSubSystemSession.Heartbeat.cs
ihsandemir: Does not fit well with C# task mechanism.

src/Hazelcast.Net/CP/FencedLock.cs
ihsandemir: Still will be changed as Java.

emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/993cabbcb9bac708cc38ea83239b98cf19ead875

src/Hazelcast.Net/CP/FencedLock.cs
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/993cabbcb9bac708cc38ea83239b98cf19ead875

src/Hazelcast.Net/CP/CpSubsystemSession.cs
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/993cabbcb9bac708cc38ea83239b98cf19ead875

src/Hazelcast.Net/CP/CpSubSystemSession.Heartbeat.cs
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/993cabbcb9bac708cc38ea83239b98cf19ead875

src/Hazelcast.Net/CP/CPSubsystemSessionState.cs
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/993cabbcb9bac708cc38ea83239b98cf19ead875

src/Hazelcast.Net/CP/CpSubsystemSession.cs
```diff
@@ -0,0 +1,264 @@
...
+                }
+                else
+                {
+                    // Wait and lock only for the groupId
+                    var semaphore = GetSemaphoreBy(groupId);
+                    await semaphore.WaitAsync().CfAwait();
+
+                    try
+                    {
+                        var session = await RequestNewSessionAsync(groupId).CfAwait();
```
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/993cabbcb9bac708cc38ea83239b98cf19ead875

src/Hazelcast.Net/CP/CpSubsystemSession.cs
```diff
@@ -0,0 +1,264 @@
...
+        /// </summary>
+        /// <param name="groupId"></param>
+        /// <returns><see cref="SemaphoreSlim"/></returns>
+        private SemaphoreSlim GetSemaphoreBy(CPGroupId groupId)
+        {
+            if (_groupIdSemaphores.TryGetValue(groupId, out var mutex))
+                return mutex;
+
+            var newMutex = new SemaphoreSlim(1, 1);
+            _groupIdSemaphores[groupId] = newMutex;
```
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/993cabbcb9bac708cc38ea83239b98cf19ead875

src/Hazelcast.Net/CP/CpSubsystemSession.cs
```diff
@@ -0,0 +1,264 @@
...
+
+        /// <summary>
+        /// Gets or createas a session if absent by group id.
+        /// </summary>
+        /// <param name="groupId"></param>
+        /// <returns><see cref="CPSubsystemSessionState"/></returns>
+        /// <exception cref="HazelcastInstanceNotActiveException"></exception>
+        private async Task<CPSubsystemSessionState> GetOrCreateSessionAsync(CPGroupId groupId)
+        {
+            await _semaphoreReadWrite.WaitAsync().CfAwait();
```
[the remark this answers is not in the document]
emreyigit: Description added for reasoning. Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/993cabbcb9bac708cc38ea83239b98cf19ead875

src/Hazelcast.Net/CP/CpSubsystemSession.cs
```diff
@@ -0,0 +1,264 @@
...
+    /// Cp Subsystem Session manages server side session requests and heartbeat
+    /// </summary>
+    internal partial class CPSubsystemSession : IAsyncDisposable
+    {
+        #region Properties
+        private readonly ConcurrentDictionary<CPGroupId, SemaphoreSlim> _groupIdSemaphores = new ConcurrentDictionary<CPGroupId, SemaphoreSlim>();
+        private readonly ConcurrentDictionary<CPGroupId, CPSubsystemSessionState> _sessions = new ConcurrentDictionary<CPGroupId, CPSubsystemSessionState>();
+        private int _disposed;
+        private readonly object _mutex = new object();
+        private readonly SemaphoreSlim _semaphoreReadWrite = new SemaphoreSlim(1, 1);
```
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/993cabbcb9bac708cc38ea83239b98cf19ead875

src/Hazelcast.Net/CP/FencedLock.cs
emreyigit: As current behavior, throws only `RemoteException. `Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/993cabbcb9bac708cc38ea83239b98cf19ead875

src/Hazelcast.Net/CP/FencedLock.cs
emreyigit: As current behavior, throws only RemoteException. Fixed at https://github.com/hazelcast/hazelcast-csharp-client/commit/993cabbcb9bac708cc38ea83239b98cf19ead875

src/Hazelcast.Net/CP/CPSubsystem.cs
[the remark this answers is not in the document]
ihsandemir: I think it should be `GetOrAdd` [method](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=net-6.0) . Can you check it?

src/Hazelcast.Net/CP/FencedLock.cs
ihsandemir: For WaitKeyCancelled , please change it to a new RemoteException with error code IllegalStateError similar to  Java. It changes the rethrown exception.

src/Hazelcast.Net/CP/CpSubsystemSession.cs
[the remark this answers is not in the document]
ihsandemir: Please change to GetOrAdd.

src/Hazelcast.Net/CP/FencedLock.cs
ihsandemir: Will be discussed in another issue.

src/Hazelcast.Net/CP/FencedLock.cs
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/590/commits/ded58466decf990883746cd622035d8c87662675

src/Hazelcast.Net/CP/FencedLock.cs
emreyigit: Related issue https://github.com/hazelcast/hazelcast-csharp-client/issues/592

ihsandemir (approved):

PR #598, opened 2022-03-21
Complex Sql Example [API-1299]

emreyigit:
Internally asked for an example about complex objects on SQL. Implemented with explanations. I think it would be nice to have it in official ones.


alisengul53 (approved):
LGTM

zpqrtbnk (approved):
OK

PR #602, opened 2022-03-23
DNS Host name test [API-1247]

emreyigit:
Missing tests are added.


src/Hazelcast.Net.Tests/Networking/NetworkHostnameTests.cs
```diff
@@ -0,0 +1,134 @@
...
+using System.Threading.Tasks;
+using Hazelcast.Testing;
+using NUnit.Framework;
+using Hazelcast.Core;
+using Hazelcast.Testing.Remote;
+using System.Threading;
+
+namespace Hazelcast.Tests.Networking
+{
+    internal class NetworkHostnameTests : RemoteTestBase
```
zpqrtbnk, line 28: You should use the already existing `ClusterRemoteTestBase` and override its `RcClusterConfiguration` property. It already provides the `RcClient`, `RcCluster` etc properties. Unless there's a reason not to do so, but I cannot figure it out.

src/Hazelcast.Net.Tests/Networking/NetworkHostnameTests.cs
```diff
@@ -0,0 +1,134 @@
...
+            if (RcClient != null)
+            {
+                if (RcCluster != null)
+                    await RcClient.ShutdownClusterAsync(RcCluster).CfAwait();
+                await RcClient.ExitAsync().CfAwait();
+            }
+        }
+
+        [Test]
+        public async Task ConnectWithDNSHostnames()
```
zpqrtbnk, line 55: I don't understand what we're trying to test. Maybe a comment is required. Are we testing that `options.Networking.Addresses.Add("localhost");` do work? And then why do we need to test against 2 members? Why do we need a special member configuration file? Are we testing that a *member* can be configured with `localhost` and why is this a test that is performed by clients and not by the core/member code itself?

src/Hazelcast.Net.Tests/Networking/NetworkHostnameTests.cs
```diff
@@ -0,0 +1,134 @@
...
+            {
+                Assert.AreEqual(1, client.Members.Count);
+
+            }, 20_000, 500);
+
+            Assert.IsTrue(eventTriggered);
+        }
+
+        [Test]
+        public async Task ListenersWorkingWhenDNSHostnamesAreUsed()
```
zpqrtbnk, line 94: What is the purpose of the test?

src/Hazelcast.Net.Tests/Networking/NetworkHostnameTests.cs
emreyigit: Yes, you are right. It just tests to connect via hostname. I ported them as it is in java. However, I took them from linked old PR but the tests are changed/removed completely on the current version of the master(hazelcast). It's silly me I didn't think that tests could be removed/changed today. I/we should be more carefully about the old issues. I think it is not a valid issue anymore. (Previous comment lost while posting)

emreyigit:
Interesting that tests are changed now. I ported them as it is in java. However, I took them from linked old PR but the tests are changed/removed completely on the current version of the master(hazelcast). I realized now while looking for perma-link for tests. It's silly me I didn't think that tests could be removed/changed today. I/we should be more carefully about the old issues. I think it is not a valid issue anymore. cc: @ihsandemir

src/Hazelcast.Net.Tests/Networking/NetworkHostnameTests.cs
emreyigit: Good catch, thanx. I think the PR is not valid anymore.

src/Hazelcast.Net.Tests/Networking/NetworkHostnameTests.cs
emreyigit: Please see the comments below. I think the PR is not valid anymore.

ihsandemir:
@emreyigit Why do you think the test does not exist? I see [ClientRegressionWithRealNetworkTest](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/test/java/com/hazelcast/client/ClientRegressionWithRealNetworkTest.java) at the master. I assume that we do counterpart test for .NET.

emreyigit:
The issue mentions only test `com.hazelcast.client.ClientRegressionWithRealNetworkTest.testConnectWithDNSHostnames` which is not exist anymore. I didn't consider the whole class.

emreyigit:
Will continue on https://github.com/hazelcast/hazelcast-csharp-client/issues/609. Closed due to changes on the original issue.

PR #608, opened 2022-03-30
Fix Unisocket [API-1218] [API-1276]

emreyigit:
This PR fixes the unisocket connection problem. .Net client was having trouble while connecting to server with unisocket mode if the configured IP address is different than in the member table. I kept the `RemoteAddress` which is returned by the member during authentication. If connection mod is unisocket and cloud is not enabled, connection acts like connected via `RemoteAddress` even socket connection established with configured one. Tested against to cloud and to local k8s setup with unisocket mode.

### DIFF
--- a/src/Hazelcast.Net.Tests/Remote/UnisocketTests.cs
+++ b/src/Hazelcast.Net.Tests/Remote/UnisocketTests.cs
@@ -102,8 +102,6 @@ namespace Hazelcast.Tests.Remote
 
             var countOfClients = int.Parse(Encoding.UTF8.GetString(response.Result));
 
-            Assert.AreEqual(0, countOfClients);
-
             // instance1 shouldn't have any clients due to unisocket mode
             script = "result = instance_0.getClientService().getConnectedClients().size().toString();";
             response = await RcClient.ExecuteOnControllerAsync(RcCluster.Id, script, Hazelcast.Testing.Remote.Lang.JAVASCRIPT);


src/Hazelcast.Net/Clustering/MemberConnection.cs
```diff
@@ -142,8 +145,14 @@ public Action<ClientMessage> ReceivedEvent
 
         /// <summary>
         /// Gets the network address the client is connected to.
+        /// <para>In case of unisocket mode, gets the <see cref="RemoteAddress"/></para>
```
ihsandemir, line 148: Correction:
In case of unisocket mode and hazelcast cloud is not enabled, gets the <

src/Hazelcast.Net/CP/CPSessionManager.cs
```diff
@@ -240,7 +240,8 @@ public async ValueTask DisposeAsync()
 
             try
             {
-                await _heartbeating.CfAwaitCanceled();
+                if (_heartbeating != null)
```
ihsandemir, line 243: Is this related to this issue fix? Looks like another unrelated fix to me.

ihsandemir (changes requested):
Will you add some unisocket tests for unisocket test coverage? We normally want to port most of related tests from Java.

src/Hazelcast.Net/CP/CPSessionManager.cs
emreyigit: I faced with an error on unisocket mode. I didn't dig it but I think because connection is not established, heartbeat is failing during disposing.

zpqrtbnk: I don't understand. We cannot add a null check without understanding why it's null. It may be hiding another issue.

src/Hazelcast.Net/Clustering/MemberConnection.cs
```diff
@@ -142,8 +145,14 @@ public Action<ClientMessage> ReceivedEvent
...
+        /// </summary>
+        public NetworkAddress Address { get { return _networkingOptions.SmartRouting || _networkingOptions.Cloud.Enabled ? _address : RemoteAddress; } }
+
+        /// <summary>
+        /// Gets network address responed by the member
```
zpqrtbnk, line 153: `// Gets the network address returned by the member.`

And, it's not "returned" by the member... or is it? Or is it... "provided by the members list"?

zpqrtbnk:
I don't understand the PR. It says that 3 files were changed, and it adds a `RemoteAddress` property to `MemberConnection` but I fail to see where the property is used and how things change? I fail to see how it addresses the issue.

src/Hazelcast.Net/CP/CPSessionManager.cs
emreyigit: Okay. Understood the reason. `_heartbeating` is a `Task` to keep track of hear beating process. It is created when needed but disposed all the time. Fixed at 529b03865325286f2dc1300d878f28465c62fcf9

src/Hazelcast.Net/Clustering/ClusterMembers.cs
```diff
@@ -757,7 +757,8 @@ public MemberInfo GetMember(Guid memberId)
         /// <inheritdoc />
         public async ValueTask DisposeAsync()
         {
-            await _memberConnectionQueue.DisposeAsync().CfAwait();
+            if (_memberConnectionQueue != null)
```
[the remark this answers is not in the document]
emreyigit: Fixed at 529b03865325286f2dc1300d878f28465c62fcf9.

src/Hazelcast.Net/Clustering/MemberConnection.cs
```diff
@@ -142,8 +145,14 @@ public Action<ClientMessage> ReceivedEvent
 
         /// <summary>
         /// Gets the network address the client is connected to.
+        /// <para>In case of unisocket mode, gets the <see cref="RemoteAddress"/></para>
+        /// </summary>
+        public NetworkAddress Address { get { return _networkingOptions.SmartRouting || _networkingOptions.Cloud.Enabled ? _address : RemoteAddress; } }
```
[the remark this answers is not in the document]
emreyigit: When `SmartRouting=false`

**Case 1: Connect to OnPrem Cluster** 

If the configured address is different than that returned by member which is connected and authenticated on(result of authention), during very first connection, client fails because of difference between addresses. Here https://github.com/hazelcast/hazelcast-csharp-client/blob/bcd328c0659c3a5b9c9edddf0cb20fc0fb64eda7/src/Hazelcast.Net/Clustering/ClusterMembers.cs#L453 returns false, due to differences and client cannot change its state to `Connected`. 
Also, that connection is removed at 
https://github.com/hazelcast/hazelcast-csharp-client/blob/bcd328c0659c3a5b9c9edddf0cb20fc0fb64eda7/src/Hazelcast.Net/Clustering/ClusterMembers.cs#L439

**Case 2: Connect to Cloud**

Member address and provided by cloud service is public and there is no need to use address provided by member during authentication which is private. 

The remote address is taken from authentication result although socket connection can be opened via configured address member replies from different address that confuses the client with current architecture. The fix does not change socket address but adds a "alias" on it with replied one by member (Go and java also behaves similar). When smart routing is enabled, situation is managed by client since it can reconnect by using member table.

Fix bypasses the https://github.com/hazelcast/hazelcast-csharp-client/blob/bcd328c0659c3a5b9c9edddf0cb20fc0fb64eda7/src/Hazelcast.Net/Clustering/ClusterMembers.cs#L453


**If configured and returned address are same, client can connect without any trouble.**

zpqrtbnk (approved):
We have moved the entire fix to `ClusterMembers` with a large blob of comments explaining the reasoning & tradeoffs - approved.

zpqrtbnk (approved):
Approved.

src/Hazelcast.Net.Tests/Remote/UnisocketTests.cs
```diff
@@ -0,0 +1,108 @@
...
+            {
+                await ShutdownMember(member.Key);
+            }
+        }
+        /// <summary>
+        /// Port of testClientListener_withDummyClient
+        /// </summary>
+        /// <returns></returns>
+        [Test]
+        public async Task TestEventsWithDummyClient()
```
ihsandemir, line 27: Java test name `testClientListener_withDummyClient`

src/Hazelcast.Net.Tests/Remote/UnisocketTests.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            var countOfClients = int.Parse(Encoding.UTF8.GetString(response.Result));
+
+            // instance1 shouldn't have any clients due to unisocket mode
+            script = "result = instance_1.getClientService().getConnectedClients().size().toString();";
+            response = await RcClient.ExecuteOnControllerAsync(RcCluster.Id, script, Hazelcast.Testing.Remote.Lang.JAVASCRIPT);
+            countOfClients += int.Parse(Encoding.UTF8.GetString(response.Result));
+
+            //sum should be 1 since don't know which member is connected to
+            Assert.AreEqual(1, countOfClients);
```
ihsandemir, line 104: Can you also map put 100 entries and check the count again just in case there may be a late opening connection?

src/Hazelcast.Net.Tests/Remote/UnisocketTests.cs
emreyigit: I mentioned to name on the description. Also, listeners mean events on .Net side, so tried to conserve it too.

src/Hazelcast.Net.Tests/Remote/UnisocketTests.cs
emreyigit: What do you mean by late opening?

ihsandemir: @emreyigit  Will update the test.

ihsandemir:
I am OK with the PR after @emreyigit updates the test as commented and discussed with @emreyigit and you can merge the PR without my approval.

zpqrtbnk (approved):
Approved.

zpqrtbnk:
OK I am going to bypass the checks and merge the PR, since Ihsan approved in a comment.

PR #611, opened 2022-04-07
Member list printing is limited [API-1277] [API-1278]

emreyigit:
Limited the printing unchanged member list to keep logs more clean, and added test to verify.


src/Hazelcast.Net/Clustering/ClusterMembers.cs
```diff
@@ -291,20 +292,33 @@ private void LogDiffs(MemberTable table, Dictionary<MemberInfo, int> diff)
...
+
                 msg.Append(' ');
                 msg.Append(status);
                 msg.AppendLine();
             }
             msg.Append('}');
 
-            _logger.LogInformation(msg.ToString());
+            //Print only if there is a change
+            if (countOfUnchanged != table.Count)
```
zpqrtbnk, line 320: can this happen? or is it only for safety reasons? like, we get 3 unchanged members and that is all *yet* the table contains 4 members, so probably something is wrong, and it's better to log?

alisengul53:
It would be nice to update the description to run Soak tests. 
`--hazelcast:examples:soak1:Duration="HH:MM:SS"` we are able to set but it doesn't work.
It works like that
`--hazelcast:examples:soak1:Duration=2h` for 2 hours or `--hazelcast:examples:soak1:Duration=2m` for 2 minutes

src/Hazelcast.Net/Clustering/ClusterMembers.cs
emreyigit: I couldn't find a way that can occur. diff and members are already converged while filling the diff dictionary but checked the count against to diff to be more strictive. Accepting the diff as a source of truth, and if all entries are unchanged then we can skip the logging.

Also, I realized that I checked the method call as AtMost(1) which could be false positive. Imroved that one too on https://github.com/hazelcast/hazelcast-csharp-client/commit/e3ec104b568f9c6dc411c5dcc0b35f0331c6e51f

emreyigit:
@alisengul53 fixed time span parsing and hanging state during end of the test on https://github.com/hazelcast/hazelcast-csharp-client/pull/611/commits/2f5db1dd9fd3dceceae91bc439425a48342e563d

zpqrtbnk (dismissed):
Approved.

alisengul53 (dismissed):
Looks good

alisengul53 (approved):

zpqrtbnk (approved):
Approved.

PR #619, opened 2022-05-11
Refactor FencedLock with explicit contexts

zpqrtbnk:
As discussed, this refactors our `FencedLock` implementation with explicit lock context.

Previously:
```csharp
// lock the fenced lock on the "current async context" (implicit)
avait fencedLock.LockAsync();
```
Now:
```csharp
// lock the fenced lock on an explicit context
var lockContext = new LockContext();
await fencedLock.LockAsync(lockContext);
```

Every `IFencedLock` method has been refactored to accept a `LockContext` parameter. The lock context needs to be explicitly instanciated and passed around in code. Each lock context instance has a unique `long` identifier which is used for the thread identifier argument of all codecs.

Note that, due to how async works, nothing prevents one lock context to be shared by two or more concurrent tasks (whereas the Java thread identifier can only be used by the very thread that owns it). Therefore, the `LockContext` class also implements an internal client-side locking mechanism that ensures that each `IFencedLock` implementation method has exclusive access to the context, just like their Java counterparts have exclusive access to the thread. 

This means that, provided that each `IFencedLock` implementation method begins with:
```csharp
using var contextLock = await lockContext.GetLockAsync().CfAwait(); // lock the context
```
The rest of the method can be identical to the Java code and does not need bother with concurrency more than Java does.

*Thoughts: I find this way of doing things more intrusive than the implicit `AsyncContext`-based version and then, explaining the `AsyncContext` to application developers has proved difficult in the past, so maybe this approach is simpler and way easier for people to understand. We may want to consider doing the same thing for maps, although it would mean adding an extra (optional?) parameter to every map method...*


yuce:
@zpqrtbnk Regarding this example:
```
using var contextLock = await lockContext.GetLockAsync().CfAwait(); // lock the context
```
Does it have to be async? I would've thought retrieving the lock from the context would be sync.

zpqrtbnk:
It has to be async. Locking the context (so that only 1 method can execute at a time for that context) is achieved by acquiring a semaphore, which is an async operation.

yuce:
I could be wrong, but as far as I see using a semaphore to retrieve the context ID may be problematic. What if the user calls `LockAndGetFenceAsync` and then `TryLockAndGetFenceAsync`? `TryLockAndGetFenceAsync` should return immediately `false`, but it won't be able to get the context ID, so it will block until the call to `LockAndGetFenceAsync` returns.

src/Hazelcast.Net/CP/FencedLock.cs
```diff
@@ -114,8 +100,18 @@ public async Task<long> GetFenceAsync()
         /// <inheritdoc/>  
         public async Task<int> GetLockCountAsync()
         {
-            var contextId = AsyncContext.Current.Id; // the asynchronous context identified (equivalent to Java thread identifier)
-            using var contextLock = await _contextLocker.LockAsync(contextId).CfAwait(); // lock the asynchronous context
+            // the original Java code (below) does some sanity-checking on this operation
```
ihsandemir, line 103: Can we please make sync with Java and if there is a valid argument for improvement, let's also change the Java side as well. I can not reason why we do not need the checks in Java. Can you please elaborate why it is not required?

src/Hazelcast.Net/CP/IFencedLock.cs
```diff
@@ -19,104 +19,126 @@
...
-        Task UnlockAsync();
+        /// <param name="lockContext">The <see cref="LockContext"/>.</param>
+        Task UnlockAsync(LockContext lockContext);
 
         /// <summary>
         /// Checks whether this lock is locked or not.
         /// </summary>
-        /// <returns><see cref="true"/> if the lock was acquired otherwise <see cref="false"/></returns>
+        /// <returns><c>true</c> if the lock was acquired; otherwise <c>false</c>.</returns>
         Task<bool> IsLockedAsync();
```
ihsandemir, line 122:
```suggestion
        Task<bool> IsLockedAsync(LockContext lockContext);
```

src/Hazelcast.Net/CP/IFencedLock.cs
```diff
@@ -19,104 +19,126 @@
...
+        /// <returns>The fencing token, if the lock is held by the current context; otherwise <see cref="InvalidFence"/>.</returns>
+        Task<long> GetFenceAsync(LockContext lockContext);
 
         /// <summary>
-        /// Gets the reentrant lock count if the lock is held by any <see cref="AsyncContext"/> or thread in the cluster.
+        /// Gets the reentrant lock count of the lock.
         /// </summary>
-        /// <returns>The reentrant lock count if the lock is held by any <see cref="AsyncContext"/> or thread in the cluster</returns>
+        /// <returns>The reentrant lock count of the lock, or zero if it is not locked.</returns>
         Task<int> GetLockCountAsync();
```
ihsandemir, line 135:
```suggestion
        Task<int> GetLockCountAsync(LockContext lockContext);
```

ihsandemir (changes requested):

src/Hazelcast.Net/CP/FencedLock.cs
zpqrtbnk: This `GetLockCountAsync` method gets the number of times the lock has been locked, it would be zero if the lock is not locked, 1 if locked once, 2 if re-entered once, etc. It has nothing to do with the current thread / context. Same happens with `IsLockedAsync` method that determines whether the lock has been locked by any thread, regardless of the current thread / context.

Yet the Java code uses the opportunity to do some cleanup and sanity check. For instance, if the lock is not owned by the current thread, it verifies that the current thread does not *think* it owns the lock, else it throws a `LockOwnershipLostException`.

This is rather bizare: you just want to know whether the lock is locked, or how many times it is locked, and you can get an exception because you lost the lock. IMHO these are totally different concerns, it is confusing and makes no sense.

But I understand the need to align with Java. It's quite easy to do and I kinda expected you to ask ;) Just let me know what you prefer.

zpqrtbnk:
>What if the user calls `LockAndGetFenceAsync` and then `TryLockAndGetFenceAsync` in *another task* ?

This is a valid concern. The scenario you describe cannot happen in Java because Java is thread-bound, so that would have to be *another thread* and then it would have a different thread ID. Since we're context-bound, it's different.

> `TryLockAndGetFenceAsync` should immediately return false.

Well, no. Imagine calling `lockAndGetFence` and `tryLockAndGetFence` simultaneously one two different threads in Java. You cannot determine which one is going to hit the cluster first, so you cannot state that `tryLockAndGetFence` should immediately return false. Maybe it returns true and the call to `lockAndGetFence` now has to wait.

It's similar in .NET. First thing is, if `LockAndGetFenceAsync` locks before `TryLockAndGetFenceAsync` runs, then `TryLockAndGetFenceAsync` will return true anyways (not false) since we are in the same context and the lock is re-entrant. Second, it may fail and return false if the re-entrancy limit is reached. We don't know, and both need to talk to the server before they can decide.

This is why all methods (copied from Java) follow the same pattern: (1) talk to the server and (2) update some data structures. Again, this is safe in Java because it happens in 1 thread so nothing else can happen between (1) and (2). In .NET it can. Think about this situation:

* Task 1 talks to server, acquires lock
* Task 2 talks to server, *destroys* lock
* Task 2 updates internal structures
* Task 1 updates internal structures -> *inconsistency*

The context lock (semaphore) ensures that this cannot happen. Yes, it linearizes the execution of each method *within a given context* with the benefit that each method's code can stay close to the Java equivalent. I am not yet entirely sure whether there is a way to rewrite all methods entirely to avoid that lock.

src/Hazelcast.Net/CP/FencedLock.cs
emreyigit: > This is rather bizare: you just want to know whether the lock is locked, or how many times it is locked, and you can get an exception because you lost the lock. IMHO these are totally different concerns, it is confusing and makes no sense.

Feeling same.

ihsandemir: Let's talk. i expect all the clients behave the same and fix all or none.

ihsandemir:
> And yet, there _is_ an issue here:
> 
> * Context 1 calls `LockAsync` and gets the lock
> * Context 2 Task 1 calls `LockAsync` and is blocked, waiting for the lock
> * Context 2 Task 2 calls `TryLockAsync` and should return immediately (since Context2 _cannot_ lock now)
> 
> BUT in fact with our current solution, Context 2 Task 2 is going to wait on the context lock until the Context 2 Task 1 call has completed. This is bad. The PR has a flaw. Meh. Back to work.

Don't context1 and context2 use totally unrelated semaphores? Why did you say `Context 2 Task 1 calls `LockAsync` and is blocked, waiting for the lock`?

emreyigit:
@yuce 
> What if the user calls LockAndGetFenceAsync and then TryLockAndGetFenceAsync in another task ? 
I think it's responsibility of user. User should aware of that each different flow should have its own context. We keep remembering that number of tasks belong to same context are just a one thread in Java. User should keep track of that a context run by a one task; otherwise I don't see a way out. Context mimics a thread, not other way around or not task=thread.

src/Hazelcast.Net/CP/FencedLock.cs
ihsandemir: My preference: Sync with Java or otherwise fix Java as well.

zpqrtbnk:
Here, I have just pushed a new, rather different iteration or the concurrency fixes. Let me know your thoughts.

src/Hazelcast.Net/CP/FencedLock.cs
```diff
@@ -30,12 +30,13 @@ namespace Hazelcast.CP
...
-        private readonly ConcurrentDictionary<long, long> _lockedSessionIds = new ConcurrentDictionary<long, long>(); // context -> session
+        private readonly Dictionary<long, LockState> _locks = new Dictionary<long, LockState>();
+        private readonly object _locksMutex = new object();
+
         private readonly CPSessionManager _cpSessionManager;
         private readonly CPGroupId _groupId;
         private readonly string _fullName;
-        private int _destroyed, _disposed;
+        private int _destroyed;
         public const long InvalidFence = 0;
```
yuce, line 40: Rider reports that this constant may be made private.

src/Hazelcast.Net/CP/IFencedLock.cs
zpqrtbnk: won't change see other comment re. threading and java

src/Hazelcast.Net/CP/IFencedLock.cs
zpqrtbnk: won't change see other comment re. threading and java

src/Hazelcast.Net/CP/FencedLock.cs
zpqrtbnk: wont fix as: it makes no sense to add an explicit context parameter to the `IsLocked` method that does not use the value of the parameter + does *not* accept such a parameter in Java, and we cannot "fix" Java as Java and .NET have to be different here due to the difference in threading / context model. See comments in the code for details.

emreyigit (approved):
Looks great.

yuce (approved):

PR #621, opened 2022-05-12
Introduce failover options

zpqrtbnk:
Introduce failover options in preparation of blue/green implementation. In this current iteration, the options become available, but only the 1st cluster is used to start a totally normal client (so no failover of any sort is actually implemented).

A normal client is created as (this is unchanged):
```csharp
var options = new HazelcastOptionsBuilder().With(...).Build();
var client = await HazelcastClientFactory.StartNewClientAsync(options);
```

Whereas a failover client is created as (this is new):
```csharp
var options = new HazelcastFailoverOptionsBuilder().With().Build();
var client = await HazelcastClientFactory.StartNewFailoverClientAsync(options);
```

The two options builders share the very same API - and both parses the `hazelcast.json` JSON file - but one parses the top-level `hazelcast` element whereas the other parses the top-level `hazelcast-failover` element. Both therefore support the same environment variables or command-line arguments or any other mechanism that was supported originally but the normal options builder.


zpqrtbnk:
This PR has been incorporated in the #587 blue/green PR and will not be merged independently. Closing.

PR #623, opened 2022-05-18
Migrating .NET docs to Antora with re-outlining

Serdaro:


JakeSCahill:
The Antora structure looks good. I won't review the content now. It makes sense to review the content after we merge this so that we can get a staging site up.

We should add the GitHub Action to trigger builds of docs.hazelcast.com on a push (merge). See an example here: https://github.com/hazelcast/hz-docs/blob/main/.github/workflows/build-site.yml. You'll need to replace the build hook with your own: https://www.netlify.com/docs/webhooks

JakeSCahill (approved):

Serdaro:
Thanks Jake.
FYI, the content has not been improved; this is the current .NET content just rearranged into the new outline. I will be doing content enhancements on the way.

emreyigit (approved):

ihsandemir:
is there a link we can see the doc preview?

Serdaro:
> is there a link we can see the doc preview?

will let you know, not for now.

JakeSCahill (approved):
Looks good as a PoC 😄

PR #624, opened 2022-05-18
Fix to unsigned assembly reference

emreyigit:
Moq fails while working with signed assemblies. Now, removed the Moq usage and the reference.

### DIFF
--- a/src/Hazelcast.Net.Tests/Clustering/FailoverTests.cs
+++ b/src/Hazelcast.Net.Tests/Clustering/FailoverTests.cs
@@ -28,7 +28,7 @@ using Hazelcast.Configuration;
 namespace Hazelcast.Tests.Clustering
 {
     [Category("enterprise")]
-    [Timeout(120_000)]
+    [Timeout(150_000)]
     internal class FailoverTests : MultipleClusterRemoteTestBase
     {
         private IDisposable HConsoleForTest()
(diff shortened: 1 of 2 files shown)


zpqrtbnk:
I *think* Moq can support signed assemblies but we need to declare the `InternalsVisibleTo` attribute correctly, and I haven't. `AssemblyInfo;cs` currently contains:
```csharp
[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")]//test mocks
```
And that line should be updated as:
```csharp
// see https://github.com/Moq/moq4/wiki/Quickstart
[assembly:InternalsVisibleTo("DynamicProxyGenAssembly2,PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
```

See [Moq documentation](https://github.com/Moq/moq4/wiki/Quickstart) and [this StackOverflow question](https://stackoverflow.com/questions/30089042/typeloadexception-using-moq-on-internal-interface-in-signed-assembly) for references.

I'd rather we keep Moq as it can be useful and I am using it in Compact.

Also... This PR also seems to change other parameters such as timeouts, is that expected?

emreyigit:
I also prefer to keep Moq. The only difference are test timeouts for blue/green. Some tests fail only on GA. I assumed there is a timing issue. So, let's continue on #625 then, we can increase the timeouts if it happens again.

PR #626, opened 2022-05-30
Fix Failed Compatibility Test

emreyigit:
- The simple authentication in `FailoverTests` is dropped from integration tests with server. `TestClusterOptionsRotate` does still validate the credentials are set correctly as an unit test. Simple simple authentication is included in server `v5.0`, we cannot complete backward compatibility tests with it. (see https://github.com/hazelcast/hazelcast/pull/18948) 
- Removed `credentials.xml` which is redundant.
- Removed `HazelcastFailoverOptions.txt` which is empty and unused.
- Increased the timeouts on `FailoverTests`.

Note: Cherry picking is required for branch `release/5.1`.


zpqrtbnk (approved):
approved

alisengul53 (approved):

PR #628, opened 2022-05-31
Fix build

zpqrtbnk:
This PR ports the changes made to the `release/5.1.0` branch in order to fix its build which was broken by:
* An expired SSL certificate & a general confusion around test certs renewal - we now generate all our certs locall - that may not be the long-term solution but it works, now
* Some .NET type tests that failed for .NET Framework and signed code (PRs builds are not signed)
* Some failover tests that depended too much on timing and failed on GitHub

### DIFF
--- a/build/certs.ps1
+++ b/build/certs.ps1
@@ -532,13 +532,35 @@ function install-root-ca-windows ( $filename, $add = $true ) {
     }
 
     try {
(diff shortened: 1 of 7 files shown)


emreyigit (approved):
LGTM.

PR #633, opened 2022-06-03
Fix RetryStrategy delay

zpqrtbnk:
As exposed by the Cloud tests during release validation, we have a bug in our RetryStrategy where, if the timeout is set to -1 (i.e. infinite timeout), the delay is consistently zero. This fixes the issue + adds tests.

### DIFF
--- a/src/Hazelcast.Net/Clustering/RetryStrategy.cs
+++ b/src/Hazelcast.Net/Clustering/RetryStrategy.cs
@@ -15,6 +15,7 @@
 using System;
 using System.Threading;
 using System.Threading.Tasks;
+using Hazelcast.Configuration;
 using Hazelcast.Core;
 using Hazelcast.Exceptions;
 using Microsoft.Extensions.Logging;
@@ -29,11 +30,11 @@ namespace Hazelcast.Clustering
         private readonly string _action;
         private readonly ILogger _logger;
 
-        private int _initialBackoffMilliseconds { get; }
-        private int _maxBackoffMilliseconds { get; }
-        private double _multiplier { get; }
-        private long _timeoutMilliseconds { get; }
-        private double _jitter { get; }
+        private readonly int _initialBackoffMilliseconds;
+        private readonly int _maxBackoffMilliseconds;
+        private readonly double _multiplier;
+        private readonly long _timeoutMilliseconds;
+        private readonly double _jitter;
         private int _currentBackOffMilliseconds;
         private int _attempts;
         private DateTime _begin;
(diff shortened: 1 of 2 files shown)


src/Hazelcast.Net/Clustering/RetryStrategy.cs
```diff
@@ -71,19 +72,39 @@ public RetryStrategy(string action, int initialBackOffMilliseconds, int maxBackO
             _action = action.ToLowerInvariant();
 #pragma warning restore CA1308
 
-
+            if (initialBackOffMilliseconds < 0) throw new ConfigurationException("Initial back-off must be greater than or equal to zero.");
             _initialBackoffMilliseconds = initialBackOffMilliseconds;
+            if (maxBackOffMilliseconds < 0) throw new ConfigurationException("Maximum back-off must be greater than or equal to zero.");
             _maxBackoffMilliseconds = maxBackOffMilliseconds;
+            if (multiplier <= 0) throw new ConfigurationException("Multiplier must be greater than zero.");
```
alisengul53, line 79: Can multiplier be zero?

src/Hazelcast.Net/Clustering/RetryStrategy.cs
```diff
@@ -71,19 +72,39 @@ public RetryStrategy(string action, int initialBackOffMilliseconds, int maxBackO
...
 
             Restart();
         }
 
+        /// <summary>
+        /// (internal for tests only) Gets the delay.
+        /// </summary>
+        internal int GetDelay(int elapsed)
+        {
+            var delay = (int)(_currentBackOffMilliseconds * (1 - _jitter * (1 - RandomProvider.NextDouble())));
```
emreyigit, line 95: Correct me if I'm wrong please, the formula shouldn't be like 
``` csharp 
var delay = _currentBackOffMilliseconds  + (int)(_currentBackOffMilliseconds * (1 - _jitter * (1 - RandomProvider.NextDouble())));
```
I mean delay should be >= `_currentBackOffMilliseconds`. In this case, formula takes `_currentBackOffMilliseconds*n where  0<=n<=1` and less than `_currentBackOffMilliseconds`.

src/Hazelcast.Net/Clustering/RetryStrategy.cs
```diff
@@ -71,19 +72,39 @@ public RetryStrategy(string action, int initialBackOffMilliseconds, int maxBackO
...
+            if (_timeoutMilliseconds >= 0) delay = Math.Min(delay, Math.Max(0, (int)(_timeoutMilliseconds - elapsed)));
+            return delay;
+        }
+
+        /// <summary>
+        /// (internal for tests only) Gets the new back-off.
+        /// </summary>
```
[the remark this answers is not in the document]
zpqrtbnk: We are using it in `WaitAsync` indeed. If we were only using it there, it would be `private`. We are making it `internal` so we can also use it in tests. But it's `internal` only for tests, it should not be used otherwise.

emreyigit (approved):

PR #649, opened 2022-06-27
`ReadStringArray` Fixed and Tests Case Enlarged [API-1413]

emreyigit:
**Problem:**
`ReadStringArray` was only reading if the total size of array is larger than 16 bytes which was not work for the strings that are shorter than 16 bytes. 

**Solution:**
Size of a UTF8 character could be 1 to 4 bytes, so I changed the size checking to 1 byte. Also, added various test cases. There are test cases that covers 0,1,2,3,4 and 10 length of arrays. Also, string arrays will be produced with different size of strings.

```
String arrays will be like;
["x"]
["x", "xx"]
["x", "xx", "xxx"]
["x", "xx", "xxx", "xxxx"]
["x", "xx", "xxx", "xxxx", ..., "xx....xxx"]
```

### DIFF
--- a/src/Hazelcast.Net.Testing/TestUtils.cs
+++ b/src/Hazelcast.Net.Testing/TestUtils.cs
@@ -90,7 +90,8 @@ namespace Hazelcast.Testing
 
         public static string RandomStringWithLength(int length)
         {
-            return RandomString().Substring(0, length);
+            var str = RandomString();
+            return str.Substring(0, length < str.Length ? length : str.Length);
         }
     }
 }
--- a/src/Hazelcast.Net.Testing/TestUtils.cs
+++ b/src/Hazelcast.Net.Testing/TestUtils.cs
@@ -21,10 +21,15 @@ namespace Hazelcast.Testing
     {
         public static T[] RandomArray<T>(Func<T> randFunc, int size = 0)
         {
+            return RandomArray<T>(p => randFunc(), size);
+        }
+
+        public static T[] RandomArray<T>(Func<int, T> randFunc, int size = 0)
+        {
             var array = new T[size == 0 ? RandomProvider.Random.Next(5) + 1 : size];
             for (var i = 0; i < array.Length; i++)
             {
-                array[i] = randFunc();
+                array[i] = randFunc(i);
             }
             return array;
         }
@@ -36,7 +41,7 @@ namespace Hazelcast.Testing
 
         public static byte RandomByte()
         {
-            return (byte) RandomProvider.Random.Next();
+            return (byte)RandomProvider.Random.Next();
         }
 
         public static byte[] RandomBytes()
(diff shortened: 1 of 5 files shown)


src/Hazelcast.Net/Serialization/ObjectDataInput.api.cs
```diff
@@ -255,7 +255,8 @@ public string[] ReadStringArray()
             if (length == BytesExtensions.SizeOfNullArray) return null;
             if (length <= 0) return Array.Empty<string>();
 
-            CheckAvailable(Position, length * BytesExtensions.SizeOfDouble);
+            //A UTF8 string is at least 1 byte.
+            CheckAvailable(Position, length * BytesExtensions.SizeOfByte);
```
mdumandag, line 259: IMHO, this check is not very useful. A string will be encoded as 4 bytes length + UTF-8 payload bytes, so this check will not really make sure anything will work as expected. It might be the case that there are 3 bytes after this position, we will miss it here, but catch it in the `ReadString` calls below. 

I believe we can get rid of this check altogether, and rely on the check in the `ReadString`. This is also the way we do it in Java

src/Hazelcast.Net.Testing/TestUtils.cs
```diff
@@ -75,12 +80,17 @@ public static long RandomLong()
...
 
         public static string RandomString()
         {
             return Guid.NewGuid().ToString();
         }
+
+        public static string RandomStringWithLength(int length)
+        {
+            return RandomString().Substring(0, length);
+        }
```
mdumandag, line 94: What if the `length` > `length(RandomString())`?

src/Hazelcast.Net.Tests/Serialization/DataInputOutputTest.cs
```diff
@@ -23,20 +23,32 @@ namespace Hazelcast.Tests.Serialization
 {
     internal class DataInputOutputTest
     {
-        private static readonly Endianness[] Endiannesses =
+        private static readonly object[] Endiannesses =
```
mdumandag, line 26: This is not just `Endiannesses` anymore. Maybe we can rename the parameter

zpqrtbnk: +1

src/Hazelcast.Net/Serialization/ObjectDataInput.api.cs
zpqrtbnk: same, we can entirely remove the check

zpqrtbnk (commented):
commented

src/Hazelcast.Net/Serialization/ObjectDataInput.api.cs
emreyigit: Agreed, removed at https://github.com/hazelcast/hazelcast-csharp-client/pull/649/commits/ed9852d55214b086cba563196c6219bf27c47b71

zpqrtbnk (approved):
Now looks good to me.

PR #662, opened 2022-07-13
Documentation Improvements [API-1396] [API-1285]

emreyigit:
- Missing link for failover documentation is added to toc.
- Read file linked to Nuget.


srknzl (approved):

zpqrtbnk (approved):
Approved. Now the only way to be sure it works is to push to NuGet, and that will only happen next time we release. Can u make sure we have a note somewhere, to remember to verify that it worked?

Issue #661, opened 2022-07-12
System.ObjectDisposedException: The semaphore has been disposed [API-1453]


drieseng:
We regularly hit the following exception in **Hazelcast.Net** version **5.1.0**:
```txt
Hazelcast.Exceptions.ClientOfflineException: Hazelcast client is not connected. ---> Hazelcast.Exceptions.TargetDisconnectedException: Target disconnected. ---> System.ObjectDisposedException: The semaphore has been disposed.
   at System.Threading.SemaphoreSlim.CheckDispose()
   at System.Threading.SemaphoreSlim.Release(Int32 releaseCount)
   at Hazelcast.Core.HSemaphore.Release()
   at Hazelcast.Messaging.ClientMessageConnection.<SendAsync>d__20.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Hazelcast.Clustering.MemberConnection.<SendAsyncInternal>d__66.MoveNext()
   --- End of inner exception stack trace ---
   at Hazelcast.Clustering.MemberConnection.<SendAsyncInternal>d__66.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Hazelcast.Clustering.ClusterMessaging.<SendAsyncInternal>d__11.MoveNext()
   --- End of inner exception stack trace ---
   at Hazelcast.Clustering.ClusterState.ThrowIfNotActive(Exception innerException)
   at Hazelcast.Clustering.ClusterMessaging.<SendAsyncInternal>d__11.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Hazelcast.Clustering.ClusterMessaging.<SendAsyncInternal>d__10.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #669, opened 2022-07-22
Add test + fix semaphore release exception in ClientMesageConnection [API-1453]

zpqrtbnk:
See discussion in issue #661 "System.ObjectDisposedException: The semaphore has been disposed".

When a connection goes down while a message is being sent, we forgot to verify and released some semaphore that would have been disposed already. Added a test to illustrate this + added a fix.


emreyigit (approved):

src/Hazelcast.Net/Messaging/ClientMessageConnection.cs
```diff
@@ -296,7 +296,15 @@ public async ValueTask<bool> SendAsync(ClientMessage message, CancellationToken
             }
             finally
             {
-                _writer?.Release();
```
srknzl, line 299: What about adding a `disposed` boolean variable to HSemaphore? See https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose?view=net-6.0 for more.

zpqrtbnk: The `IHSemaphore` interface exists only to wrap a `SemaphoreSlim` instance in production, but replace it and mock it during tests. So we want it to behave exactly as `SemaphoreSlim` does. And, `SemaphoreSlim` throws if you release it once disposed. We try to keep that very same behavior.

srknzl (approved):

src/Hazelcast.Net/Messaging/ClientMessageConnection.cs
srknzl: Ok. Approved.

PR #678, opened 2022-08-02
Fix of Unstable Tests

emreyigit:
Improved;

- TestClientCannotFailoverToDifferentPartitionCount
- TestOperationsContinueWhenClientDisconnected(True,ReconnectAsync)
- TestClientCanFailoverFirstClusterNotUp(False,1)	                    
- TestClientCanFailoverFirstClusterNotUp(True,1)

Skipped due to non-reproducibility;
- ConnectAsyncCanceled
- ConnectMembersTests  -> increased timeouts
- TestListenersAfterClientDisconnected

Needs more work;
- CompletionSourceCompletesResultAsynchronously(True)


src/Hazelcast.Net.Tests/Clustering/ConnectMembersTests.cs
```diff
@@ -73,7 +73,7 @@ await foreach (var request in memberConnectionQueue.WithCancellation(cancellatio
                 Assert.That(addresses.Count, Is.EqualTo(2));
                 Assert.That(addresses, Does.Contain(NetworkAddress.Parse("127.0.0.1:1")));
                 Assert.That(addresses, Does.Contain(NetworkAddress.Parse("127.0.0.1:2")));
-            }, 2000, 200);
+            }, 3000, 200);
```
srknzl, line 76: I think this is the timeout. This is by default 60 seconds in Node.js I guess if you don't specify it. You can do something similar I think. If there is optional parameters in C#, you can set a relatively high defualt value to avoid unstable tests.

srknzl (dismissed):

src/Hazelcast.Net.Tests/Clustering/ConnectMembersTests.cs
emreyigit: It shouldn't take that time actually but time to time they work slow on Github.

emreyigit:
You can merge it as long as approvals are done. Don't wait for me. Let's get rid of red labels.

src/Hazelcast.Net.Tests/Core/SocketExtensionsTests.cs
```diff
@@ -78,7 +78,7 @@ public async Task ConnectAsyncConnectionRefused1()
         [Test]
         public async Task ConnectAsyncConnectionRefused2()
         {
-            var endpoint = IPEndPointEx.Parse("127.0.0.1:11000");
+            var endpoint = IPEndPointEx.Parse("127.0.0.1:11002");
```
zpqrtbnk, line 81: The socket listener is already going to randomize / offset the port - how is this fixing?

emreyigit: I didn't see that they are randomized in the socket listener. Now, I reverted these port values. In this case, the test is not touched. After all test fixes merged, we may need going over it one more time if there is need.

zpqrtbnk: ok

emreyigit:
Changes moved to https://github.com/hazelcast/hazelcast-csharp-client/pull/695, will continue on that PR.

PR #682, opened 2022-08-15
Build Warnings Removed

emreyigit:
Removed the all build warnings. There are few FIXME warnings but they require architectural changes.


srknzl (dismissed):

src/Hazelcast.Net.Tests/Networking/SslLayerTests.cs
```diff
@@ -58,7 +58,9 @@ public void ValidateCertificate()
             var loggerFactory = LoggerFactory.Create(builder => builder.AddStringBuilder(text));
 
             var sender = new object();
+#pragma warning disable SYSLIB0026
```
zpqrtbnk, line 61: add a comment: "X509Certificate empty constructor is obsolete BUT here we DO want to use it to create an empty cert"

src/Hazelcast.Net/Networking/CloudDiscovery.cs
```diff
@@ -68,8 +68,12 @@ internal static void SetResponse(string response)
 
             try
             {
-                // TODO: is this the best way to do an http request?
+                // TODO: Make it async and consider to use HttpClientFactory
+#pragma warning disable SYSLIB0014
```
zpqrtbnk, line 72: Ideally we want to have this code (maybe you can add it but comment it out):
```
string response;
using (var httpClient = new HttpClient())
using (var httpRequest = new HttpRequestMessage(HttpMethod.Get, _endpointUrl))
{
    httpClient.Timeout = TimeSpan.FromMinutes(_connectionTimeoutMilliseconds);
    httpRequest.Headers.AcceptCharset.ParseAdd("UTF-8");
    response = await httpClient.GetStringAsync(_endpointUrl).CfAwait();
}
```
problem as you have noted is that this forces us to make `CloudDiscovery.Scan()` async which eventually bubbles up to making some of `AddressProvider` methods async... so let's not do it now but maybe be more explicit in the TODO?

src/Hazelcast.Net/Models/IndexOptions.cs
```diff
@@ -87,10 +87,12 @@ public IndexOptions AddAttribute(string attribute)
         /// <returns>This instance for chaining.</returns>
         public IndexOptions AddAttributes(params string[] attributes)
         {
+#pragma warning disable CA1062 // Validate arguments of public methods
```
zpqrtbnk, line 90: this is a public method, better actually validate the arguments with
```
if (attributes is null) throw new ArgumentNullException(nameof(attributes));
```
and while we're at it, there should be the same line in the constructor that accepts an `IEnumerable<string>` - no idea why it does not trigger a warning.

src/Hazelcast.Net/Exceptions/ClientNotAllowedInClusterException.cs
```diff
@@ -74,7 +74,9 @@ public ClientNotAllowedInClusterException(string message, Exception innerExcepti
         /// about the exception being thrown.</param>
         /// <param name="context">The <see cref="StreamingContext"/> that contains contextual information
         /// about the source or destination.</param>
+#pragma warning disable CA2229 // Implement serialization constructors
```
zpqrtbnk, line 77: it's complaining that the ctor should not be `private` if the class is unsealed (so potentially can have classes inheriting from it, which would need to use the ctor, so it should be `protected`) - but we don't plan to have classes inherit from that one, so better mark the class `sealed` and then the warning goes away

src/Hazelcast.Net/Clustering/ClusterMembers.cs
```diff
@@ -315,7 +315,7 @@ public async Task RemoveConnectionAsync(MemberConnection connection)
                         // otherwise, we're really disconnecting: flip _connected, and change the state
                         _connected = false;
                         _logger.IfDebug()?.LogDebug("Removed connection {ConnectionId} to member {MemberId}, disconnecting.", connection.Id.ToShortString(), connection.MemberId.ToShortString());
-                        // FIXME ClusterChanged CANNOT go back to DISCONNECTED it WILL go to CONNECTED no matter what
+                        // ClusterChanged hold Connected state. It's just a transation.
```
zpqrtbnk, line 318: This FIXME exists because: the call to `ChangeState` switches the state to `Disconnected` and the state before that is expected to be either `Connected` or `ClusterChanged` BUT as far as I understand, `ClusterChanged` can only lead to `Connected` (if we connected successfully to the new cluster) or shutdown - I don't think there is a way to go from `ClusterChanged` to `Disconnected` - and therefore `ClusterChanged` should not be in the list of expected types here - so... either we understand and fix it, or it should remain a warning.

zpqrtbnk (changes requested):
Have commented.

src/Hazelcast.Net/CP/CPSubsystem.cs
```diff
@@ -163,8 +163,9 @@ public static (string groupName, string objectName, string fullName) ParseName(s
 
             if (groupName.Length == 0)
                 throw new ArgumentException("CP group name cannot be an empty string.", nameof(name));
-
+#pragma warning disable CA1847 //Use string.Contains(char) instead
```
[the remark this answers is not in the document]
emreyigit: There is a contradiction actually. If we use `char` then it will ask to to add comparison parameter but `string.Contains(char)` doesn't have overload for `string.Contains(char, StringComparision.XXXX)`

src/Hazelcast.Net/Models/IndexOptions.cs
emreyigit: I thought throwing could be breaking change since it's public but I'll add the check. ... I couldn't find anything meaninfull for the ctor case.

src/Hazelcast.Net/Clustering/ClusterMembers.cs
emreyigit: Removed `ClusterChanged` from expected states and kept the note as a warning at https://github.com/hazelcast/hazelcast-csharp-client/pull/682/commits/146cc487e6723800e0609781507b11b70e1c1c41

src/Hazelcast.Net/Exceptions/ClientNotAllowedInClusterException.cs
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/682/commits/146cc487e6723800e0609781507b11b70e1c1c41

src/Hazelcast.Net/Models/IndexOptions.cs
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/682/commits/146cc487e6723800e0609781507b11b70e1c1c41

src/Hazelcast.Net/Networking/CloudDiscovery.cs
emreyigit: Yeap, I didn't want to mix the removing warnings and the refactoring, so added a better explanation at https://github.com/hazelcast/hazelcast-csharp-client/pull/682/commits/146cc487e6723800e0609781507b11b70e1c1c41

src/Hazelcast.Net.Tests/Networking/SslLayerTests.cs
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/682/commits/146cc487e6723800e0609781507b11b70e1c1c41

src/Hazelcast.Net/AssemblySigning.cs
```diff
@@ -56,7 +56,9 @@ internal static string GetKeyToken(string key)
                 bytes[i] = byte.Parse(key.Substring(i * 2, 2), NumberStyles.AllowHexSpecifier);
 #pragma warning restore CA1305 // Specify IFormatProvider
 #pragma warning disable CA5350 // Do Not Use Weak Cryptographic Algorithms - well, that's what PublicKeyToken uses
+#pragma warning disable SYSLIB0021 //Derived cryptographic types are obsolete. Use the Create method on the base type instead.
```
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/682/commits/146cc487e6723800e0609781507b11b70e1c1c41

src/Hazelcast.Net/Networking/CloudDiscovery.cs
```diff
@@ -105,7 +109,7 @@ internal static void SetResponse(string response)
             {
                 _logger.LogWarning(e, "Hazelcast cloud discovery failed.");
             }
-            return null;
+            return null;
```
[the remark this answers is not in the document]
emreyigit: Yeah it wasn't nice. Replaced with an empty dictionary at https://github.com/hazelcast/hazelcast-csharp-client/pull/682/commits/146cc487e6723800e0609781507b11b70e1c1c41

src/Hazelcast.Net/Core/ExceptionExtensions.cs
```diff
@@ -49,23 +49,23 @@ public static TException SetCurrentStackTrace<TException>(this TException except
...
             var stackTraceString = new StackTrace(fNeedFileInfo: true).ToString();
-            var pos = stackTraceString.IndexOf('\n');
+            var pos = stackTraceString.IndexOf('\n', StringComparison.OrdinalIgnoreCase);
             if (pos > 0) stackTraceString = stackTraceString.Substring(pos + 1);
             SetStackTraceField(exception, stackTraceString);
 #endif
             return exception;
         }
-
+        #pragma warning disable CA1823
```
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/682/commits/146cc487e6723800e0609781507b11b70e1c1c41

src/Hazelcast.Net/Clustering/ClusterConnections.cs
```diff
@@ -590,7 +590,9 @@ private async Task ConnectFirstAsync(CancellationToken cancellationToken)
                             else
                             {
                                 isExceptionThrown = true;
-                                _logger.LogError(attempt.Exception, "Failed to connect to address {address}.", address.ToString());
+#pragma warning disable CA1727
```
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/682/commits/146cc487e6723800e0609781507b11b70e1c1c41

src/Hazelcast.Net/Exceptions/ClientNotAllowedInClusterException.cs
zpqrtbnk: ok

src/Hazelcast.Net/Models/IndexOptions.cs
zpqrtbnk: ok

src/Hazelcast.Net/Networking/CloudDiscovery.cs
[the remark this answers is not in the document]
zpqrtbnk: sorry my bad it's OK to return `null` not to add the whitespaces, can u fix?

src/Hazelcast.Net/Networking/CloudDiscovery.cs
zpqrtbnk: ok

src/Hazelcast.Net.Tests/Networking/SslLayerTests.cs
zpqrtbnk: ok

src/Hazelcast.Net/AssemblySigning.cs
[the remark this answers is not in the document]
zpqrtbnk: ok

src/Hazelcast.Net/Clustering/ClusterConnections.cs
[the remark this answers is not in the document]
zpqrtbnk: ok

src/Hazelcast.Net/Core/ExceptionExtensions.cs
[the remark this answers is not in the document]
zpqrtbnk: can u also check the 2 `using` statements I mention, else they'll be flagged by ReSharper?

src/Hazelcast.Net/Clustering/ClusterMembers.cs
zpqrtbnk: Ok the FIXME comment I added because I did not understand why `ClientState.ClusterChanged` appeared in the list of valid previous states, since `ClusterChanged` is not supposed to ever transition to `Disconnected`. The correct fix IMO is to remove the `ClusterChanged` value from the list *and* to remove the comment entirely - but I'd like to understand why `ClusterChanged` was there in the first place. Do you remember,

src/Hazelcast.Net/CP/CPSubsystem.cs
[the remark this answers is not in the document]
zpqrtbnk: my bad, what I'd do then, is use the `groupName.Contains('@', StringComparison.OrdinalIgnoreCase)` syntax here and then in `Polyfills/StringExtensions.cs` I'd add the following function in the `#if` block
```
public bool Contains(this string s, char c, StringComparison comparisonType)
    => s.Contains(new string(c), comparisonType);
```

src/Hazelcast.Net/Clustering/ClusterMembers.cs
emreyigit: No, I don't unfortunately. Guessing that was a leftover from the big failover refactoring?? I kept the comment as a warning but I'll remove it anyway since we have reached natural flow of the code. There is no exception to note it down.

src/Hazelcast.Net/CP/CPSubsystem.cs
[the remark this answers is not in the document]
emreyigit: Added the extension at https://github.com/hazelcast/hazelcast-csharp-client/pull/682/commits/fdd2f7002d520a65a8fb4a501d422bd923d02334

src/Hazelcast.Net/CP/CPSessionManager.Heartbeat.cs
```diff
@@ -96,8 +96,10 @@ private async Task RunAsync(CPGroupId groupId, CPSession sessionState, Cancellat
             }
             catch (Exception e)
             {
+#pragma warning disable CA1508
```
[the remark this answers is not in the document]
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/682/commits/fdd2f7002d520a65a8fb4a501d422bd923d02334

src/Hazelcast.Net/CP/CPSubsystem.cs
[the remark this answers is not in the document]
zpqrtbnk: ok

src/Hazelcast.Net/Core/ExceptionExtensions.cs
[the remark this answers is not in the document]
zpqrtbnk: ok

src/Hazelcast.Net/Clustering/ClusterMembers.cs
zpqrtbnk: I think you removed one of them but there's another FIXME exactly for the same thing somewhere else in the file, any chance you can clear it too?

As a side note... my original idea was that the method that change the state would throw if the current state is not one of the expected one, that's the reason for passing the "expected states" - it's not activated yet but I think we should activate it eventually. what do you think?

src/Hazelcast.Net/CP/CPSessionManager.Heartbeat.cs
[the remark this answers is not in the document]
zpqrtbnk: ok

src/Hazelcast.Net/Networking/CloudDiscovery.cs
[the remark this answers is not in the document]
zpqrtbnk: ok

src/Hazelcast.Net/Clustering/ClusterMembers.cs
zpqrtbnk: so around line 318 it's OK now but not around line 582?

emreyigit: Removed the second occurrence at https://github.com/hazelcast/hazelcast-csharp-client/pull/682/commits/d316b8b7c8052dd9006ed9fbafb1ba652c50bd06. I think we should discuss throwing exception in a separated issue. WDYT? So, we can experiment whether work as expected.

zpqrtbnk (approved):
Happy with changes, approved.

srknzl (approved):

PR #689, opened 2022-08-25
Prepare for Compact - Simplify Default vs Constant Serializers

zpqrtbnk:
For some reasons the set of built-in serializers is split between "constant" and "default" serializers though they practically all are constant serializers. This PR merges them all into one namespace and one definition so it simplifies the code in various places.


zpqrtbnk:
not going to merge from fork - only way to stack PRs is directly on the origin repo = this becomes #690

PR #690, opened 2022-08-25
[compact-3] Prepare for Compact - Simplify Default vs Constant Serializers

zpqrtbnk:
For some reasons the set of built-in serializers is split between "constant" and "default" serializers though they practically all are constant serializers. This PR merges them all into one namespace and one definition so it simplifies the code in various places.


zpqrtbnk:
killing the top-level `compact-3` branch as it has issues with GitHub, this PR becomes #696

PR #692, opened 2022-08-25
[compact-5] Upgrade to C# 9

zpqrtbnk:
In preparation for compact serialization, we upgrade our C# language version to version 9 (current latest version being 10). This will allow us to simplify how we handle generic constraints in some situations. This is purely a client code level thing and has no impact on the resulting code that we ship or on people using our client.


zpqrtbnk:
replaced by #698

PR #693, opened 2022-08-25
[compact-6] Fix critical issue in ObjectDataOutput

zpqrtbnk:
This fixes a very small but kinda critical issue in `ObjectDataOutput` that only revealed when implementing compact - where we would block-copy a buffer using a wrong length. 1-line fix.


zpqrtbnk:
replaced by #699

PR #695, opened 2022-08-27
Fix most-unstable tests

zpqrtbnk:
This PR goes in parallel with #678 and fixes tests that are unstable. We *need* to fix these because it's becoming more difficult every day to work on large PRs such as Compact, since builds keep failing for silly reasons.

* `FailoverTests` had a range of issues that caused them to be really unstable. They have been refactored.
* `ConnectMembersTests` could fail some `AssertEx.SucceedsEventually` assertions which had a 2s timeout and this really is GitHub being slow since this test is purely in-memory (not running against real members). Timeouts have been increased to 20s.
* `RegressionWithRealNetworkTest` had a range of issues that caused them to be unstable, they have been refactored.
* `CompletionSourceCompletesResultAsynchronously` had timing issues, removed all timing-dependent code and doing things differently.


emreyigit (approved):
Wondering that if there is a benefit reading alternative hazelcast configs via `TestFiles` instead as a resource file?

srknzl (approved):

PR #697, opened 2022-08-30
[compact 4] Compact - Schemas [API-971]

zpqrtbnk:
Introduce compact serialization schemas.


src/Hazelcast.Net/Serialization/FieldKindExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+            switch (kind)
+            {
+                case FieldKind.Boolean:
+                case FieldKind.Int8:
+                case FieldKind.Int16:
+                case FieldKind.Int32:
+                case FieldKind.Int64:
+                case FieldKind.Float32:
+                case FieldKind.Float64:
+                    return false;
```
emreyigit, line 56: How about decimal and similar stuff? Isn't HBigDecimal value type?

src/Hazelcast.Net/Serialization/Compact/CompactSerializationHook.cs
```diff
@@ -0,0 +1,53 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#nullable enable
+
+using System;
+
+namespace Hazelcast.Serialization.Compact
+{
```
emreyigit, line 20: I couldn't understand aim of the class.

src/Hazelcast.Net/Serialization/Compact/SchemaBuilder.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        /// <summary>
+        /// Adds a field to the schema.
+        /// </summary>
+        /// <param name="fieldName">The name of the field.</param>
+        /// <param name="kind">The field kind of the field.</param>
+        /// <returns></returns>
+        public SchemaBuilder WithField(string fieldName, FieldKind kind)
+        {
+            _fields.Add(new SchemaField(fieldName, kind));
```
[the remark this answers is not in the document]
emreyigit: I saw the test.

src/Hazelcast.Net/Serialization/Compact/CompactSerializationHook.cs
emreyigit: It looks like testing purposes but wouldn't be better if we move it to Hazelcast.Testing?

emreyigit:
Have commented. As a side note, I couldn't matched the tests with Java stated in the TDD.

src/Hazelcast.Net/Serialization/FieldKindExtensions.cs
zpqrtbnk: Good question, will add a comment to explain: a "reference" type here is something that is managed by reference by compact serialization. A pure `FieldKind.Int32` cannot be `null` and is a compact value type. A `FieldKind.NullableInt32` can be `null` and is a compact reference type. A `FieldKind.Decimal` (which corresponds to `HBigDecimal` which is indeed a .NET value type) *can* be `null` in compact and therefore always is a compact reference type.

I proposed we named it `FieldKind.NullableDecimal` to limit the confusion but that was rejected ;) Because we do *not* support non-nullable decimals.

src/Hazelcast.Net/Serialization/Compact/CompactSerializationHook.cs
zpqrtbnk: This class (the `CompactSerializationHook`) is referenced in `HazelcastClientFactory` method `CreateSerializationService` where it's added as a hook to the serialization configuration, with the following comment:
```
// add a hook that constructs and provides the compact IIdentifiedDataSerialization factory
// because some basic compact functionality such as schema distribution don't use compact
// serialization themselves.
```

zpqrtbnk:
Re. not matching tests: there are zero tests in these PRs, but one last *Compact Tests* PR that will contain all tests. It currently provides 100% test coverage *but* not by matching Java tests 1-to-1. I am going to implement that shortly.

src/Hazelcast.Net/Serialization/FieldKindExtensions.cs
emreyigit: I see, thanks for the explanation :)

src/Hazelcast.Net/Serialization/Compact/CompactSerializationHook.cs
emreyigit: Got it!, thanks.

emreyigit (approved):

zpqrtbnk:
merged manually and somehow GitHub does not detect it - closing

PR #700, opened 2022-08-30
[compact 7] Compact - Introduce options + some interfaces [API-971]

zpqrtbnk:
Introduce compact serialization options + some interfaces required for compact-serialization.

### DIFF
--- a/src/Hazelcast.Net/Serialization/Compact/CompactOptions.cs
+++ b/src/Hazelcast.Net/Serialization/Compact/CompactOptions.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) 2008-2021, Hazelcast, Inc. All Rights Reserved.
+﻿// Copyright (c) 2008-2022, Hazelcast, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -336,7 +336,7 @@ namespace Hazelcast.Serialization.Compact
             _typeName_isClusterSchema[schema.TypeName] = isClusterSchema;
         }
 
-        private bool GetIsClusterSchema(Type serializedType, string typeName)
+        private bool IsClusterSchema(Type serializedType, string typeName)
         {
             var isClusterSchema = false;
             if (_serializedType_isClusterSchema.TryGetValue(serializedType, out var ics1))
@@ -373,7 +373,7 @@ namespace Hazelcast.Serialization.Compact
             // process the serializers (all other registrations below will use the reflection serializer)
             foreach (var serializer in _typeName_serializer.Values)
             {
-                var isClusterSchema = GetIsClusterSchema(serializer.GetSerializedType(), serializer.TypeName);
+                var isClusterSchema = IsClusterSchema(serializer.GetSerializedType(), serializer.TypeName);
                 var withSchema = _typeName_schema.TryGetValue(serializer.TypeName, out var schema);
 
                 // since the serializer has been declared, there *has* to be at least one serialized type for the type name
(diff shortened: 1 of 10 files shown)


src/Hazelcast.Net/Serialization/Compact/ICompactReader.cs
```diff
@@ -0,0 +1,296 @@
...
+
+#nullable enable
+
+namespace Hazelcast.Serialization.Compact
+{
+    /// <summary>
+    /// Defines methods for reading fields from a compact-serialized blob.
+    /// </summary>
+    public interface ICompactReader
+    {
```
emreyigit, line 25: Is `ReadChar `discarded intentionally?

src/Hazelcast.Net/Serialization/Compact/ICompactSerializer.cs
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright (c) 2008-2021, Hazelcast, Inc. All Rights Reserved.
```
emreyigit, line 1: As a general note, we should re-update the headers after the merging.

src/Hazelcast.Net/Serialization/Compact/ICompactReader.cs
zpqrtbnk: "char" is not a valid compact `FieldKind`. It _is_ defined (`FieldKind.Char`) because `FieldKind` could also be used with Portable (somehow) but it is not supported by compact. There is no `readChar` method in Java's `CompactReader` interface.

src/Hazelcast.Net/Serialization/Compact/CompactOptions.cs
```diff
@@ -0,0 +1,462 @@
...
+        {
+            if (schema == null) throw new ArgumentNullException(nameof(schema));
+
+            EnsureUniqueSchemaPerTypeName(schema);
+
+            _typeName_schema[schema.TypeName] = schema;
+            _typeName_isClusterSchema[schema.TypeName] = isClusterSchema;
+        }
+
+        private bool GetIsClusterSchema(Type serializedType, string typeName)
```
emreyigit, line 339: I think the prefix `Get` on the method name is unnecessary but no harm.

zpqrtbnk: Makes sense, will rename

src/Hazelcast.Net/Serialization/Compact/CompactOptions.cs
```diff
@@ -0,0 +1,462 @@
...
+                isClusterSchema |= ics2;
+            return isClusterSchema;
+        }
+
+        /// <summary>
+        /// Creates and returns the serializer registrations corresponding to options.
+        /// </summary>
+        /// <returns>The serializer registrations corresponding to options.</returns>
+        /// <exception cref="ConfigurationException">The options contain conflicts that prevent the registrations from being created.</exception>
+        internal IEnumerable<CompactRegistration> GetRegistrations()
```
emreyigit, line 354: Just wondering, probably answer is in other PRs. There are other type serializers between compact and no-config compact. However, `GetRegistrations` handles them all together.  I think there will be another layer between serialization and registration.

emreyigit:
I've some small comments. The rest is LGTM.

src/Hazelcast.Net/Serialization/Compact/CompactOptions.cs
zpqrtbnk: `GetRegistration` in invoked when initializing the compact serialization service - then when we lookup for a serializer, we ask the compact serialization service whether it has a registration for the object's type. if not, we move on and try custom serializers, portable, etc. If, in the end, we still don't have a serializer, we'll ask the compact serialization service to please create a registration for that type and handle it from now on.

emreyigit (approved):

PR #702, opened 2022-08-30
[compact 9] Serialization - Introduce IRead/WriteObjectsFrom/ToObjectDataInput/Output [API-971]

zpqrtbnk:
Abstract the capability of reading & writing objects to `ObjectDataOutput` and from `ObjectDataInput` in an interface that can easily be mocked in tests.


emreyigit (dismissed):
LGTM

PR #703, opened 2022-08-30
[compact 10] Compact - Implement CompactReader and CompactWriter [API-971]

zpqrtbnk:
Implement the `CompactReader` and `CompactWriter`.


src/Hazelcast.Net/Serialization/Compact/CompactReaderWriterBase.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// <exception cref="SerializationException">No such field.</exception>
+        protected virtual SchemaField GetValidField(string name, FieldKind kind)
+        {
+            var field = GetValidField(name);
+            if (field.Kind == kind) return field;
+
+            throw new SerializationException($"Invalid kind \"{kind}\" for field \"{name}\" of schema {Schema}, which is of kind \"{field.Kind}\".");
+        }
+
+        protected (int, byte) GetBooleanFieldPosition(string name)
```
emreyigit, line 88: What does it make boolean special?

src/Hazelcast.Net/Serialization/Compact/CompactWriter.cs
```diff
@@ -0,0 +1,472 @@
...
+        private int _dataPosition;
+        private bool _completed;
+
+        public CompactWriter(IWriteObjectsToObjectDataOutput objectsWriter, ObjectDataOutput output, Schema schema)
+            : base(schema, output.Position)
+        {
+            _objectsWriter = objectsWriter ?? throw new ArgumentNullException(nameof(objectsWriter));
+            _output = output ?? throw new ArgumentNullException(nameof(output));
+
+            if (schema.HasReferenceFields) _offsets = new int[schema.ReferenceFieldCount];
```
emreyigit, line 43: I see it's used with null forgiving operator all over there but `_offsets` is only initialized if `HasReferenceFields`. Isn't there is room for null exception  🤕?

src/Hazelcast.Net/Serialization/Compact/CompactReaderWriterBase.cs
zpqrtbnk: Each integer, decimal, string, etc. is serialized as an independent field. OTOH non-nullable booleans are packed into bytes bit-fields. In other words, say you have 4 boolean fields, they will be serialized as 1 single byte, with each boolean corresponding to a bit in the byte. If you have 12 booleans you would have 2 bytes (8+4), etc. So booleans are special.

src/Hazelcast.Net/Serialization/Compact/CompactWriter.cs
zpqrtbnk: `_offsets` are only used if we have reference fields, so we should never use it if we don't have reference fields (i.e. when it's `null`). There is room for a nullref exception... if our code has bugs. We could test for `null` everywhere, but we'd just have to throw some sort of "internal error" exception... so I though, let's just keep the nullref.

emreyigit: > _offsets are only used if we have reference fields

That's fair enough.

src/Hazelcast.Net/Serialization/Compact/CompactWriterExtensions.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) 2008-2021, Hazelcast, Inc. All Rights Reserved.
```
emreyigit, line 1: Will it be filled in other PRs?

src/Hazelcast.Net/Serialization/Compact/CompactReader.cs
```diff
@@ -0,0 +1,577 @@
...
+        {
+            var value = ReadMaybeBoolean(name);
+            if (value.HasValue) return value.Value;
+            throw new SerializationException($"Null value for field \"{name}\" of schema {Schema}, which is of kind \"{FieldKind.NullableBoolean}\".");
+        }
+
+        public bool? ReadNullableBoolean(string name)
+            => ReadMaybeBoolean(name);
+
+        private T[]? ReadArrayOfBoolean<T>(SchemaField field, Func<bool, T> f)// where T : struct
```
emreyigit, line 305: I'm confused with generic and `boolean` here. Is it expected the `T` is a type of collection or what?

emreyigit: Oh, got it it's for nullable types.

emreyigit:
I'm okay with the PR, just have a small comment.

src/Hazelcast.Net/Serialization/Compact/CompactWriterExtensions.cs
zpqrtbnk: yes I'll do one general cleanup of all headers eventually - don't want to fix each PR one by one

emreyigit: Actually I meant the file not the header. The position of comment is misleading, sorry. The file is empty. I wondered the reason.

zpqrtbnk: oh? you're right. i created this file along with the `CompactReaderExtensions.cs` file but it never got anything into it.
removing it.

emreyigit (dismissed):

PR #704, opened 2022-08-30
[compact 11] Compact - Introduce serializer [API-971]

zpqrtbnk:
Introduce the compact serializer.

### DIFF
--- a/src/Hazelcast.Net.Tests/Serialization/Compact/CompactOptionsTests.cs
+++ b/src/Hazelcast.Net.Tests/Serialization/Compact/CompactOptionsTests.cs
@@ -339,16 +339,6 @@ namespace Hazelcast.Tests.Serialization.Compact
 
             options.SetSchema(SchemaBuilder.For("duh").Build(), true);
             Assert.Throws<ConfigurationException>(() => options.GetRegistrations().ToList());
-
-            options = new CompactOptions();
-            options.ReflectionSerializer = Mock.Of<ICompactSerializer<object>>();
-            options.SetSchema(SchemaBuilder.For("System.Object").Build(), true);
-            Assert.That(options.GetRegistrations().Count(), Is.EqualTo(1));
-
-            options = new CompactOptions();
-            options.ReflectionSerializer = Mock.Of<ICompactSerializer<object>>();
-            options.AddType<object>();
-            Assert.That(options.GetRegistrations().Count(), Is.EqualTo(1));
         }
     }
 }
--- a/src/Hazelcast.Net.Tests/Serialization/Compact/CompactOptionsTests.cs
+++ b/src/Hazelcast.Net.Tests/Serialization/Compact/CompactOptionsTests.cs
@@ -339,6 +339,20 @@ namespace Hazelcast.Tests.Serialization.Compact
 
             options.SetSchema(SchemaBuilder.For("duh").Build(), true);
             Assert.Throws<ConfigurationException>(() => options.GetRegistrations().ToList());
+
+            options = new CompactOptions
+            {
+                ReflectionSerializer = Mock.Of<ICompactSerializer<object>>()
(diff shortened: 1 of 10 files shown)


src/Hazelcast.Net/Serialization/Compact/SchemaBuilderWriter.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        private readonly string _typeName;
+        private readonly List<SchemaField> _fields = new();
+
+        public SchemaBuilderWriter(string typeName)
+        {
+            _typeName = typeName;
+        }
+
+        public FieldKind GetFieldKind(string name)
```
emreyigit, line 33: What's the goal here?

src/Hazelcast.Net/Serialization/Compact/CompactSerializationSerializer.cs
```diff
@@ -0,0 +1,375 @@
...
+            {
+                obj = null;
+                missingSchemaId = schemaId;
+                return false;
+            }
+
+            var registration = GetOrCreateRegistration(schema);
+
+            var reader = new CompactReader(this, input, schema, registration.SerializedType);
+            // FIXME - but then, each serializer MUST support a TryRead method AND return a state and oh my...
```
emreyigit, line 260: So, there is no `TryXXX()` approach in other clients as wwll :|

src/Hazelcast.Net.Tests/Serialization/Compact/CompactOptionsTests.cs
```diff
@@ -339,16 +339,6 @@ public void GetRegistrations()
 
             options.SetSchema(SchemaBuilder.For("duh").Build(), true);
             Assert.Throws<ConfigurationException>(() => options.GetRegistrations().ToList());
-
-            options = new CompactOptions();
```
emreyigit, line 343: Why it's removed?

emreyigit:
Have a few small questions, rest looks nice.

src/Hazelcast.Net/Serialization/Compact/SchemaBuilderWriter.cs
zpqrtbnk: `SchemaBuilderWriter` is a `ICompactWriter` but a special one: we pretend to serialize an object through it and it collects the fields and produces the corresponding schema. But it does not serialize anything really. So although it's a `ICompactWriter` you cannot use it as a true writer. Some methods are not supported. That one (`GetFieldKind`) cannot be supported since we don't know the schema beforehand. But... since it's defined on the interface, we have to implement it - and throw.

src/Hazelcast.Net/Serialization/Compact/CompactSerializationSerializer.cs
zpqrtbnk: True. And that was a leftover comment, thanks for catching it, removing it. And in fact the problem it mentioned is solved in a different way so it's totally not relevant anyways.

src/Hazelcast.Net.Tests/Serialization/Compact/CompactOptionsTests.cs
zpqrtbnk: It's just creating a new `CompactOptions` and not using it, probably a leftover, makes no sense so I removed it.

emreyigit: Yeah but coming lines are also removed and the assertions were looking promising.

zpqrtbnk: weird - have put them back.

emreyigit (dismissed):

PR #707, opened 2022-08-30
[compact 14] Compact - Wire Compact Serializer [API-971]

zpqrtbnk:
Wire the compact serializer - so now it's available for real!

### DIFF
--- a/src/Hazelcast.Net/Serialization/SerializationService.ReadWrite.cs
+++ b/src/Hazelcast.Net/Serialization/SerializationService.ReadWrite.cs
@@ -210,7 +210,20 @@ namespace Hazelcast.Serialization
         /// <exception cref="HazelcastException">Internal error.</exception>
         public async ValueTask<object> ToObjectAsync(object dataObj, ToObjectState state)
         {
-            throw new NotSupportedException("We don't have a compact serializer for now.");
+            var schemaId = state.SchemaId;
+            var safety = 1000;
+            while (safety-- > 0)
+            {
+                if (!await _compactSerializer.FetchSchema(schemaId).CfAwait())
+                    throw new UnknownCompactSchemaException(schemaId);
+
+                // and then we try again - we cannot do it on the fly during deserialization
+                // because that would force the entire deserialization (including ICompactReader)
+                // to turn asynchronous - and then we would have issues with lazy deserialization
+                if (TryToObject(dataObj, out var obj, out state)) return obj;
+                schemaId = state.SchemaId;
+            }
+            throw new HazelcastException(ExceptionMessages.InternalError);
         }
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


src/Hazelcast.Net/Serialization/SerializationService.ReadWrite.cs
```diff
@@ -210,7 +210,20 @@ public async ValueTask<T> ToObjectAsync<T>(object dataObj, ToObjectState state)
         /// <exception cref="HazelcastException">Internal error.</exception>
         public async ValueTask<object> ToObjectAsync(object dataObj, ToObjectState state)
         {
-            throw new NotSupportedException("We don't have a compact serializer for now.");
+            var schemaId = state.SchemaId;
+            var safety = 1000;
```
emreyigit, line 214: I think it's too much try. It looks like pooling to me. Why don't we fail more fast?

src/Hazelcast.Net/Serialization/SerializationService.cs
```diff
@@ -113,6 +116,12 @@ internal sealed partial class SerializationService : IDisposable, IReadObjectsFr
             _dataSerializerAdapter = CreateSerializerAdapter<IIdentifiedDataSerializable>(dataSerializer);
             RegisterConstantSerializer(_dataSerializerAdapter, typeof(IIdentifiedDataSerializable));
 
+            // Registers the constant 'compact serializer', which implements compact serialization.
+            // The two adapters are *not* registered, but handled as special cases by the Lookup methods.
```
emreyigit, line 120: Probably I'm missing something here. What is the other adapter here?

src/Hazelcast.Net/Serialization/SerializationService.ReadWrite.cs
zpqrtbnk: This is the idea: this method gets an object to deserialize + the ID of a schema that we know is missing. We fetch the schema (and throw if not found), then we try again to deserialize the object. This may fail if *another* schema is missing, in which case we loop. Eventually we'll have all schemas that we need. Except if there's some kind of a bug in our code, and we end up looping forever trying to somehow fetch the same missing schema over and over again. That should of course never happen. This safety thing here exists "just in case" to make sure the loop absolutely cannot become infinite. Let's say it's me not trusting me :) The number needs to be greater than the maximum number of schemas that we might need to fetch when deserializing. 1000 sounded big enough. I mean, 10 would be too small, we *could* need to fetch 10 schemas.

So now... if you look at our code globally, this could be replaced by `while (true)` but I am always scared of these things. Maybe I can add a comment to explain what I just wrote above?

src/Hazelcast.Net/Serialization/SerializationService.cs
zpqrtbnk: Ouch. That comment is a leftover from a time we had *two* compact serializers (when I thought we needed to support the mode where we embed the schemas along with the serialized payload). In fact client do not need to support that mode. So the comment is totally bogus, removing it.

emreyigit (approved):

PR #710, opened 2022-08-30
[compact 17] Compact Tests [API-971]

zpqrtbnk:
Bring in all the compact tests, so that the compact serialization namespace reaches 100% test coverage.


zpqrtbnk:
Re. the attribute: you mean that all compact-serialization tests that require a cluster, cannot work against pre-5.2 clusters, right? makes sense, will add the attribute.

Re. Java tests: although we have 100% coverage Ihsan has already pointed out that we should port all Java tests - that will be completed in another PR.

emreyigit:
Yes and great!

emreyigit (approved):

PR #711, opened 2022-08-30
[compact 18] Compact - Import documentation [API-971]

zpqrtbnk:
And finally, the very last PR for compact serialization initial implementation : documentation.


doc/dev/doc/serialization/json.md
```diff
@@ -0,0 +1,30 @@
...
+```csharp
+await map.PutAsync("item1", new HazelcastJsonValue("{ \"age\": 4 }"));
+await map.PutAsync("item2", new HazelcastJsonValue("{ \"age\": 20 }"));
+```
+
+You can query JSON objects in the cluster using the `Predicate`s of your choice. An example JSON query for querying the values whose age is less than 6 is shown below:
+
+```csharp
+    // Get the objects whose age is less than 6
+    var result = await map.GetValues(Predicates.IsLessThan("age", 6));
```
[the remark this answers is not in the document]
zpqrtbnk: you are right - though that is just a copy of some other documentation - will fix.

emreyigit (approved):

doc/dev/doc/serialization/compact.md
```diff
@@ -0,0 +1,77 @@
...
+
+It also supports zero-configuration use cases by automatically extracting schemas out of the classes and structs records using reflection, which is cached and reused later, with no extra cost.
+
+The underlying format of the compact serialized objects is platform and language independent. Native client supports will be added shortly after promoting this feature to stable status.
+
+## Zero Configuration
+
+Compact serialization can be used without registering a serializer for a type. Hazelcast will then try to extract a schema out of the class, using reflection, by inspecting all public properties. If successful, it registers the reflection-based serializer associated with the extracted schema and uses it while serializing and deserializing instances of that class. If the automatic schema extraction fails, Hazelcast throws an exception.
+
+Currently, most primitive types (`bool`, `int`...) are supported, as well as enums, arrays of those types, and nested classes. More advanced classes (e.g. `List<T>`) are not supported yet. Intefaces are not supported.
```
srknzl, line 28: @zpqrtbnk List, Set, Map interfaces in Java are supported. Why not support them in .net?

srknzl: Also,

Are datetime classes, bigdecimal and string primitive in .net? In  https://docs.hazelcast.com/hazelcast/5.1/serialization/compact-serialization#using-compact-serialization-with-zero-configuration, all fields are listed. For your sentence to include them they should be primitive.

doc/dev/doc/serialization/compact.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Compact Serializer
+
+Another way to use compact serialization is to implement the `ICompactSerializer<T>` interface for a type, and register it in the configuration. A basic serializer could look like:
+
+```csharp
+public class EmployeeSerializer : ICompactSerializer<Employee>
```
srknzl, line 35: @zpqrtbnk 

Don't we have `getClassName()` in .net client? Why?

doc/dev/doc/serialization/compact.md
```diff
@@ -0,0 +1,77 @@
...
+
+Hazelcast handles the versioning internally. So, you don’t have to change anything in the classes or serializers apart from the added, removed, or changed fields.
+
+Hazelcast achieves this by identifying each version of the class by a unique fingerprint. Any change in a class results in a different fingerprint. Hazelcast uses a 64-bit Rabin Fingerprint to assign identifiers to schemas, which has an extremely low collision rate.
+
+Different versions of the schema with different identifiers are replicated in the cluster and can be fetched by clients or members internally. That allows old readers to read fields of the classes they know when they try to read data serialized by a new writer. Similarly, new readers might read fields of the classes available in the data, when they try to read data serialized by an old writer.
+
+This means that for one *type name*, there can be several schemas.
+
+In addition, the `ICompactReader` interface exposes methods such as `FieldKind GetFieldKind(string name)` which returns the *kind* (i.e. the actual type) of the field.
```
srknzl, line 77: I don't know 5.2 documentation is up to date (I am comparing yours with version [5.1](https://docs.hazelcast.com/hazelcast/5.1/serialization/compact-serialization)) but maybe we can add a schema evolution usage here. In version [5.2](https://docs.hazelcast.com/hazelcast/5.2-snapshot/serialization/compact-serialization), there is an example.

@zpqrtbnk

doc/dev/doc/serialization/compact.md
zpqrtbnk: This was discussed with Metin. In .NET types are either open generics, or closed generics. For instance, `List<>` is an open generic and more or less corresponds to the Java type (where types are erased). And `List<int>` is a closed generic. If we have a `ICompactSerializer<Thing<Foo>>` then `serializer.GetType()` is going to return `Thing<Foo>` as a closed generic, the type is not erased, we can fully determine the serialized type - something you cannot do in Java. So... .NET does not require a `getClassName()` method because we can obtain the complete details from the type - such a method would be redundant and error-prone.

doc/dev/doc/serialization/compact.md
zpqrtbnk: `string` and `DateTime` is "primitive" as in, provided by .NET, wheras `BigDecimal` is provided by us. But they should all be treated as primitives. Do you mean that the `(bool, int...)` list should mention them all?

srknzl: No, I mean that we should list the data types which are not primitive, like `BigDecimal` because `primitive types` does not include them. If String, Date, Time, DateTime and Timestamp are primitive, then just listing `BIgdecimal` is enough

PR #713, opened 2022-08-31
Upgrade to C# 10 [API-1091]

emreyigit:
No warnings occurred with the upgrade.

### DIFF
--- a/src/Hazelcast.Net/AssemblySigning.cs
+++ b/src/Hazelcast.Net/AssemblySigning.cs
@@ -62,7 +62,9 @@ namespace Hazelcast
             var text = new StringBuilder();
             for (var i = 0; i < 8; i++)
                 //token[i] = hash[^(i + 1)];
+#pragma warning disable CA1305 // Specify IFormatProvider
                 text.Append($"{hash[^(i + 1)]:x2}");
+#pragma warning restore CA1305
             return text.ToString();
         }
     }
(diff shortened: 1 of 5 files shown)


src/Hazelcast.Net/Hazelcast.Net.csproj
```diff
@@ -5,6 +5,11 @@
     <RootNamespace>Hazelcast</RootNamespace>
   </PropertyGroup>
 
+  <PropertyGroup>
```
zpqrtbnk, line 8: The default language version for all our projects is specified in `src/Directory.Build.Props` so I think you want to update that file, not any of the `.csproj` files. Other than that... if the PR triggers no new warning then it's all good.

zpqrtbnk (changes requested):
See comment.

zpqrtbnk (approved):
Approved.

src/Hazelcast.Net/AssemblySigning.cs
```diff
@@ -62,7 +62,9 @@ internal static string GetKeyToken(string key)
             var text = new StringBuilder();
             for (var i = 0; i < 8; i++)
                 //token[i] = hash[^(i + 1)];
+#pragma warning disable CA1305 // Specify IFormatProvider
```
srknzl, line 65: Isn't this a warning?

srknzl (approved):
Approved but aren' there some warnings so you used pragma disables?

emreyigit:
@srknzl they can be neglectable ones. Also, those part are mostly in test helpers.

PR #717, opened 2022-09-08
Fix unstable tests

zpqrtbnk:
Fixing a test that was unstable. It was doing:
```
using var cancellationSource = new CancellationTokenSource(50);
await result.Take(5).ToListAsync(cancellationSource.Token);
```
And expecting an exception, but depending on timing, the `CancellationTokenSource` could cancel after the `ToListAsync` has completed. It's a bad idea to depend on timing. Replacing with:
```
using var cancellationSource = new CancellationTokenSource();
var count = 0;
await result.Take(5).Select(x =>
{
    if (count++ == 2) cancellationSource.Cancel();
    return x;
}).ToListAsync(cancellationSource.Token);
```
So essentially cancelling the `CancellationTokenSource` after 3 items have been returned, so we *know* it's going to be cancelled in a controlled way and before `ToListAsync` completes.

### DIFF
--- a/src/Hazelcast.Net.Tests/Sql/SqlQueryResultTests.cs
+++ b/src/Hazelcast.Net.Tests/Sql/SqlQueryResultTests.cs
@@ -114,8 +114,13 @@ namespace Hazelcast.Tests.Sql
 
             await AssertEx.ThrowsAsync<OperationCanceledException>(async () =>
             {
-                using var cancellationSource = new CancellationTokenSource(50);
-                await result.Take(5).ToListAsync(cancellationSource.Token);
+                using var cancellationSource = new CancellationTokenSource();
+                var count = 0;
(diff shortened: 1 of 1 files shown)


srknzl:
Not familiar with .net async stuff but seems reasonable. Does the CancellationToken only affect toListAsync? Let's say 50 ms passed but only 3 rows are iterated. Will `take(5)` be cancelled? Or the cancellation happen when `toListAsync()` is called?

srknzl (dismissed):

zpqrtbnk:
the cancellation token cancels to ToList, which is propagated to the Take and ultimately to the `result` enumeration

emreyigit (dismissed):

zpqrtbnk:
force-pushed rebase to master, been approved already, force-merging

PR #723, opened 2022-09-19
Implement Compact Updates [API-1532]

zpqrtbnk:
This PR implements changes as per API-1250 and API-1320 + updates the values in `FieldKind` to assign value zero to `NOT_AVAILABLE`. Note that most of these changes require that we change the default server version we test against to `5.0-SNAPSHOT` as `5.2` is n/a at the moment.

## FieldKind

Have updated all `FieldKind` enumeration values so that `FieldKind.NotAvailable` can be zero.

### 1. See if changing Treemap inside Schema to hashmap improves performance on java as well.

.NET already uses dictionaries, nothing to do.

### 2. Is fieldName needed in FieldDescriptor , we removed it in c++.

Abandoned, see Metin's comment on Jira.

### 3. Fix comments see same fixed in cpp:

These are wording changes in comments, we don't have the same comments, ignoring.

### 4. RabingFingerPrint methods with String can not have null. Java can be fixed.

Applied change with corresponding test.

### 5. We can add schema to error messages in compact_reader like here [...]

Abandoned, see Metin's comment on Jira. In addition: .NET *does* display the field kinds in error messages.

### 6. Add check for writing same field twice in default_compact_writer.

Not in default compact writer actually, but in the schema writer.
Applied change with corresponding test.

### 7. DefaultCompactWriter.writeBooleanBits can not have Nullable boolean array. 

Plus: Make it nonnull. Also remove NULL_ARRAY_LENGTH in readBooleanBitsandreadBooleanBitsAsNullables.

.NET already does it differently, ok.

### 8. rename getFieldDefinition to getFieldDescriptor on CompactInternalGenericRecord 

Not applicable, we don't have GenericRecord in .NET for now.

### 9. Change the order of serializer checks in serializerFor +  compact should not override builtinserializers. 

Discussed with Metin & during standup, we decided to *not* implement that option in .NET and to *not* allow overriding constant serializers. Have updated our code to throw if a compact serializer handles a constant type, with corresponding test.

### 1. GenericRecord.cloneWithBuilder will be renamed to newBuilderWithClone to have it similar to existing newBuilder method. 
   
We do not have generic records in .NET.

### 2. Introduce GenericRecord.get methods with default values similar to CompactReader to make the versioning case much easier to use. 
   
We do not have generic records in .NET.

### 3. Move GenericRecord/Builder/FieldKind to its own package com.hazelcast.nio.serialization.genericrecord - except, later it was decided not to move FieldKind
   
We do not have generic records in .NET. And the rest is OK already.

### 4. Move type_name from config to CompactSerializer. 

Already implemented in .NET.

### 5. For Zero config, expect the fields that we are going to serialize are public non final field or do have public getters and setters (both getter and setter should be available together). In other words, we serialize only fields that we can get/set publicly. 
   
Was discussed with Metin and Emre. The default practice in all .NET world (see for instance System.Text.Json) is to serialize public properties and to ignore fields. System.Text.Json recently got an option flag to enable dealing with fields, but it's not the common practice in .NET. In order to remain "idiomatic" we will keep serializing properties and not fields.

Dealing with fields in Java is because serialization in Java is more or less "state transfer" whereas in .NET it's "DTO serialization". Also, this does not prevent interoperability.

### 6. Don’t serialize jdk classes. 

Have added an exclusion filter in `ReflectionSerializer` which is going to exclude (a) all types within a `System.*` namespace and (b) all anonymous types. It's based on a list so we will be able to add more namespace in the future if we need to.

### 7. Document we don’t support different type/version of items in arrays at API javadoc.   Check if the type is same when writing on DefaultCompactWriter 
   
Have added a filter in `CompactWriter` to ensure that all members of the `T[]` array are of the exact type `T` (with minor twitches for nullable value types). And a comment so we remember that this is just to be aligned with Java but there is no real reason for this constraint.

### 8. Remove the CompactReader methods with default values and introduce `@Nonnull FieldKind getFieldKind()` which will return `FieldKind.NOT_AVAILABLE` and don’t throw any exception. 

Implemented in .NET. With tests.

### 9. Update API documentation and TDD: readIntxx method can read from NULLABLE_INTxx type field. readNullableIntxx method can read INTxx type field.

Already implemented & documented in .NET.

### 10. Follow decisions made here [...] Config and CompactSerializer is changed for java.
    
Discussed with Metin and in standup, `ICompactSerializer<T>` will *not* get a `getCompactClass` equivalent in .NET since the serialized type can always be derived from the serializer's type thanks to .NET type system being more complete than Java's.

Discussed, options will *not* get a `setSerializers` method as this is not .NET idiomatic (just add them all)


src/Hazelcast.Net/Serialization/Compact/CompactSerializationSerializer.cs
```diff
@@ -70,6 +71,13 @@ public CompactSerializationSerializer(CompactOptions options, ISchemas schemas,
 
         public bool HasRegistrationForType(Type type) => _registrationsByType.ContainsKey(type);
 
+        public bool TryGetSerializer(Type type, out ICompactSerializer? serializer)
+        {
+            var hasSerializer = _registrationsByType.TryGetValue(type, out var registration);
+            serializer = hasSerializer ? registration!.Serializer?.Serializer : null;
```
emreyigit, line 77: I'm confused here. So, there could be a serializer but null?

src/Hazelcast.Net/Serialization/Compact/ReflectionSerializer.cs
```diff
@@ -528,5 +534,26 @@ public virtual void Write(ICompactWriter writer, object obj)
...
             return r;
         }
+
+        private static void VerifyTypeIsSupported(object o)
+        {
+            // for FullName to be null, the type would need to be derived from an open generic somehow,
+            // which makes no sense since it is the actual type of a concrete object. we can assume that
+            // type is not going to be null.
+            var type = o.GetType();
+            var name = type.FullName!;
```
emreyigit, line 544: _Just trying to understand..._ Is there any possibility that the name could be  null?

emreyigit, line 544: I can't image even an open type name could be null 🤔

src/Hazelcast.Net/Serialization/SerializationService.Register.cs
```diff
@@ -20,6 +20,13 @@ internal partial class SerializationService
     {
         #region RegisterConstantSerializer
 
+        private static void EnsureValidConstantSerializer(ISerializerAdapter adapter)
+        {
+            if (adapter == null) throw new ArgumentNullException(nameof(adapter));
+            if (adapter.TypeId > 0) throw new ArgumentException("Constant serializer type-id must be <=0.", nameof(adapter));
+            if (-adapter.TypeId >= ConstantSerializersCount) throw new ArgumentException($"Constant serializer type-id must be >-{ConstantSerializersCount}.", nameof(adapter));
```
emreyigit, line 27: I didn't get the reason of the check.

emreyigit:
```suggestion
            if (-adapter.TypeId >= ConstantSerializersCount) throw new ArgumentException($"Constant serializer type-id must be <-{ConstantSerializersCount}.", nameof(adapter));
```

Since we compare the negative numbers :|

src/Hazelcast.Net/Serialization/SerializationService.cs
```diff
@@ -174,6 +179,19 @@ internal sealed partial class SerializationService : IDisposable, IReadObjectsFr
...
 
+        private void EnsureConstantNotOverridenWithCompact()
+        {
+            foreach (var error in _constantByType.Keys
+                         .Select(x => (oops: _compactSerializer.TryGetSerializer(x, out var y), type: x, serializer: y))
+                         .Where(x => x.oops))
+            {
+                var serializer = error.serializer == null ? "the reflection serializer" : $"serializer {error.serializer}";
+                throw new ConfigurationException($"Cannot register {serializer} for type {error.type} " +
+                                                 "because this type is a constant type and SerializationOptions.AllowOverrideDefaultSerializers " +
```
emreyigit, line 190: Where do you check `AllowOverrideDefaultSerializers` ?

src/Hazelcast.Net/Serialization/Compact/CompactSerializationSerializer.cs
zpqrtbnk: so... `_registrationsByType.TryGetValue(type, out var registration);`  is going to return `true` and set `registration` if there is a registration, otherwise return `false`. And then, the registration *may* have a serializer, if one has been specified, or none, if no serializer has been specified and we're going to use the reflection-based serializer. So:
```
serializer = hasSerializer  // as returned by the line above

  // we know registration is not null, but registration.Serializer may be null
  ? registration!.Serializer?.Serializer  

  // we don't have a registration, hence we don't have a serializer
  : null; 
```

That being said, the code being tricky, I am going to verify again that this is correct. Will update here.

src/Hazelcast.Net/Serialization/Compact/ReflectionSerializer.cs
zpqrtbnk: Point is, it *is* possible to override `GetType()` and return a class that derives from `Type` but provides a `null` `FullName`. I have no idea why this is possible nor why would anyone do it, but nevertheless it *is* possible and hence nullable checks raise a warning. I choose to ignore it with the `!`.

src/Hazelcast.Net/Serialization/SerializationService.cs
zpqrtbnk: Hold on, good catch, this is an error in the message - we don't have `AllowOverrideDefaultSerializers` in .NET I believe, Metin said don't put it, it's very Java-specific. Going to fix the message.

emreyigit:
Have commented.

srknzl:
> quire that we change the default server version we test against to 5.0-SNAPSHOT as 5.2 is n/a at the moment.

Do you mean 5.2-SNAPSHOT? :)

srknzl:
> .NET already uses dictionaries, nothing to do.

I can't comment on it because it is not changed in this PR. See the following code from your PR's head in [Schema.cs](https://github.com/hazelcast/hazelcast-csharp-client/blob/b6811b38609185c74cc46f6001385fe378998f99/src/Hazelcast.Net/Serialization/Compact/Schema.cs#L119)

You are using a SortedDictionary which must be a corresponding data structure to TreeMap in Java. Java uses different data structures. See [here](https://github.com/hazelcast/hazelcast/pull/21734/files#diff-4475b8a1fb2d8744f4f87b3d866fd7b70e19305e3c99f927fd47bfc5da253618R61). It uses a HashMap for the field mapping and stores the list that is passed to the Schema constructor. Then, it sorts the list in the `Schema.init()` method according to the field names. 

Can you change the logic to be the same? We want to use a HashMap because most of the time Schema's field map are read, it is written only once.

srknzl (changes requested):

src/Hazelcast.Net/Serialization/Compact/RabinFingerprint.cs
```diff
@@ -66,10 +72,13 @@ public static ulong Fingerprint(ulong fingerprint, int value)
         // fingerprints a string value
         public static ulong Fingerprint(ulong fingerprint, string value)
         {
-            if (value == null) return Fingerprint(fingerprint, 0);
+            if (value == null) throw new ArgumentNullException(nameof(value));
```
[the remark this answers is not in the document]
zpqrtbnk: This class ~is~ should be marked with a `#nullable enable` line, which is missing and I am going to add it, and then it means that a `string` parameter cannot be `null` (it would have to be `string?`) -- however, as a general rule of best-practice, we tend to keep all null checks in .NET because these "non-null" things are only compile-time verifications, and there are ways that at runtime shit happens. I know (discussed with Asim) that in many places in the Java code, tests have been removed because "things will be all right" -- we tend to keep them everywhere in .NET.

src/Hazelcast.Net/Serialization/FieldKind.cs
```diff
@@ -150,86 +150,86 @@ public enum FieldKind
...
 
         /// <summary>The array-of-compact-object primitive type.</summary>
         /// <para>This type is not supported by portable serialization.</para>
-        ArrayOfCompact = 29,
+        ArrayOfCompact = 30,
 
         /// <summary>The array-of-portable-object primitive type.</summary>
         /// <para>This type is not supported by compact serialization.</para>
-        ArrayOfPortable = 31
+        ArrayOfPortable = 32
```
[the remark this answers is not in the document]
zpqrtbnk: Totally happy removing these.

src/Hazelcast.Net/Serialization/SerializationService.cs
zpqrtbnk: have pushed an update that fixes this

src/Hazelcast.Net/Serialization/FieldKind.cs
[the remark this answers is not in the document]
zpqrtbnk: have pushed an update that fixes this

src/Hazelcast.Net/Serialization/Compact/RabinFingerprint.cs
```diff
@@ -42,7 +45,10 @@ static RabinFingerprint()
         // fingerprints an array of bytes
         public static ulong Fingerprint(ulong fingerprint, byte[] bytes)
         {
+            if (bytes == null) throw new ArgumentNullException(nameof(bytes));
```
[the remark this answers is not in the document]
zpqrtbnk: have pushed an update that fixes this

zpqrtbnk:
Re. the Schema class, concern is valid, no need to use a `SortedDictionary` for something that we sort only once, have pushed an update that fixes this

src/Hazelcast.Net/Serialization/Compact/Schema.cs
```diff
@@ -131,10 +127,16 @@ private void Initialize(string typeName, IEnumerable<SchemaField> typeFields)
             {
```
srknzl, line 127: See the comment above this line: `// build the fields map, which is sorted (see above)`. Is it correct? I don't think typeFields is sorted here.

src/Hazelcast.Net/Serialization/Compact/Schema.cs
```diff
@@ -131,10 +127,16 @@ private void Initialize(string typeName, IEnumerable<SchemaField> typeFields)
...
-                fieldsMap[field.FieldName] = field;
+                _fieldsMap[field.FieldName] = field;
             }
 
-            foreach (var field in fieldsMap.Values)
+            // the sorted set of fields, which will be used for fingerprinting - needs to be ordered
+            // exactly in the same way as Java, which uses Comparator.naturalOrder() i.e. "natural
+            // order", and good luck finding a definition for this, apart from it being case-sensitive,
+            // so we're going with whatever seems best in C# and hope it works.
+            Fields = _fieldsMap.Values.OrderBy(x => x.FieldName, StringComparer.InvariantCulture).ToArray();
```
srknzl, line 137: Can we reuse the typeFields variable here not to create another variable? 🤔 java stores whatever is passed to its constructor and then sorts it instead of creating a new.

zpqrtbnk: Problem is that `typeFields` is an un-ordered enumerable of fields, whereas `Fields` has to be ordered. In Java `fields` is a list of fields that we sort and capture, here `typeFields` is an `IEnumerable` which we cannot capture. Don't think it makes a big difference.

srknzl (approved):

emreyigit (approved):

PR #735, opened 2022-11-01
Workflow | report only origin on `Build&Test Branch`

emreyigit:
With recent updates on the workflow, forked repo tries to run `Build&Test Branch` each time and fail eventually. So, I also skipped these steps that keep failing not to get annoying notifications.


zpqrtbnk:
One strange thing: the `Ensure Repository` job is supposed to run only if `github.repository == 'hazelcast/hazelcast-csharp-client'` and then we have a chain of dependencies: `Build&Test Results` needs `Build&Test` which needs `Get Frameworks` which needs `Ensure Repository`. So in my mind... if `Ensure Repository` does not run, nothing runs.

Can u tell me which of these jobs run in your case?

emreyigit:
`Build&Test Result` and `Create Checks` fail each time whether `Ensure Repository` success or not. As we discussed, the condition in the step works only for that step. It doesn't effect the others. We may need somehow dependency relation with each other.

zpqrtbnk:
Responding to myself... at https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow#defining-prerequisite-jobs they explain that `needs` ensures that a job runs after another one and, by default, if that other job is successful. But as soon as an `if` is also added, that last condition (job being successful) is not working anymore. So in our case, the `if: always()` condition really_ means "always".

Now, we want to run these "always" _but_ only if on the proper repository. What you propose achieve this. Now, being pedantic, I though I'd rather want to depend on the "ensure repo" job to not duplicate the check. Which would give us the following code:

```
needs: [ "build-test", "ensure-repo" ]
if: always() && needs.ensure-repo.result == 'success'
```

Not 100% sure it works. What do you think? Is this making sense, or just complicating things?

emreyigit:
Makes sense to me. I'll update it accordingly. Just one thing, we put `always()` because action should be checked on every time. Otherwise, it's always true which makes it redundant, right?

.github/workflows/build-branch.yml
```diff
@@ -84,10 +84,8 @@ jobs:
 
   # create checks
   create-checks:
```
zpqrtbnk, line 86: ah but we want to keep the "always" for that one too :)

emreyigit: Even in the foreign repo?

emreyigit: Updated.

zpqrtbnk (approved):
Approve

PR #741, opened 2022-11-09
Enable .Net7 Build

emreyigit:
- Enabled the .Net 7 build. 
- Add option to show EOL target warnings. Since we intentionally support them, no need to keep it printing.
- Handled warnings about a TLS version on the tests.


zpqrtbnk (approved):
Approved

PR #742, opened 2022-11-14
LINQ Part 4 - Query formatter [API-1672]

emreyigit:
The PR contains code for query formatter which translates cleaned tree to string SQL statement.


src/Hazelcast.Net.Linq.Async/Visitors/QueryFormatter.cs
```diff
@@ -0,0 +1,501 @@
...
+    /// <summary>
+    /// Traverse and translate tree to SQL string statement.
+    /// </summary>
+    internal class QueryFormatter : HzExpressionVisitor
+    {
+        private StringBuilder _sb;
+        private List<object> _values;
+
+        // internal for test
+        internal QueryFormatter()
```
zpqrtbnk, line 36: since the class is `internal` you can probably make it `public`?

src/Hazelcast.Net.Linq.Async/Visitors/QueryFormatter.cs
```diff
@@ -0,0 +1,501 @@
...
+            f.Visit(expression);
+            return (f.ToString(), f._values);
+        }
+
+        public override string ToString()
+        {
+            return _sb.ToString();
+        }
+
+#pragma warning disable CS8765 // Nullability of type of parameter doesn't match overridden member (possibly because of nullability attributes).
```
zpqrtbnk, line 59: is this due to a conflict with the implemented interface not being nullable-enabled? like... if we implement nullable in our file, and the interface is not enabled, we have that type of conflicts?

zpqrtbnk (commented):
Small questions.

src/Hazelcast.Net.Linq.Async/Visitors/QueryFormatter.cs
zpqrtbnk: ok

src/Hazelcast.Net.Linq.Async/Visitors/QueryFormatter.cs
zpqrtbnk: ok

zpqrtbnk (approved):
Approved

PR #743, opened 2022-11-15
More stable tests coverage

zpqrtbnk:
This PR adds some tests and fixes a few things, in order to ensure a more stable tests coverage. Looking at [this Codecov report](https://app.codecov.io/gh/hazelcast/hazelcast-csharp-client/pull/731) for instance, it was clear at one point that the PR was reducing coverage because it did not run through some code paths, mostly due to timing issues. This is annoying. With these tests, we should *always* run through these code paths, and things should be better.

### DIFF
--- a/src/Hazelcast.Net/Core/Clock.cs
+++ b/src/Hazelcast.Net/Core/Clock.cs
@@ -84,7 +84,7 @@ namespace Hazelcast.Core
         /// </summary>
         /// <remarks>The epoch time in milliseconds.</remarks>
         public static long Milliseconds
-            => ToEpoch(DateTime.UtcNow);
+            => ToEpoch(Now);
 
         public static DateTime Now
             => _clockSource.Now;
--- a/src/Hazelcast.Net/Messaging/ClientMessageConnection.cs
+++ b/src/Hazelcast.Net/Messaging/ClientMessageConnection.cs
@@ -128,6 +128,12 @@ namespace Hazelcast.Messaging
                 }
             }
 
+            // NOTE
+            // bufferReference.Buffer is a 'readonly struct', so its content (i.e. the content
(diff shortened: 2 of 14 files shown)


srknzl:
can you merge master branch into your branch? (or rebase) Your branch is out of date as github shows

src/Hazelcast.Net/Core/Clock.cs
```diff
@@ -84,7 +84,7 @@ internal static void Reset()
         /// </summary>
         /// <remarks>The epoch time in milliseconds.</remarks>
         public static long Milliseconds
-            => ToEpoch(DateTime.UtcNow);
+            => ToEpoch(Now);
```
emreyigit, line 87: Just to be sure. I know we use it almost everywhere but how global is this?

src/Hazelcast.Net/Messaging/ClientMessageConnection.cs
```diff
@@ -128,9 +128,6 @@ private bool ReceiveMessageBytesAsync(SocketConnectionBase connection, IBufferRe
                 }
             }
 
-            // update the reference
-            bufferReference.Buffer = bytes;
```
emreyigit, line 132: What's reason to keep update the reference? I don't see any change.

src/Hazelcast.Net/Core/Clock.cs
zpqrtbnk: as long as the `Now` property returns `DateTime.UtcNow` it should not break anything

src/Hazelcast.Net/Messaging/ClientMessageConnection.cs
zpqrtbnk: guess what, I wondered the same, then realized that we *do* change it in fact, I am going to add a comment in the code to explain

emreyigit (approved):

PR #506, opened 2021-09-01
Guid serialization fix

alexb5dh:
Fixes https://github.com/hazelcast/hazelcast/issues/19371

PR #745, opened 2022-11-17
Guid polyglot error [API-1706]

zpqrtbnk:
> **NOTE:** depends on PR #747 which should be reviewed and merged first.

OK this is a pretty nasty bug.

Here is the thing: at one point [#hazelcast/19371](https://github.com/hazelcast/hazelcast/issues/19371) was reported: the `Guid` (`UUID`) values that we handled in .NET map keys and values had a different value in Java. And so it was fixed with #506. Without realizing (for lack of tests) that it was breaking the `Guid` values that we handled in .NET codecs. So, when we send a *codec* `Guid` to Java it receives the wrong `UUID`, and when Java sends a *codec* `UUID` to us, we receive the wrong `Guid`.

The thing is, codec `Guid` values are always serialized as little-endian, whereas data `Guid` values respect the endianness of the serializer. The way #506 fixed the original issue was not handling this carefully and ended up fixing one part, but breaking the other.

So, the first commit of this PR starts by introducing new tests that *do* fail and verify that we indeed have an issue. The test class also contains a long comment text that explains what is going on in details.

Next, I realized I had to cleanup the `Guid` serialization to make it more rational. This is what the *last* commit of the PR does. Now serialization is done in one unique place, in a clean way, and all tests are green.

In between, I had to distinguish between a codec `Guid` and a data `Guid`, because
* A codec `Guid` is encoded as 1 byte that indicates whether the value is the empty `Guid` or not, plus 16 bytes for the `Guid` itself but only when it's not the empty value.
* A data `Guid` is encoded as 16 bytes in all cases (no empty value management, it's just encoded as 16 bytes too)

Which means that `BytesExtensions.SizeOfGuid` should be 16 and we should introduce `BytesExtensions.SizeOfCodecGuid` which is 17. Which implies regenerating the codecs, etc.

All in all, in the end it's all clean now. Happy review. Sorry about that.


emreyigit:
I think this isn't the first time we have a problem about uuid, please see #506. So, it looks we keep reordering position of the bytes. I feel like there is another thing which causes the problem, like frameworks, guid versions or java versions etc. I'm also looking to it. Currently it's moving target to me. Please give me some time to understand whole. Anyway, it's a weird bug 🤔

zpqrtbnk:
Note that the new tests guarantee that this time we are using the correct byte order, since we are directly testing against Java. Also, all test pass in all framework versions. Now, I had forgotten about #506 and I agree that it is disturbing. Reviewing this other PR now.

emreyigit (dismissed):
Thanks. It looks rigid.

src/Hazelcast.Net/Protocol/BuiltInCodecs/EntryListIntegerGuidCodec.cs
```diff
@@ -21,7 +21,7 @@ namespace Hazelcast.Protocol.BuiltInCodecs
 {
     internal static class EntryListIntegerUUIDCodec
     {
-        private const int EntrySizeInBytes = BytesExtensions.SizeOfInt + BytesExtensions.SizeOfGuid;
+        private const int EntrySizeInBytes = BytesExtensions.SizeOfInt + BytesExtensions.SizeOfCodecGuid;
```
emreyigit, line 24: Does this change apply to codec template? Or should we check codecs each time after regenerated?

zpqrtbnk: The codec templates have been updated. So there should be a protocol repo PR too. You'll need to have a look at [this PR](https://github.com/hazelcast/hazelcast-client-protocol/pull/442) and ideally review/approve too.

emreyigit (approved):

emreyigit (approved):

PR #749, opened 2022-11-30
Fix alias and map name confusion

emreyigit:
It fixes the failed linq tests.


emreyigit:
It's strange tests run on somewhere doesn't belong to PR. I'll re create the PR.

PR #751, opened 2022-12-01
Fix failed linq tests 

emreyigit:
Copy of #750


emreyigit:
No need anymore, problem detected.

PR #752, opened 2022-12-02
LINQ 5 - Fetch and Construct Results [API-1673]

emreyigit:
This is the last PR of Linq series with current scope. The PR enables end user to use supported LINQ functions on IHMap based on SQL service.

- `IAsyncQueryable` is implemented. 
- Querable type is `HKeyValuePair`. It does same thing with `KeyValuePair` but more flexable.
- `AsAsyncQueryable()` extension is introduced to provide LINQ facilities from a `Hazelcast.Net.Linq.Async` package. 
- The result object is `IAsyncEnumerable`. It can be consumed with `await foreach`. No support for `ToXXXAsync()` transformations.
- Key or Value type can be both complex or primitive. Supported anyway.
- No override or configuration for property name casing currently. Public property names will be accepted as same with column names.
- Current LINQ operation are `Select` and `Where`. Only supported operator for string is `==`.
- No caching on compiled queries currently.

LINQ provider only interest with column names which are created while creating a mapping for the HMap. De/serialization and SQL part are done by the client. The provider uses `ISqlService` underneath. 

### Example usage;

**Sample object:**
```csharp
public class Person
{
    public int Id { get; set; }
    public string Name { get; set; }
    public string LastName { get; set; }
}
```
**Creating the mapping:**
```csharp
await Client.Sql.ExecuteCommandAsync(
                "CREATE MAPPING \"personMap\" (Id int, Name varchar, LastName varchar) " +
                "TYPE IMap " +
                "OPTIONS ('keyFormat' = 'int'," +
                "'valueFormat' = 'compact'," +
                "'valueCompactTypeName' = 'person')");
```
_Note: Column and property names should match. Column names are case sensitive on server side._

Querying with LINQ _(Assume there is data in it.)_:
```csharp
var _map = await Client.GetMapAsync<int, Person>("personMap");
var result = _map.AsAsyncQueryable().Where(p => p.Value.Name == "James Brown");
await foreach (var entry in result)
   // Here entry is a `HKeyValuePair<int, Person>`
   // Do your stuff...
```
The query will be executed and fetched from server when foreach invokes enumerator of `result ` with `MoveNextAsync()`. Rest will be on enumerator returned from `ISqlService`.

_Note: Documentation and examples will be on another PR._


Hazelcast.Net.sln.DotSettings
```diff
@@ -1,4 +1,5 @@
 ﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+	<s:Boolean x:Key="/Default/CodeEditing/SuppressUninitializedWarningFix/Enabled/@EntryValue">False</s:Boolean>
```
zpqrtbnk, line 2: What is that one doing?

src/Hazelcast.Net.Linq.Async/Expressions/ColumnExpression.cs
```diff
@@ -40,13 +40,19 @@ internal class ColumnExpression : Expression
         /// <inheritdoc />
         public override Type Type { get; }
 
-        public ColumnExpression(Type type, string alias, string name, int ordinal)
+        /// <summary>
+        /// It holds value whether Column is part of the Key part of HMap.
```
zpqrtbnk, line 44:
```suggestion
        /// Whether Column is part of the Key part of HMap.
```

src/Hazelcast.Net.Linq.Async/HKeyValuePair.cs
```diff
@@ -0,0 +1,39 @@
...
+// limitations under the License.
+
+namespace Hazelcast.Linq
+{
+    /// <summary>
+    /// A key value pair for <see cref="IHMap"/> entries on LINQ.
+    /// </summary>
+    /// <typeparam name="TKey">Type of Key</typeparam>
+    /// <typeparam name="TValue">Type of Value</typeparam>
+    public struct HKeyValuePair<TKey, TValue>
```
zpqrtbnk, line 22: Why do we need this alternative to `KeyValuePair`? Also, it's a non-readonly `struct`. Feels dangerous but there has to be a reason which we should at least document in comments.

src/Hazelcast.Net.Linq.Async/HMapExtension.cs
```diff
@@ -0,0 +1,36 @@
...
+    public static class HMapExtension
+    {
+        /// <summary>
+        /// Extension of <see cref="IHMap"/> for LINQ functionalities. It provides IAsyncQueryable part of the <see cref="IHMap"/>.  
+        /// </summary>
+        /// <param name="hMap"><see cref="IHMap"/> to be queried.</param>
+        /// <typeparam name="TKey">Type of key of the <see cref="IHMap"/></typeparam>
+        /// <typeparam name="TValue">Type of value of the <see cref="IHMap"/></typeparam>
+        /// <returns></returns>
+        public static IAsyncQueryable<HKeyValuePair<TKey, TValue>> AsAsyncQueryable<TKey, TValue>(this IHMap<TKey, TValue> hMap)
```
zpqrtbnk, line 30: So we've probably discussed this but I forgot - why cannot `IHMap` directly implement the interface and why do ppl have to do the `AsAsyncQueryable` thing? Can we document the reason?

src/Hazelcast.Net.Linq.Async/ObjectReader.cs
```diff
@@ -0,0 +1,76 @@
...
+    internal class ObjectReader<TResult> : IAsyncEnumerator<TResult>
+    {
+        private readonly QueryProvider _queryProvider;
+        private readonly Expression _expression;
+        private IAsyncEnumerator<SqlRow>? _sqlEnumerator;
+        private ISqlQueryResult? _queryResult;
+        private Func<SqlRow, TResult>? _projector;
+        private readonly CancellationToken _cancellationToken;
+
+#pragma warning disable 8618 Non-nullable variable must contain a non-null value when exiting constructor. Consider declaring it as nullable.
```
zpqrtbnk, line 35: Mixed feeling about the warning. Cannot we fix the original issue?

src/Hazelcast.Net.Linq.Async/Visitors/QueryFormatter.cs
```diff
@@ -44,18 +44,18 @@ internal QueryFormatter()
         /// </summary>
         /// <param name="expression">tree</param>
         /// <returns>(SQL statement, Value of variables)</returns>
-        public static (string, IEnumerable<object>) Format(Expression expression)
+        public static (string, object[]) Format(Expression expression)
         {
             var f = new QueryFormatter();
             f.Visit(expression);
-            return (f.ToString(), f._values);
+            return (f.ToString(), f._values.ToArray());
```
zpqrtbnk, line 51: It's usually more efficient to `ToList`, do we need `ToArray`?

src/Hazelcast.Net.Linq.Async/Visitors/QueryFormatter.cs
```diff
@@ -376,6 +378,23 @@ protected virtual Expression VisitPredicate(Expression predicate)
             return predicate;
         }
 
+        protected override Expression VisitMember(MemberExpression node)
+        {
+            /*if (node.Expression != null && node.Expression.Type.IsGenericType && node.Expression.Type.GetGenericTypeDefinition() == typeof(HKeyValuePair<,>))
```
zpqrtbnk, line 383: ?

src/Hazelcast.Net.Testing/Remote/Resources/hazelcast.xml
```diff
@@ -58,5 +58,5 @@
     <capacity>10</capacity>
     <time-to-live-seconds>180</time-to-live-seconds>
   </ringbuffer>
-
+  <jet enabled="true"/>
```
zpqrtbnk, line 61: this is going to enable JET for all our tests by default, which is a good thing - but then there's a special other config file with JET enabled in the tests project that we don't need anymore... think you can find it and remove it?

src/Hazelcast.Net/DistributedObjects/DistributedObjectBase.cs
```diff
@@ -128,7 +128,7 @@ public Action<DistributedObjectBase> ObjectDisposed
         /// <summary>
         /// Gets the logger factory.
         /// </summary>
-        protected ILoggerFactory LoggerFactory { get; }
+        internal ILoggerFactory LoggerFactory { get; }
```
zpqrtbnk, line 131: Making this `internal` visible, along with the `SqlService` in `HMap`, feels like it's breaking encapsulation... I am not too sure... since it's all internal we can merge it but I'd like we think about it.

zpqrtbnk (commented):
Have commented.

Hazelcast.Net.sln.DotSettings
emreyigit: It should added by Rider or Resharper. I'll check it again.

src/Hazelcast.Net.Linq.Async/HMapExtension.cs
emreyigit: Extension method provides Linq abilities from an another package. Otherwise, ÌHMap` should implement the query provider interfaces, so we cannot separate Linq into a different package.

src/Hazelcast.Net.Linq.Async/ObjectReader.cs
emreyigit: Warning is  for `Current` which is not initialized at ctor but `ObjectReader` is used as enumerator. So, that wouldn't be problem as long as result of `MoveNextAsync()` is obeyed.

src/Hazelcast.Net.Linq.Async/Visitors/QueryFormatter.cs
emreyigit: `_values` is already a List but I need a copy of the params, otherwise `SqlService` tries to serialize the list as whole instead of its params.

src/Hazelcast.Net.Linq.Async/Visitors/QueryFormatter.cs
emreyigit: Leftover from an experiment, removed.

src/Hazelcast.Net.Testing/Remote/Resources/hazelcast.xml
emreyigit: Done.

src/Hazelcast.Net.Linq.Async/HMapExtension.cs
zpqrtbnk: totally make sense. maybe at the top of the class, so we remember:
```
// async LINQ is provided as an extension package - so IHMap cannot directly expose async LINQ - so we have to
// go with an extension method to map it to something async-queryable
```

src/Hazelcast.Net.Linq.Async/ObjectReader.cs
zpqrtbnk: `Current = null!`

src/Hazelcast.Net/DistributedObjects/DistributedObjectBase.cs
zpqrtbnk: we will think about it later

src/Hazelcast.Net.Linq.Async/ObjectReader.cs
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/752/commits/4506e1ba5fd0bae9f9c763cf789a39e0f652006f

src/Hazelcast.Net.Linq.Async/HMapExtension.cs
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/752/commits/4506e1ba5fd0bae9f9c763cf789a39e0f652006f

zpqrtbnk (approved):
Approved!

PR #756, opened 2022-12-08
Compact collections support [API-1617]

zpqrtbnk:
> NOTE: depends on #755 

This PR adds compact serialization support for collections such as lists, sets and dictionaries.

### DIFF
--- a/src/Hazelcast.Net.Tests/Serialization/Compact/ReflectionSerializerTests.cs
+++ b/src/Hazelcast.Net.Tests/Serialization/Compact/ReflectionSerializerTests.cs
@@ -524,6 +524,124 @@ namespace Hazelcast.Tests.Serialization.Compact
             });
         }
 
+        private static readonly object[] ExceptionObjects = 
+        {
+            PoisonClass1.CreateInstance("duh"),
+            PoisonClass2.CreateInstance("duh")
+        };
+
+        [TestCaseSource(nameof(ExceptionObjects))]
+        public void ReflectionSerializerMoreExceptions(object obj)
+        {
+            var serializer = new ReflectionSerializer();
+            var sw = new SchemaBuilderWriter("thing");
+            serializer.Write(sw, obj);
+            var schema = sw.Build();
+
+            var orw = new ObjectReaderWriter(serializer);
+
+            var output = new ObjectDataOutput(1024, orw, Endianness.LittleEndian);
+            var writer = new CompactWriter(orw, output, schema);
+            serializer.Write(writer, obj);
+            writer.Complete();
+
+            var buffer = output.ToByteArray();
+
+            var input = new ObjectDataInput(buffer, orw, Endianness.LittleEndian);
+            var reader = new CompactReader(orw, input, schema, obj.GetType());
+
+            Assert.Throws<SerializationException>(() => serializer.Read(reader));
+        }
+
+        [Test]
+        public void ReflectionSerializerInvalidDictionaryData()
+        {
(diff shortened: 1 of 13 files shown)


src/Hazelcast.Net.Tests/Serialization/Compact/ReflectionSerializerTests.cs
```diff
@@ -81,6 +81,40 @@ public void SchemaBuilderWriterExceptions()
...
+
+        private static Dictionary<int, string> GetDictionary()
+        {
+            var dict = new Dictionary<int, string>();
+            for (var i = 0; i < 10; i++) dict[i] = i.ToString();
+            return dict;
+        }
+
+        [TestCaseSource(nameof(SerializeExceptionSource))]
+        public void SerializeOneExceptions((Type PropertyType, object? PropertyValue) testCase)
```
emreyigit, line 105: What do you test here? Do you test `ReflectionSerializer` throws if the property defined/added to type at runtime?

src/Hazelcast.Net/Serialization/Compact/ReflectionSerializer.cs
```diff
@@ -24,527 +25,634 @@
...
+                value = Enum.Parse(enumType, s);
+                return true;
             }
 
-            return Writers.TryGetValue(type, out var write)
-                ? write
-                : (writer, name, obj) => WriteUnknownType(writer, name, type, obj);
+            Verify.Condition<SerializationException>(isNullableEnum, "Read null value for non-nullable enum.");
+            value = null;
+            return true;
```
emreyigit, line 300: In which case `TryReadCustomEnum` fails to read and return `false`?

src/Hazelcast.Net/Serialization/Compact/ReflectionSerializer.cs
```diff
@@ -24,527 +25,634 @@
...
+            GetWriter(property.PropertyType)(writer, fieldName, property.GetValue(obj));
+        }
+    }
+
+    private static bool GetValidFieldName(ICompactWriter iwriter, Type propertyType, string propertyName, [NotNullWhen(true)] out string? fieldName)
+    {
+        // writer here can be a true CompactWriter, which holds a schema, and is able to provide
+        // 'valid' field names (mostly mapping different casing in names e.g. 'Name' to 'name'),
+        // but it also can be a SchemaBuilderWriter in which case we are writing the schema, and
+        // have to accept the property names as field names.
```
emreyigit, line 590: I assume field names are counterparts of C# properties on compact and Java side, right? It doesn't relate C# fields.

zpqrtbnk: if you write your own serializer then the field name is whatever you specify when you do `WriteInt32("fieldname", value)` - OTOH if you rely on the reflection serializer things start to get different. if you do not provide a schema, we'll create one, and then if the C# property is named `Value` the field will be named `Value`. if you *do* provide a schema we'll try to match a C# property named `Value` with a field named `Value` *or* `value` - it's best effort really.

as soon as you want true reliable interop with Java you should probably write your own serializer anyways.

src/Hazelcast.Net.Tests/Serialization/Compact/ReflectionSerializerTests.cs
zpqrtbnk: testing that some invalid values for collections actually trigger an exception

emreyigit (approved):

PR #757, opened 2022-12-08
Compact schema distrib [API-1617]

zpqrtbnk:
> NOTE: depends on #756 

This PR refactors how schemas are replicated as per the changes in Java.

### DIFF
--- a/src/Hazelcast.Net.Tests/Networking/NetworkingTests.cs
+++ b/src/Hazelcast.Net.Tests/Networking/NetworkingTests.cs
@@ -338,12 +338,15 @@ namespace Hazelcast.Tests.Networking
         }
 
         [Test]
-        [Timeout(10_000)]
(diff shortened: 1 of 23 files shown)


emreyigit (approved):
Although it looks nice to me, if QA could also match the behaviors with Java side, it would be better.

PS: Could you also check the approximate coverage of public APIs of compact as a whole feature? If you can added it as a comment, it would be great before the release.

srknzl:
> Although it looks nice to me, if QA could also match the behaviors with Java side, it would be better.

@emreyigit As far as I understand, this is the schema distribution change like returning replicated member uuids to clients etc. 
 So, this is an internal behavior change that is hard to verify without tests or observing the whole code. I can't just run a code sample and verify it works. Therefore, a QA cannot check this IMO. Just to let you know.

PR #760, opened 2022-12-14
Codecs version

zpqrtbnk:
Update submodule to valid latest commit + regenerate all codecs.


emreyigit:
While checking the public API  coverage last week, I realize that codecs decreases test coverage significantly. Can we exclude not used ones from coverage or disabled at [__init__.py](https://github.com/hazelcast/hazelcast-client-protocol/blob/master/cs/__init__.py) ?

emreyigit (approved):

PR #769, opened 2023-01-03
Fix various issues that have appeared in the master branch

zpqrtbnk:
... and were discovered during release testing ... nothing serious but breaks tests etc.


src/Hazelcast.Net.Tests/Sql/SqlJetTests.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) 2008-2022, Hazelcast, Inc. All Rights Reserved.
+﻿// Copyright (c) 2008-2021, Hazelcast, Inc. All Rights Reserved.
```
emreyigit, line 1: Where does 2021 come from?

emreyigit:
While fixing things, could you also look that condition in the work flow. Since Github doesn't have `array.length`,  `inputs.fwks[inputs.fwks.length-1]` doesn't work.

https://github.com/hazelcast/hazelcast-csharp-client/blob/b1b9367fbbe8686757247dfb8e9aa97bb9637a1c/.github/workflows/build-and-test.yml#L185

src/Hazelcast.Net.Tests/Sql/SqlJetTests.cs
zpqrtbnk: a mystery to me, fixing.

zpqrtbnk: fixed in latest commit, think I checked-out some files from old revisions to revert some changes, and picked the year by mistake, fixed in many files now.

emreyigit (approved):

PR #771, opened 2023-01-03
Rewire Events [API-1824]

zpqrtbnk:
This PR groups all events wiring in one single place in `HazelcastClient`, thus giving a clear view of what happens, and when.

Also, it re-orders some events to ensure that, for instance, Compact Serialization schemas are sent to a new cluster *before* the client is considered connected and can issue user-level invocations. This is documented in the source code.

### DIFF
--- a/src/Hazelcast.Net.Tests/Serialization/Compact/SchemasLocalTests.cs
+++ b/src/Hazelcast.Net.Tests/Serialization/Compact/SchemasLocalTests.cs
@@ -19,6 +19,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Hazelcast.Clustering;
 using Hazelcast.Core;
 using Hazelcast.Exceptions;
 using Hazelcast.Messaging;
@@ -50,6 +51,11 @@ public class SchemasLocalTests
         {
             if (triggerEvents && SendingMessage != null) await SendingMessage.AwaitEach(requestMessage).CfAwait();
             return await _respond(requestMessage);
+        }
+
+        public Task<ClientMessage> SendToMemberAsync(ClientMessage message, MemberConnection memberConnection, CancellationToken cancellationToken = default)
+        {
+            throw new NotImplementedException();
         }
 
         public async Task<ClientMessage> SendAsync(ClientMessage requestMessage, CancellationToken cancellationToken)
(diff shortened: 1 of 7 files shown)


src/Hazelcast.Net/HazelcastClient.cs
```diff
@@ -79,51 +79,88 @@ public HazelcastClient(HazelcastOptions options, Cluster cluster, ILoggerFactory
...
+            // when the first connection to a cluster is opened, make sure we recreate the distributed objects
+            Cluster.Connections.ConnectionOpened += _distributedOjects.OnConnectionOpened;
 
             // when the first connection to a cluster is opened, make sure we republish all schemas
-            Cluster.Connections.ConnectionOpened
-                += Cluster.SerializationService.CompactSerializer.Schemas.OnConnectionOpened;
+            Cluster.Connections.ConnectionOpened += Cluster.SerializationService.CompactSerializer.Schemas.OnConnectionOpened;
+
+            // when a connection is opened, install event subscriptions + ensure there is a cluster views connection
+            Cluster.Connections.ConnectionOpened += Cluster.Events.OnConnectionOpened;
```
emreyigit, line 152: May be it's a silly question but how is it likely that a subscription is triggered before client state is set to "connected"?

zpqrtbnk: as soon as the subscription is registered on the server, we can receive an event ... and the client may end up handling that event *before* it is considered "connected" which might indeed be considered a Bad Idea. 

but if we do it the other way round, the client may do a map.put operation as soon as connected, assuming it will trigger and event and *not* get an event because the subscription hasn't been registered yet.

java does subscriptions as "urgent" invocations - so they'll be registered before the client is fully connected, too. so I *think* we're OK.

src/Hazelcast.Net.Tests/Serialization/Compact/SchemasLocalTests.cs
```diff
@@ -52,6 +53,11 @@ public async Task<ClientMessage> SendAsync(ClientMessage requestMessage, bool tr
             return await _respond(requestMessage);
         }
 
+        public Task<ClientMessage> SendToMemberAsync(ClientMessage message, MemberConnection memberConnection, CancellationToken cancellationToken = default)
+        {
+            throw new NotImplementedException();
```
emreyigit, line 58: What about them? When will you complete?

zpqrtbnk: Will not - this is a stub messaging object, we only need to implement what's needed for the tests where it's used.

emreyigit (approved):

PR #773, opened 2023-01-09
Fix Compact again

zpqrtbnk:
Latest fixes for 5.2 - includes:
* Can deserialize Compact objects in SQL queries (i.e. ensure that the lazy deserialization of rows will work)
* Can send compact objects in the Authentication message with embedded schema
Comes with tests.


src/Hazelcast.Net.Linq.Async/Naming.cs
```diff
@@ -0,0 +1,28 @@
...
+
+using System;
+
+namespace Hazelcast.Linq
+{
+    internal class Naming
+    {
+        public static string EscapeColumnName(string s)
+        {
+            return s.Equals("value", StringComparison.OrdinalIgnoreCase)
```
emreyigit, line 23: If we escape that, user also should create the mapping accordingly which is like you did in the tests. I'm still not sure whether to touch naming since user should also participate to make it work but could you at least mention name escaping at LINQ docs?

zpqrtbnk: Valid concern. I did it because it was messing with tests, as you mention, but now ... yes ... it can be confusing. The situation "before" was that if I have a `HMap<string, Thing>` and `Thing` has a `Value` property, any LINQ query such as `.Where(x => x.Value == 2)` would cause an SQL exception. With this change... it works with the adjusted mapping, else it also produces a SQL exception.

zpqrtbnk: Have removed the magic thing and will note it as an issue.

emreyigit (dismissed):

src/Hazelcast.Net.Linq.Async/Naming.cs
emreyigit: Thanks. We can provide extensions for custom naming later.

emreyigit (dismissed):

emreyigit (approved):

PR #777, opened 2023-01-17
Fix null connection exception during dispose [API-1821]

emreyigit:
Fix null connection exception during dispose at cluster view subscription. `WaitForConnection ` returns null if cancelation is requested. During process, `SubscribeToClusterViewsAsync` takes null connection as an argument and fails. That causes problem while disposing the client, and so on tests.

Edit: Also, fixed the default value of multiplier in `ConnectionRetryOptions`.

### DIFF
--- a/src/Hazelcast.Net/Clustering/ClusterEvents.cs
+++ b/src/Hazelcast.Net/Clustering/ClusterEvents.cs
@@ -503,6 +503,10 @@ namespace Hazelcast.Clustering
                 // try to subscribe, relying on the default invocation timeout,
                 // so this is not going to last forever - we know it will end
                 var correlationId = _clusterState.GetNextCorrelationId();
+                
+                // Check once before subscription, state could be changed while waiting for a connection.
+                if(cancellationToken.IsCancellationRequested) break;
+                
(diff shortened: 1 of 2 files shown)


src/Hazelcast.Net/Clustering/ClusterEvents.cs
```diff
@@ -503,6 +503,10 @@ private async Task AssignClusterViewsConnectionAsync(MemberConnection connection
                 // try to subscribe, relying on the default invocation timeout,
                 // so this is not going to last forever - we know it will end
                 var correlationId = _clusterState.GetNextCorrelationId();
+
```
zpqrtbnk, line 506: I am not sure about this test. What happens is, `await WaitForConnection(cancellationToken)` above may return `null` in case of a cancellation, in which case we should break/loop. I *think* it would be cleaned this way:

`if (connection == null) continue:`

emreyigit: Sure, it leads to same end. If we continue, while look will break it. Changing it.

zpqrtbnk (approved):
approved

PR #780, opened 2023-01-20
Make Retry Strategy Logs More Visible [API-1832]

emreyigit:
Retry strategy was logging in debug level. It's hard to see if retry strategy is working in normal circumstances. So, aligned the log level with Java client and made it more visible.


zpqrtbnk (approved):
approved

PR #782, opened 2023-01-30
[API-1824] Fix Compact Serialization as per QA

zpqrtbnk:
This PR addresses issues raised during QA review of Compact Serialization. This issues are highlighted by the newly introduced `CompactQaTests` class. The two issues were:
* As per `ClusterRestart` test, the client would hang when trying to reconnect to a cluster. This was caused by `Schemas.PublishSchemasAsync` where we need to use `SendToMemberAsync` not `SendAsync` when reconnecting, but we forgot to update one of the two cases of the `switch` statement.
* As per `ReadSchemaAfterWrite_withObjectValueType`, a map of `<int, object>` would fail to deserialize an object `IGenericRecord` and return it as `object`. This was due to `CompactSerializationSerializer.GetOrCreateRegistration` throwing an exception - there was even a comment about it not being OK for generic records. The feature has now been implemented.

In addition working on these issues has revealed some more networking instabilities.

In addition, while working on the first issue, we realized that, should an exception occur during the execution of the `ConnectionOpened` event handlers, the connection could end up in an unstable state (never disposed nor closed) and the client may hang. That is illustrated by the `ClientGoesOffline` test. Issue is fixed in `ClusterConnections` where we now correctly handle the situation (if a handler throws, the connection is teared down).

This was mostly due to the background connection task running even when the client is not connected. This is fixed by making sure this task only runs when the client is considered connected. However, this breaks the "address-rerouting" feature of the .NET client where we redirected a connection to the exact address reported by a member. This feature has therefore been disabled.

### DIFF
--- a/src/Hazelcast.Net/Clustering/ClusterConnections.cs
+++ b/src/Hazelcast.Net/Clustering/ClusterConnections.cs
@@ -895,6 +895,11 @@ namespace Hazelcast.Clustering
             // isFirst connection will indeed trigger before any other connection is created - think
             // about it if adding support for parallel connections!
 
+            void RemoveCompletion()
+            {
+                if (_completions.TryRemove(connection, out var completion)) completion.TrySetResult(null);
+            }
+
             // connection is opened
             try
             {
@@ -907,19 +912,12 @@ namespace Hazelcast.Clustering
                 // else dispose leads to OnConnectionClosed which would wait on the completion and thus
                 // would hang - the completion is here to prevent the closed event from triggering before
                 // the opened even, but only if we manage to open properly.
-                // there is always a completion, but we have to TryRemove from concurrent dictionaries
-                if (_completions.TryRemove(connection, out var completion)) completion.SetResult(null);
-                // now it's safe to dispose
+                RemoveCompletion();
                 await connection.DisposeAsync().CfAwait();
                 throw;
             }
 
-            lock (_mutex)
-            {
-                // there is always a completion, but we have to TryRemove from concurrent dictionaries
-                if (_completions.TryRemove(connection, out var completion)) completion.SetResult(null);
-            }
-
+            RemoveCompletion();
             return connection;
         }
 


mdumandag (dismissed):
I am approving for Compact related changes & tests, but there are lots of places that I am not familiar with. I will conduct some tests after the PR is merged as well.

BTW, I believe the changes here are substantial and we should perform all of our release tests again (not just the soak test) after merging this one. Also, regarding the functionality you said to disable, why did we add it? Was it a customer request? I am asking this to make sure that we are not breaking anybody's use case by disabling it

src/Hazelcast.Net/Clustering/ClusterConnections.cs
```diff
@@ -889,7 +896,23 @@ private async Task<MemberConnection> ConnectAsync(NetworkAddress address, Cancel
             // about it if adding support for parallel connections!
 
             // connection is opened
-            await RaiseConnectionOpened(connection, isFirstEver, isFirst, isNewCluster).CfAwait();
+            try
+            {
+                await RaiseConnectionOpened(connection, isFirstEver, isFirst, isNewCluster).CfAwait();
```
emreyigit, line 901: Confused a bit here. We set completion null in either case here and just below in the locked scope after `RaiseConnectionOpened`. Isn't this should set exception instead? I see that we should await here before throw but I couldn't predict the result of setting null in both cases.

src/Hazelcast.Net/Clustering/ClusterConnections.cs
```diff
@@ -912,7 +935,18 @@ public async ValueTask DisposeAsync()
             // ended by now
             _cancel.Cancel();
             if (_connectMembers != null)
-                await _connectMembers.CfAwaitCanceled();
+            {
+                try
+                {
+                    await _connectMembers.CfAwait(120_000); // give it 2 mins
```
emreyigit, line 941: What could be result if we don't wait here since disposing is happening? Is there a reason to wait?

src/Hazelcast.Net/Clustering/ClusterMembers.cs
```diff
@@ -328,8 +333,9 @@ public async Task RemoveConnectionAsync(MemberConnection connection)
             }
             finally
             {
-                // don't forget to resume the queue
-                _memberConnectionQueue?.Resume(drain);
```
emreyigit, line 332: Verifying my understanding. We disabled the whole member connection queue since client is not connected but we know that when client disconnected, reconnect will be triggered  (by cient state changed event), right?

emreyigit:
Have a few comments. Also, I agree with @mdumandag we should test the release although I don't think disabling `IsMatchAddress` will effect the user.

src/Hazelcast.Net/Clustering/ClusterConnections.cs
zpqrtbnk: if `RaiseConnectionOpened` fails, we're going to trash the connection and throw - but before that we want to get rid of the completion. OTOH if it does not fail, we also need to remove the completion. So it's either (a) `RaiseConnectionOpened` succeeds, we enter the mutex block and remove the completion, or (b) it fails, we remove the completion, and throw.

We cannot remove the completion in a `finally` block because we need to do it *before* we dispose the connection.

Now... what strucks me is... why use a mutex in one case and not the other? Thinking about it...

src/Hazelcast.Net/Clustering/ClusterMembers.cs
zpqrtbnk: When the client is disconnected.... we will enter the "first connection" process again, ie we'll try to establish a new connection to the cluster again. if that succeeds... the new connection will be added to members, we'll realize we have a member <--> connection match, we'll deduce that we are connected - and *then* we'll resume the connection queue so it handles all the other members and constantly make sure that all members are connected.

when we lose 1 connection... if it's one amongst others (there still are some members connected) the queue will take care of it: the member is queued, and eventually re-connected.

OTOH if we lose the last connection... we go disconnected and we stop the queue, because we're going to start connecting the "first member" and we cannot have a parallel process *also* connecting members - because the "first member" is special.

src/Hazelcast.Net/Clustering/ClusterConnections.cs
emreyigit: I think that's okay since we expect that exception is not likely and operation is just removing from the dictionary.

emreyigit (dismissed):

mdumandag (approved):
re-approving since the previous approval got discarded

PR #785, opened 2023-02-03
Default Reconnect Mode Set to Sync [API-1857]

emreyigit:
Corrected the default mode, and fixed the related tests. Documentation matches with the behavior now.


zpqrtbnk (approved):
Approved

PR #786, opened 2023-02-03
Session State Provider [API-83]

emreyigit:
Async extensions and documentation added for `ISession` interface.


zpqrtbnk:
So just to be sure - the simple fact that we register our own `IDistributedCache` implementation for the web app cache, means that sessions will rely on it, and load/store themselves from it, all magically? That's great. Have you tested that it works with a sample web app?

And then, the few extension methods that you provide are for providing asynchronous access to the session, right? Because the default mode is synchronous and can cause threading issues, right?

src/Hazelcast.Net.Caching/Session/SessionExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// Gets a string value from <see cref="ISession"/>.
+    /// </summary>
+    /// <remarks>It loads the session as async. It is strongly advised to use async extensions over Hazelcast distributed
+    /// cache implementation.</remarks>
+    /// <param name="session">The <see cref="ISession"/>.</param>
+    /// <param name="key">The key to read.</param>
+    /// /// <param name="token">Cancellation Token</param>
+    public static async Task<string?> GetStringAsync(this ISession session, string key, CancellationToken token = default)
+    {
+        await session.LoadAsync(token);
```
zpqrtbnk, line 36: make sure we `ConfigureAwait(false)` here and in other methods, it's important in web apps (I believe).

doc/dev/doc/extensions/caching.md
```diff
@@ -51,4 +51,20 @@ With these declarations, the Hazelcast cache implementation would be injected in
...
+in your ASP.NET Core project. 
+
+```csharp
+var builder = WebApplication.CreateBuilder(args);
+builder.Services.AddHazelcastCache();
+var app = builder.Build();
+app.UseSession(); // Session state will be stored in distributed cache provided by AddHazelcastCache()
+```
+> [!WARNING]
+> While reading `ISession` object, it is strongly advised to use async extensions which are provided with this package under `Hazelcast.Caching.Session` namespace. Otherwise, you may not benefit
```
zpqrtbnk, line 68: Maybe: When the `ISession` object is supported by the Hazelcast client, its synchronous operations should be avoided, as the Hazelcast client is fully asynchronous and therefore calls to `task.GetAwaiter().GetResult()` may be involved and, potentially, cause threading issues. Prefer the asynchronous operations, such as those provided in the `SessionExtensions` class.

zpqrtbnk (approved):
Commented

doc/dev/doc/extensions/caching.md
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/commit/535f7e6a3a83daba26e75bf2a0903c63fbf3ed73.

zpqrtbnk (approved):
approved

PR #798, opened 2023-02-28
Sql Partition Aware Client [API-1861]

emreyigit:
Sql partition aware feature implemented. Since feature requires protocol updates, protocol submodule and the files are regenerated and aligned.


src/Hazelcast.Net/Sql/SqlService.cs
```diff
@@ -13,25 +13,36 @@
...
     {
         private readonly Cluster _cluster;
         private readonly SerializationService _serializationService;
+        // internal for tests only
+        internal readonly ReadOptimizedLruCache<string, int> _queryPartitionArgumentCache;
+        private readonly ILogger<SqlService> _logger;
+        private readonly HazelcastOptions _options;
 
-        internal SqlService(Cluster cluster, SerializationService serializationService)
+        internal SqlService(HazelcastOptions options, Cluster cluster, SerializationService serializationService, ILoggerFactory loggerFactory)
```
zpqrtbnk, line 39: i'd like to avoid passing the entire `HazelcastOptions` to the SQL service. looks like it only requires a `SqlOptions` instance + a flag to tell whether to do SQL partitioning or not. could be an internal `SqlOptions.PartitionEnabled` or something, that would be set by the client when creating the service. that way we don't leak any network stuff into the SQL service.

src/Hazelcast.Net/Protocol/Codecs/ClientAuthenticationCodec.cs
```diff
@@ -198,10 +198,21 @@ public sealed class ResponseParameters
             /// Returns true if server supports clients with failover feature.
             ///</summary>
             public bool FailoverSupported { get; set; }
+
```
zpqrtbnk, line 201: leaking some Alto changes that should not be part of this PR

src/Hazelcast.Net/Protocol/Codecs/ClientAuthenticationCustomCodec.cs
```diff
@@ -189,10 +189,21 @@ public sealed class ResponseParameters
             /// Returns true if server supports clients with failover feature.
             ///</summary>
             public bool FailoverSupported { get; set; }
+
```
zpqrtbnk, line 192: leaking some Alto stuff that should not be part of this PR

src/Hazelcast.Net/Sql/SqlService.cs
```diff
@@ -112,36 +124,88 @@ public Task<long> ExecuteCommandAsync(string sql, SqlStatementOptions options =
...
+            return (new SqlRowMetadata(result.RowMetadata), result.RowPage, partitionId);
         }
 
-        private async Task<SqlPage> FetchNextPageAsync(SqlQueryId queryId, int cursorBufferSize, CancellationToken cancellationToken)
+        private async Task<SqlPage> FetchNextPageAsync(SqlQueryId queryId, int cursorBufferSize, int partitionId, CancellationToken cancellationToken)
         {
             var requestMessage = SqlFetchCodec.EncodeRequest(queryId, cursorBufferSize);
-            var responseMessage = await _cluster.Messaging.SendAsync(requestMessage, cancellationToken).CfAwait();
+            var responseMessage = partitionId == -1 ?
+                await _cluster.Messaging.SendAsync(requestMessage, cancellationToken)
```
zpqrtbnk, line 206: can you put the `?` and `:` aligned:
```var ... = ...
    ? .....
    : .... ;```
as it's easier to read?
and then, you're issing a `CfAwait()` on one of the two calls

**Review by zpqrtbnk — COMMENTED** [review]
have reviewed it all except the `ReadOptimizedLruCache` but already posting comments.

File: src/Hazelcast.Net/Core/ReadOptimizedLruCache.cs [hunk 5]
```diff
@@ -0,0 +1,165 @@
...
+        if (_cleaningSlim.CurrentCount == 0) return;
+
+        _cleaningSlim.Wait();
+
+        try
+        {
+            if (Cache.Count < _threshold) return;
+
+            var countOfEntriesToRemoved = Cache.Count - _capacity;
+#if NET6_0_OR_GREATER
```
zpqrtbnk, line 129: have you benchmarked these methods? on my machine,
```var cutoff = Cache.OrderBy(x => x.Value.LastTouch).ElementAt(Index).Value.LastTouch;```
is the fastest option, putting the `long` values in a `List<long>` and then using `List.Sort()` is only slightly slower, and the priority queue method is generally slower. Or, almost all of the time. What do you think?

**Review by zpqrtbnk — COMMENTED** [review]
more comments

File: src/Hazelcast.Net/Core/ReadOptimizedLruCache.cs (hunk 5 above)
Comment (emreyigit) [reply src/Hazelcast.Net/Core/ReadOptimizedLruCache.cs:129]: I've committed a benchmark min-heap vs ordering at `Hazelcast.Net.Benchmarks/LruCacheEviction.cs` . Min-heap performed better in any size of cache. Also, Java implementation has similar results here https://github.com/hazelcast/hazelcast/pull/22659#discussion_r1111783551 . I'm surprised that you found the opposite.

File: src/Hazelcast.Net/Core/ReadOptimizedLruCache.cs [hunk 6]
```diff
@@ -0,0 +1,165 @@
...
+        {
+            value = entry.Value;
+            return true;
+        }
+
+        value = default;
+        return false;
+    }
+
+    private void DoEviction()
```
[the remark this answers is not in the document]
emreyigit: Yes, it's sync but it's not expect to evict the cache frequently. The expectation is based on reading. I think this block is acceptable.

src/Hazelcast.Net/Protocol/Codecs/ClientAuthenticationCodec.cs
emreyigit: Let me also add this for the record. The changes will be part of 5.3 which covers alto and SQL partition aware feature, and codecs cover both cases.

src/Hazelcast.Net/Protocol/Codecs/ClientAuthenticationCustomCodec.cs
emreyigit: Let me also add this for the record. The changes will be part of 5.3 which covers alto and SQL partition aware feature, and codecs cover both cases.

src/Hazelcast.Net/Sql/SqlService.cs
emreyigit: Sure we can separate options into required fragments but I couldn't understand the concern here. What does it mean we don't leak network stuff since everything is internal in the context?

src/Hazelcast.Net/Protocol/Codecs/ClientAuthenticationCustomCodec.cs
zpqrtbnk: same as above

src/Hazelcast.Net/Sql/SqlService.cs
zpqrtbnk: don't think we understand each other, my idea was:
```
var responseMessage = partitionId == -1 
    ? await _cluster.Messaging.SendAsync(requestMessage, cancellationToken).CfAwait()
    : await _cluster.Messaging.SendToPartitionOwnerAsync(requestMessage, partitionId, cancellationToken).CfAwait();
```

src/Hazelcast.Net/Core/ReadOptimizedLruCache.cs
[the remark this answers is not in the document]
zpqrtbnk: ok

src/Hazelcast.Net/Core/ReadOptimizedLruCache.cs
zpqrtbnk: strange indeed. that being said, my benchmark just determines the cutoff value, it does not do the whole eviction. ok, let's say it's fine the way it is now.

src/Hazelcast.Net/Sql/SqlService.cs
```diff
@@ -112,36 +124,88 @@ public Task<long> ExecuteCommandAsync(string sql, SqlStatementOptions options =
...
             if (response.Error != null)
                 throw new HazelcastSqlException(_cluster.ClientId, response.Error);
 
-            return response;
+            return (response, partitionId);
+        }
+
+        private void SetArgumentIndex(string sql, int argIndexFromServer)
+        {
+            if (_options.Networking.SmartRouting &&
```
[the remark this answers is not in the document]
emreyigit: `cache.Add` built on `GetOrAdd`, so we don't update if it exists. I think current state is good enough.

src/Hazelcast.Net/Sql/SqlService.cs
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/798/commits/a91d960491af90459c71f0821c8ccfdd5e85d119

src/Hazelcast.Net/Protocol/Codecs/ClientAuthenticationCustomCodec.cs
emreyigit: Re generated the files with latest protocol at https://github.com/hazelcast/hazelcast-csharp-client/pull/798/commits/a91d960491af90459c71f0821c8ccfdd5e85d119

src/Hazelcast.Net/Protocol/Codecs/ClientAuthenticationCodec.cs
emreyigit: Regenerated the files with latest protocol at https://github.com/hazelcast/hazelcast-csharp-client/pull/798/commits/a91d960491af90459c71f0821c8ccfdd5e85d119

src/Hazelcast.Net/Sql/SqlService.cs
emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/798/commits/a91d960491af90459c71f0821c8ccfdd5e85d119

zpqrtbnk (approved):
approved

PR #803, opened 2023-03-03
Viridian Cloud Tests [API-1883]

emreyigit:
The tests and related structure is implemented to work with https://github.com/hazelcast/client-compatibility-suites/pull/51


zpqrtbnk:
I am OK with the changes but have 1 question. IIC the Thift files have been generated by Thrift version 0.18. But the files were previously generated by Thrift version 0.15 and the reference to the Thrift package in the tests is version 0.15 and I seem to remember that we could not change the version because the RC is using 0.15 and we have to use the same version. If there any chance you can generate the files with version 0.15 just to be sure? And maybe verify that it's the version used by the RC?

emreyigit:
Well, it's interesting that RC uses Thrift 0.13, and tests are passed. Do you have any further suggestion?

zpqrtbnk (approved):
Approved

PR #804, opened 2023-03-12
Implement invocation flags + a/sync reconnect modes [API-1803]

zpqrtbnk:
This PR implements "invocation flags" which is a better name for what Java names "priority" or "urgent invocations" and really are invocations that should be invoked even though the client is not considered as "connected". This allows us to also implement a/sync reconnect modes properly in the .NET client.


src/Hazelcast.Net.Tests/Cloud/CloudTests.cs
```diff
@@ -95,15 +94,7 @@ public async Task SampleClient(bool previewOptions)
...
-            if (previewOptions)
-            {
-                options.Preview.EnableNewReconnectOptions = true;
-                options.Preview.EnableNewRetryOptions = true;
-            }
-            else
-            {
-                options.Networking.ReconnectMode = ReconnectMode.ReconnectAsync;
-            }
+            options.Networking.ReconnectMode = ReconnectMode.ReconnectAsync;
```
emreyigit, line 97: I think it should be `ReconnectSync` since test doesn't expect to face an exception.

doc/dev/doc/network.md
```diff
@@ -28,7 +28,9 @@ In the unisocket client mode, the client will only connect to one of the configu
...
+The client can, at times, become disconnected from the cluster, for instance in case of a brief network issue. What happens then is determined by the `@Hazelcast.Networking.NetworkingOptions.ReconnectMode` configuration option. Its default value is `@Hazelcast.Networking.ReconnectMode.ReconnectSync` which means that the client will try to reconnect transparently, queuing the invocations until the client is reconnected, or the invocations time out.
+
+Alternatively, the option can be set to `@Hazelcast.Networking.ReconnectMode.ReconnectAsync` to let the client try to reconnect, but fail all invocations immediately for as long as the client has not reconnected. Finally, it can be set to `@Hazelcast.Networking.ReconnectMode.DoNotReconnect` to prevent the client from reconnecting and switching it to a non-recoverable error state (i.e. the client must be destroyed and a new client must be recreated).
```
emreyigit, line 33: What will it causes to non-recoverable error? I exclude the environmental problems, network etc. But invocations are in our control. I'm not sure about this sentence.

emreyigit:
Looks nice, I have few small comments.

doc/dev/doc/network.md
zpqrtbnk: What this mean is... you can instruct the client to not reconnect and shutdown instead, in "error state"... that sentence comes from Java and I am not sure I fully understand it either, will update.

src/Hazelcast.Net.Tests/Cloud/CloudTests.cs
zpqrtbnk: ok

doc/dev/doc/network.md
zpqrtbnk: pushed a fix

emreyigit (approved):
Approved.

Issue #799, opened 2023-03-01
Missing "GetDistributedObjects()" function [API-1916]


neilstevenson:
On C# 5.2.0,
https://hazelcast.github.io/hazelcast-csharp-client/5.2.0/api/Hazelcast.IHazelcastClient.html
there is no way to list the distributed objects in the system

`GetDistributedObjects()` or similar

mdumandag:
closing this as it is done in https://github.com/hazelcast/hazelcast-csharp-client/pull/805

PR #805, opened 2023-03-14
Implement GetDistributedObjects [API-1916]

zpqrtbnk:
Implements `GetDistributedObjects` as is available in Java (see #799).


src/Hazelcast.Net/DistributedObjects/DistributedObjectFactory.cs
```diff
@@ -57,6 +58,49 @@ public DistributedObjectFactory(Cluster cluster, SerializationService serializat
...
+                // at that point Java would actually make sure that the client has proxies for
+                // each object that was reported by the cluster - however that *cannot* work in
+                // .NET due to strong generics, because we cannot, for instance, create a
+                // HMap<TKey, TValue> object, since we don't know TKey nor TValue. this is not
+                // an issue really, proxies will be created on-demand.
+
+                result.Add(info); // build result
+            }
+
+            // locally destroy objects that are not remotely-known
```
emreyigit, line 94: I know it the logic is from Java but removing the local object is odd. Invoking a read only method modifies the client :S It should be idempotent, right? However, most probably, this is the reason that this method is not much promoted.

zpqrtbnk: agree, but I've tried to stay close to Java (for once)

src/Hazelcast.Net/IHazelcastClient.cs
```diff
@@ -153,6 +153,12 @@ public interface IHazelcastClient : IAsyncDisposable
         /// </remarks>
         ValueTask DestroyAsync(IDistributedObject o);
 
+        /// <summary>
+        /// Gets information about all distributed objects known to the cluster.
+        /// </summary>
+        /// <returns>Information about all distributed objects know to the cluster.</returns>
+        Task<ICollection<DistributedObjectInfo>> GetDistributedObjectsAsync();
```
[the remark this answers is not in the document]
zpqrtbnk: makes sense, pushed a fix

emreyigit (approved):
Approved.

PR #809, opened 2023-03-17
Updates to be work on backward compatibility.

emreyigit:
There are some changes for beta tests and backward compatibility tests.

### DIFF
--- a/hz.ps1
+++ b/hz.ps1
@@ -144,6 +144,10 @@ $params = @(
     },
     @{ name = "yolo";            type = [switch]; default = $false;
        desc = "confirms excution of sensitive actions"
+    },
+    @{
+        name="source"; type = [string];  default = $null;
+        desc = "source folder to be copied"
     }
 )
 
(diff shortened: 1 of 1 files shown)


hz.ps1
```diff
@@ -2562,6 +2590,30 @@ function hz-getfwks-json {
     ConvertTo-Json -InputObject $frameworks -Compress
 }
 
+## Copies files from given path to project folder with respect to source hierarhcy.
+function hz-copy-source (){
```
zpqrtbnk, line 2594: not sure I understand the nature of the function from its name. is the purpose to copy "some files" from "somewhere" to "somewhere else"? then I'd rename the command as `copy-files` and its corresponding option as `copy-files-source` to be super explicit. we might have something else having a "source" in the future.

hz.ps1
```diff
@@ -2562,6 +2590,30 @@ function hz-getfwks-json {
...
+
+    $source = $options.source
+
+    if(-not (test-path -path $source)){
+        Die "$($source) is not exist."
+    }
+
+    $currentPath = get-location
```
zpqrtbnk, line 2602: not sure I like this. the script *could* potentially run from anywhere. we have a range of defined directories (see top of `hz.ps1`) including the solution's root. think it would be more reliable to reference one of these instead of "the current location"

zpqrtbnk (approved):
happy

PR #812, opened 2023-04-03
Fix of Invalid Server Name Information (SNI) Issue [API-1919]

emreyigit:
See the issue for details. If no certificate name is given as target host, connection IP address will be accepted as target hostname. It is tested with .Net4.8 and .Net7.0 against to Hazelcast 5.3.0-SNAPSHOT on Java 17.


src/Hazelcast.Net/Networking/SslLayer.cs
```diff
@@ -190,5 +188,17 @@ internal bool ValidateCertificate(object sender, X509Certificate cert, X509Chain
...
+        // Methodized for testing. 
+        public string GetTargetHostNameOrDefault()
+        {
+            // if targetHost does not match the server certificate name then a RemoteCertificateNameMismatch error will
+            // be reported, which can be ignored with options.ValidateCertificateName being false. If it is true, then
+            // options.CertificateName *must* be set to the server certificate name.
+            // + if no options.CertificateName is given, .Net 4.8 assigns random numbers which fails on Hazelcast at Java 17+.
+            // We set the target host name to connecting address if nothing to have. See issue #800.
+
+            return _options.CertificateName ?? _ipAddress.ToString();
```
zpqrtbnk, line 201: can u add a //TODO comment saying we should refactor this and use a new _options.RemoteName or whatever?

emreyigit: Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/812/commits/334a80863b25f869cc68077a4e41ebaec9fe1f33

zpqrtbnk (approved):
approved

PR #815, opened 2023-04-07
Drop .net5 and .net core 3.1 support [API-2015]

emreyigit:


emreyigit:
Since `Get Frameworks` step runs over `hazelcast/hazecast-csharp-client`, it cannot find the builds of dropped versions. The PR should be merged without the checks.

zpqrtbnk (approved):
approved

Issue #818, opened 2023-04-12
Client becomes stale when forcing to use `async` APIs `sync`. [API-2023]


emreyigit:
Client becomes stale, most probably entering a deadlock, when style of initialization client and invoking async APIs are diverged.

```csharp
var client = await HazelcastClientFactory.StartNewClientAsync(options);
var map = client.GetMapAsync<string, string>("mapname").Result;
```

In this case, the map never created, and client hung. Although, `Task.Result` is dangerous, sync usage should be provided as much as possible.

PR #819, opened 2023-04-13
Issue 818, asynchronous deadlocks [API-2023]

zpqrtbnk:
As reported on #818 and the corresponding JIRA ticket, client can hang when mixing synchronous (`task.GetAwaiter().Result`) and synchronous (`await task`) patterns. This PR addresses one situation where a `TaskCompletionSource` was aggressively left synchronous where we should force it to be asynchronous.

### DIFF
--- a/src/Hazelcast.Net.Tests/Support/Issue818.cs
+++ b/src/Hazelcast.Net.Tests/Support/Issue818.cs
@@ -1,0 +1,67 @@
+﻿// define SYNC_CLIENT
+#define NOT_SYNC_CLIENT
+
+// define CFAWAIT_CLIENT
+#define NOT_CFAWAIT_CLIENT
+
+// define SYNC_MAP
+#define SYNC_MAP
+
+using System;
+using System.Threading.Tasks;
+using Hazelcast.Networking;
+using Microsoft.Extensions.Logging;
+using NUnit.Framework;
+
+namespace Hazelcast.Tests.Support;
+
+[TestFixture]
+public class Issue818
+{
+    [Test]
+    [Timeout(30_000)]
+    public async Task Reproduce()
+    {
+        var options = new HazelcastOptionsBuilder()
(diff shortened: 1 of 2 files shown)


src/Hazelcast.Net.Tests/Support/Issue818.cs
```diff
@@ -0,0 +1,76 @@
...
+
+[TestFixture]
+public class Issue818 : SingleMemberClientRemoteTestBase
+{
+    [Test]
+    [Timeout(30_000)]
+    public async Task Reproduce()
+    {
+        var options = new HazelcastOptionsBuilder()
+            //.With((config, options) => options.Networking.Addresses.Add("127.0.0.1"))
```
emreyigit, line 27: As a note, this usage pattern of the options was user behavior.

src/Hazelcast.Net/Core/SocketExtensions.cs
```diff
@@ -49,50 +49,59 @@ public static async Task ConnectAsync(this Socket socket, EndPoint endPoint, int
...
+            {
+                timeoutCancel.Cancel(); // don't leave the timeoutTask running
+                await timeoutTask.CfAwaitNoThrow(); // and make sure it does not leak unobserved exception
+            }
 
-            if (task == connected.Task)
+            if (task != timeoutTask)
             {
-                await task.CfAwait(); // throw if there is anything to throw (failed to connect...)
+                connected.Task.GetAwaiter().GetResult(); // we know the task has completed - throw if needed
```
emreyigit, line 100: Here is `GetAwaiter()` called because `await` can introduce unnecessary state machine on a completed task, right?

zpqrtbnk: correct

emreyigit (approved):

PR #833, opened 2023-06-05
Implement TLS/SSL workaround for issue with client cert. chain on Linux [API-2044]

zpqrtbnk:
When the client certificate is composed of a chain (root CA → sub CA → cert) the TLS/SSL handshake fails on Linux but works fine on Windows. This seems to be an issue with the way the underlying .NET / OS layer manages TLS/SSL certificate - not an issue with our code.

This PR implements a workaround in our `SslLayer` that implements an explicit hook for selecting the certificate, when running on Linux and the client certificate is actually a chain. This workaround:
* Ensures we select the proper certificate within the chain
* Ensures that the entire chain is sent to the cluster

This workaround has been sent to the customer as part of a preview release, and tested by them. They have confirmed it is working in their case.


src/Hazelcast.Net/Networking/SslLayer.cs
```diff
@@ -65,9 +66,42 @@ public async ValueTask<Stream> GetStreamAsync(Stream stream)
...
+                // for the whole cert chain to be sent as part of the handshake we need to copy the
+                // certs that don't have a PK to the current user CA certificate store
+                var store = new X509Store(StoreName.CertificateAuthority, StoreLocation.CurrentUser);
+                try
+                {
+                    store.Open(OpenFlags.ReadWrite);
+                    foreach (var cert in clientCertificates)
+                    {
+                        _logger.IfDebug()?.LogDebug($"CertFix: {cert.Subject}   (issuer: {cert.Issuer})   {cert.Thumbprint}   [{(cert.HasPrivateKey ? "PK" : "ADD")}]");
+                        if (!cert.HasPrivateKey) store.Add(cert);
```
emreyigit, line 93: Here, given certificates are added to the store. Is it a permanent action on current user's store? If yes, I think having a warning message could be appropriate.

zpqrtbnk: Makes sense, will do. I also notice that the test coverage is decreasing with this PR, going to look into it.

emreyigit (approved):

PR #851, opened 2023-09-27
Remove DotNetZip dependency [API-2140]

zpqrtbnk:
DotNetZip is for some reason flagged with a MEDIUM security issue by SonaType, most probably due to the use of `Random` even though not for security purposes. DotNetZip is required for our .NET Framework support as its supports a ZLib compression mode that only .NET 6+supports, and is required for compressing our metrics so that Java can uncompress them.

DotNetZip is old and not going to be fixed. There are no easy alternative that are 100% to be safe.

So this PR is: use built-in .NET support for .NET 6+ versions of our client, and embed the very minimal part of DotNetZip (that does NOT use `Random`) for .NET Framework versions of our client. Comes with revised tests that should validate that the metrics can be unzipped by Java.


protocol
emreyigit, line 1: Was this necessary? I think when we sync the latest protocol commit, we should also re-generate the protocols. I think it is irrelevant to the PR.

zpqrtbnk: my mistake, removing.

PR #865, opened 2023-12-18
Implement dynamic configuration / map+ringbuffer [API-2175]

zpqrtbnk:
This PR implements the core dynamic configuration support with all codecs etc + the map and ringbuffer configuration user code. Anything else is not part of this PR as it will quite probably be exposed via a REST API and *not* via dynamic configuration fluent API. Users can do, for instance:
```
await client.DynamicOptions.ConfigureMapAsync("map-name", mapOptions => 
{
    map.EventJournal.Enabled = true;
});
```
Note
* in order to be "C# idiomatic" and to align with .NET naming everything "config" is "options" in .NET code
* the codecs have been rebuilt with the latest protocol changes which means that a few things had to be adusted here and there in unrelated code as well.

We have tests to ensure that (1) the server would accept a complete configuration and (2) Java and .NET would produce the very same client message for the default dynamic configuration.

There are also a number of changes to files due to our public API surface going a bit out-of-control in the past PRs. In *theory* we have files that define our public API and detect any unwanted changes, but this detection seems to have been ignored in past PRs so it's all out of sync. Bringing it back in sync here.

NOTE: before merging this PR, ensure that https://github.com/hazelcast/hazelcast-client-protocol/pull/496 is merged and update the state of the protocol sub-repository accordingly.

NOTE: as part of the merge, ensure that all `v5.3.1` tag and `5.3.z` branch have been properly merged into `master`.


src/Hazelcast.Net.Tests/NearCache/NearCacheTest.cs
```diff
@@ -475,6 +475,10 @@ public async Task TestNearCacheLruEviction()
             await using var _ = new AsyncDisposable(dictionary.DestroyAsync);
             var cache = GetNearCache(dictionary);
 
+            var cachedMap = dictionary as Hazelcast.DistributedObjects.Impl.HMapWithCache<int, int>;
```
emreyigit, line 478: What is the goal here?

src/Hazelcast.Net/Models/CacheSimpleEntryListenerOptions.cs
```diff
@@ -0,0 +1,111 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Hazelcast.Configuration;
+using Hazelcast.Serialization;
+using System;
+
+namespace Hazelcast.Models;
+
+public class CacheSimpleEntryListenerOptions : IIdentifiedDataSerializable
```
emreyigit, line 21: Is this option used somewhere or related with map or ringbuffer?

emreyigit (changes requested):
I'm ok with the PR generally. I've left a few comments. Also, as I can see there are options but cannot be set or used yet. We should decrease their access modifiers since they are not avaiable tough declared as public. Also, excluding them from tests could be good idea until related implementation completed.  They decrease the overall coverage.

src/Hazelcast.Net/Configuration/DynamicOptions.cs
```diff
@@ -0,0 +1,138 @@
...
+            mapOptions.StatisticsEnabled,
+            mapOptions.SplitBrainProtectionName,
+            MapStoreConfigHolder.Of(mapOptions.MapStore),
+            NearCacheConfigHolder.Of(mapOptions.NearCache),
+            mapOptions.WanReplicationRef,
+            mapOptions.Indexes,
+            mapOptions.Attributes,
+            MapQueryCacheConfigs(mapOptions.QueryCaches),
+            mapOptions.PartitioningStrategy?.PartitioningStrategyClass,
+            null,
```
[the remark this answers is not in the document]
zpqrtbnk: In Java, you can pass classes (implementations) in the codec, that will be serialized. Like, the actual code. In .NET we don't have such Java code so we have nothing to pass, so in some places the options don't support everything supports, and we pass `null` in the codec.

src/Hazelcast.Net/Models/Capacity.cs
```diff
@@ -59,7 +53,106 @@ MemoryUnit.MegaBytes or
...
+    public static string ToPrettyString(long capacity, MemoryUnit unit)
+    {
+        // following Java's pattern
+
+        var bytes = MemoryUnitExtensions.Convert(MemoryUnit.Bytes, unit, capacity);
+        return bytes switch
+        {
+            >= 10_000_000_000 => To(MemoryUnit.GigaBytes),
+            >= 10_000_000 => To(MemoryUnit.MegaBytes),
+            >= 10_000 => To(MemoryUnit.KiloBytes),
```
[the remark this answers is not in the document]
zpqrtbnk: This is how it's done in Java, so 8192B shows as 8192B but 16384B shows as 16KB... and we do the same.

src/Hazelcast.Net.Tests/NearCache/NearCacheTest.cs
zpqrtbnk: ok, removing

src/Hazelcast.Net/Models/CacheSimpleEntryListenerOptions.cs
zpqrtbnk: It is indeed not used anywhere. Is it a leftover? In any case it should then be `internal`. Stephan to review all of them and mark `internal` where appropriate.

src/Hazelcast.Net.Tests/NearCache/NearCacheTest.cs
zpqrtbnk: Fixed with 7d76d74cd

src/Hazelcast.Net/Models/CacheSimpleEntryListenerOptions.cs
zpqrtbnk: Fixed with e23ed07ca it seems to be the only one

emreyigit (approved):

emreyigit (approved):

Issue #854, opened 2023-10-05
Error executing SQL query using IN operator [API-2160]


federicocodo:
I'm trying to execute a parameterized sql query using the IN operator, my query is very simple:
```sql
SELECT *
FROM map
WHERE columnName IN (?)
```
As a parameter I pass the Hazelcast client a simple merged list of structured strings like: 'value1', 'value2' but I get an error when running the Hazelcast client.

![image](https://github.com/hazelcast/hazelcast-csharp-client/assets/62144863/106889fc-3f3b-43da-b3d2-d7f2962bbc7c)

Update, I'm trying to debug to see what's happening and found out that it's caused because:

- the query shows more than one row and in my Hazelcast map one row is valued with double while the other is null;
- the hazelcast double class is trying to decode the frame that contains a null value and causes an exception.

Is possible to resolve?

emreyigit:
Hey @federicocodo, 

Could you share your Hazelcast .Net version, the code snippet that you execute the query and your mapping details of the map? 
Also, I don't recall that Hazelcast supports `IN` in SQL. You can see the supported functions [here](https://docs.hazelcast.com/hazelcast/5.3/sql/select).

federicocodo:
I'm using Hazelcast 5.3.0, the query i'm trying to execute it's the above one
```sql
SELECT *
FROM map
WHERE columnName IN (?)
```
The map it's saved in Hazelcast using the `HazelcastJsonValue`, I never had any problems until I tried to do the query and this it'snt a map problem.
I already looked and the documentation says that you can do queries using `IN`, [you can't just do subqueries only](https://docs.hazelcast.com/hazelcast/5.3/sql/functions-and-operators#in-operator).
I also opened the issue in the `Hazelcast` repository.

emreyigit:
I meant the type mappings of the object in the map. For example, if value is `HazelcastJsonValue`, then you can't reach to `columnName` directly since it is the part of json value. In order to reach a field from json, you should use json [functions](https://docs.hazelcast.com/hazelcast/5.3/sql/functions-and-operators#json-functions).  If you provide more information about type mappings and consuming part of the query, I can try to reproduce the problem.

federicocodo:
[...]

PR #872, opened 2024-01-19
[API-2160] Add Nullable Support for Value Types on SQL Queries

emreyigit:
If result of a SQL query contains `null` values, .Net client throws. See #854 for more details. This is because although SQL supports null for primitive types. The client cannot handle it. I've added nullable support for fixed sized type which are primitive ones. 

Also, current implementation were failing while parsing null or mixed result due to wrong assumptions on `List`. Defining a size during initialization, does not mean that the list created with null values. This is fixed in the PR.


zpqrtbnk (approved):
Approved

PR #875, opened 2024-01-29
Add .NET 8 support

zpqrtbnk:
This PR adds support for .NET 8 and updates dependencies accordingly.

NOTE: does *not* update Thrift as we *have* to remain compatible with the remote controller.
NOTE: does *not* update NUnit as moving from version 3 to 4 is maybe not trivial.


.github/actions/install/action.yml
```diff
@@ -53,6 +53,7 @@ runs:
           5.0.x
```
emreyigit, line 53: Should we also drop support of .Net 5? It's already dropped by Microsoft.

emreyigit (dismissed):

emreyigit (approved):

emreyigit (approved):

PR #886, opened 2024-03-15
Drop CLR `BinaryFormatter` Serialization Support on .NET8+ [API-2225]

emreyigit:
Hazelcast .Net Client drops support for CLR serialization due to `BinaryFormatter` is obsolete with .NET8. Previous .Net targets will continue to have `BinaryFormatter`. However, starting from .NET8 the support will be dropped. We strongly suggest to use alternatives. 

Please see https://learn.microsoft.com/en-us/dotnet/standard/serialization/binaryformatter-security-guide

The PR also disables the TPC tests since current implementation is beta version and doesn't match with latest Hazelcast. And, fixes other .Net8 breaking changes.


src/Hazelcast.Net.Examples/DistributedObjects/MapCustomSerializerExample.cs
```diff
@@ -25,6 +25,8 @@ public class MapCustomSerializerExample
     {
         public static async Task Main(string[] args)
         {
+            Console.WriteLine("NOTE: BinaryFormatter is obsolete and should not be used. This example is for demonstration purposes only.");
```
ihsandemir, line 28: Why use binaryformatter in the code sample if it is bad practice?

src/Hazelcast.Net.Tests/Configuration/ExceptionsTests.cs
```diff
@@ -33,12 +33,12 @@ public void ConfigurationExceptionConstructors()
             Assert.That(e.Message, Is.EqualTo("exception"));
             Assert.That(e.InnerException, Is.Not.Null);
             Assert.That(e.InnerException.Message, Is.EqualTo("bang"));
-
+#if !NET8_0_OR_GREATER
             e = e.SerializeAndDeSerialize();
```
ihsandemir, line 37: Is `ConfigurationException` Serializable? Is this a protocol defined exception?

src/Hazelcast.Net.Tests/Serialization/PortableSerializationTest.cs
```diff
@@ -397,7 +397,7 @@ public void TestSerializationService_CreatePortableReader()
             Assert.AreEqual(parent.timestamp, reader.ReadLong("child.timestamp"));
             Assert.AreEqual(child.timestamp, reader.ReadLong("child.child.timestamp"));
         }
-
+#if !NET8_0_OR_GREATER
```
ihsandemir, line 400: Does custom serialization not work for version 8 and above?

src/Hazelcast.Net/Serialization/InvalidPortableFieldException.cs
```diff
@@ -60,15 +62,18 @@ public InvalidPortableFieldException(string message, Exception innerException)
...
 
+#if !NET8_0_OR_GREATER
         /// <summary>
         /// Initializes a new instance of the <see cref="InvalidPortableFieldException"/> class with serialized data.
         /// </summary>
         /// <param name="info">The <see cref="SerializationInfo"/> that holds the serialized object data
         /// about the exception being thrown.</param>
         /// <param name="context">The <see cref="StreamingContext"/> that contains contextual information
         /// about the source or destination.</param>
+        [Obsolete("This constructor is obsolete due to BinaryFormatter being obsolete. Use the constructor without this parameter.")]
```
ihsandemir, line 73: why don't you totally remove this constructor also for .net 7 and lower?

src/Hazelcast.Net.Examples/DistributedObjects/MapCustomSerializerExample.cs
emreyigit: It works on other targets -> .net6, .net7, .net48, .net462

src/Hazelcast.Net.Tests/Configuration/ExceptionsTests.cs
emreyigit: It's a Hazelcast exception. The `SerializeAndDeSerialize()` is .Net binary serializer. There is nothing to do with serialization actually. Our exception has `[Serializable]` attribute but I'm not sure about the purposes. If it is designed for internal purposes there is no usage. If it is there for .net diagnostic reporting, it is deprecated.

src/Hazelcast.Net/Serialization/InvalidPortableFieldException.cs
emreyigit: It works on lower targets. I didn't want to break it suddenly.

ihsandemir: Is this public API?

emreyigit: Some of the exceptions are public. While refactoring, I didn't separate them.

PR #888, opened 2024-03-26
Use new Hazelcast logo

JackPGreen:
The logos used in the documentation are outdated - update to the latest versions.

I've tried testing locally with the [docfx quickstart](https://dotnet.github.io/docfx/) & `docfx doc\_docfx.json --serve`, but even on `master` cannot get the images to load locally at all to validate this fix, so have **not been able to test**.

Originally [reported via Slack](https://hazelcast.slack.com/archives/C035HQET5/p1711471674415239).


doc/templates/hz/cover-index.html
```diff
@@ -51,7 +51,7 @@
               </button>
               
               <a class="navbar-brand" href="../../index.html">
-                <img id="logo" src="../../images/hazelcast-white.png" alt="">
+                <img id="logo" src="https://hazelcast.com/files/brand-images/logo/hazelcast-logo.png" alt="">
```
emreyigit, line 54: Is there a white logo? It doesn't look good on a dark background. Also, I'm not sure having an external link to logo. The logo can change and it may not fit like size, color etc.

emreyigit:
Btw, we cannot update the nuget page without having a release. So, it's better to wait for 5.4.0 release. However, I'll look for the documentation page.

emreyigit:
Updated the logo on docs, fyi. https://hazelcast.github.io/hazelcast-csharp-client/index.html

doc/templates/hz/cover-index.html
JackPGreen: > Is there a white logo? It doesn't look good on a dark background.

https://hazelcast.com/files/brand-images/logo/hazelcast-logo-reverse.png

> Also, I'm not sure having an external link to logo. The logo can change and it may not fit like size, color etc.

I was trying to avoid hardcoded images going out of date, but to be fair the URLs don't look permanent enough to support that

emreyigit: Could you change it back to hard coded than?

JackPGreen: > Could you change it back to hard coded than?

I can replace external links with local copies of those images if that's what you mean?

emreyigit: Yes, please.

JackPGreen: > > Could you change it back to hard coded than?
> 
> I can replace external links with local copies of those images if that's what you mean?



> Yes, please.

7107e351ddd488db442dbed3b7dc5cf9882f59ae

emreyigit (approved):

PR #890, opened 2024-04-26
#889 Only include system.memory on framework

### LINKED ISSUE #889: Optimise dependencies to leverage framework

thompson-tomo:
Closes #889

### DIFF
--- a/src/Hazelcast.Net.Tests/DotNet/MemoryTests.cs
+++ b/src/Hazelcast.Net.Tests/DotNet/MemoryTests.cs
@@ -1,11 +1,11 @@
 ﻿// Copyright (c) 2008-2024, Hazelcast, Inc. All Rights Reserved.
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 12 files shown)


emreyigit:
Hi @thompson-tomo, the client doesn't support framework 4.5. In this case, we can't merge this PR. 
https://hazelcast.github.io/hazelcast-csharp-client/versions.html

thompson-tomo:
sorry, i think i got confused looking at the conditional compilation hence have gone and cleaned up the conditional compilation and simply dropped the dependency all together.

emreyigit:
Closing the issue since cannot accept the changes on the repository due to differences between .net2.0 and .net framework 4.x

thompson-tomo:
@emreyigit i am little bit confused by your comment below,

> due to differences between .net2.0 and .net framework 4.x

I cleaned up the conditional compilation statements to no longer have checks for frameworks which are not being targeted. Hence would not be changing the output. Note even nuget.org shows net framework is not being targeted.

PR #891, opened 2024-05-13
Clear Members Before Failover Over to New Cluster [API-2217]

### LINKED ISSUE #834: [TRACKING ISSUE] Do not try to connect to the old member list after the cluster changes [API-2038] [API-2217]
The tracking issue for the Java side PR.

See https://github.com/hazelcast/hazelcast/pull/24745 for details.

---

During failover, when the client is about to try to connect to the next cluster, it resets some services including the cluster service.

In its reset logic, cluster service clears the member list version, but leaves the members as it is, as there are some services/proxies that rely on the fact that the member list cannot be empty.

This creates a problem for the failover case. After the cluster change, the client still tries to connect to the last known member list, which is the members of the previous cluster.

To solve the problem, I have introduced a new method to the cluster service, which returns the "effective" member list. That is, it returns an empty list after the service is reset, but returns the member list as expected otherwise. With that, the connection logic simply skips the old member list after the client decides to change the cluster to try to connect to.

emreyigit:
Fixes #834.

### DIFF
--- a/src/Hazelcast.Net/Clustering/ClusterMembers.cs
+++ b/src/Hazelcast.Net/Clustering/ClusterMembers.cs
@@ -33,7 +33,7 @@ namespace Hazelcast.Clustering
     internal class ClusterMembers : IAsyncDisposable
     {
         private const int SqlConnectionRandomAttempts = 10;
-
+        private const int InvalidMemberTableVersion = -1;
         private readonly object _mutex = new();
         private readonly ClusterState _clusterState;
         private readonly ILogger _logger;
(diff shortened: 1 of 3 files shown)


src/Hazelcast.Net/Clustering/ClusterMembers.cs
```diff
@@ -80,6 +80,12 @@ public ClusterMembers(ClusterState clusterState, TerminateConnections terminateC
                     lock (_mutex) return _members.ContainsMember(x);
                 }, _clusterState.LoggerFactory);
             }
+
+            _clusterState.Failover.ClusterChanged += cluster =>
+            {
+                // Clear old member table since failover happened.
+                _members = new MemberTable();
```
ihsandemir, line 87: Member removed events may be missing during cluster switch, let's have a look at this.

emreyigit: Great catch! I've covered that part at https://github.com/hazelcast/hazelcast-csharp-client/pull/891/commits/285b613d5b6de629b9ac9d086637dc8d6a839e98

ihsandemir (approved):
May be we can also open an issue to write a test about MembershipListener works as expected during failover. This test may be missing also for Java. We can add to both clients. It can be a separate PR, not urgent.

PR #901, opened 2024-06-26
Introduce New Routing Mode Options [API-2240] (2/4)

emreyigit:
The PR introduces the new routing mode options and refactors the old smart mode usage in the code. 

This PR is part of new routing implementations. And built on #899. So, it must be merged after #899


doc/dev/doc/network.md
```diff
@@ -13,18 +13,37 @@ You can specify multiple addresses, with or without the port information. If the
 By default, the provided list is shuffled and tried in a random order. You can disable this behaviour by setting the configuration option `hazelcast.networking.shuffleAddresses` to `false`. In this case the address list will be tried in the specified order. 
 
 ## Operation Mode
+The client has three operation modes because of the distributed nature of the data and cluster: all members, multi members and single member. All members routing is enabled by default, and is controlled by the `@Hazelcast.NetworkingOptions.RoutingMode` configuration option.
 
-The client has two operation modes because of the distributed nature of the data and cluster: smart and unisocket. Smart routing is enabled by default, and is controlled by the `hazelcast.networking.smartRouting` configuration option, but you may want to enable unisocket mode.
+> [!NOTE]
+> Smart Routing is deprecated and will be removed in the next major release. Please use the new RoutingMode mode instead.
```
gbarnett-hz, line 19: When I read that it sounded like the semantics was deprecated -- this is in reference to `hazelcast.networking.smartRouting` and not the smart routing semantics, that's just been renamed. Maybe clarify it a bit?

src/Hazelcast.Net.Tests/Clustering/FailoverTests.cs
```diff
@@ -286,7 +287,7 @@ public async Task TestClientCanFailover(bool smartRouting, int memberCount)
                         o.Networking.Addresses.Clear();
                         o.Networking.Addresses.Add(Cluster0Address);
                         o.Networking.ReconnectMode = Hazelcast.Networking.ReconnectMode.ReconnectAsync;
-                        o.Networking.SmartRouting = smartRouting;
+                        o.Networking.RoutingMode.Mode = smartRouting ? RoutingModes.AllMembers : RoutingModes.SingleMember;
```
gbarnett-hz, line 290: `smartRouting ? RoutingModes.AllMembers : RoutingModes.SingleMember;` 

I see this logic a few places -- maybe use some helper for it? I understand why you're doing it w.r.t. old logic but new reader without the diff won't know.

gbarnett-hz (commented):
LGTM -- I skipped the bits my eyes remembered from #899. @emreyigit could you ping me again after #899 is merged so I can take another look?

doc/dev/doc/network.md
emreyigit: I see, you are right. I've paraphrased it. How about now?

src/Hazelcast.Net.Tests/Clustering/FailoverTests.cs
emreyigit: I've refactored the smart routing argument to `RoutingModes` directly, no middle transformation anymore.

gbarnett-hz (dismissed):
LGTM -- like #899 I've done on best effort.

doc/dev/doc/failover.md
```diff
@@ -42,7 +42,7 @@ Let's assume both clusters are down, and we just initialize the client. Client w
...
 
-**(3)** Alternative cluster options. Alternative options are only for connection and authentication. Alternative client options cannot override other fields. General configuration should be done in first client options which is step **1**. For example, here **3**, we cannot change the `SmartRouting` for each cluster. Also, for server side, provided cluster must have **same partition count**. Client cannot failover to a cluster with different partition count.
+**(3)** Alternative cluster options. Alternative options are only for connection and authentication. Alternative client options cannot override other fields. General configuration should be done in first client options which is step **1**. For example, here **3**, we cannot change the `~~~~RoutingMode` for each cluster. Also, for server side, provided cluster must have **same partition count**. Client cannot failover to a cluster with different partition count.
```
JamesHazelcast, line 45:
```suggestion
**(3)** Alternative cluster options. Alternative options are only for connection and authentication. Alternative client options cannot override other fields. General configuration should be done in first client options which is step **1**. For example, here **3**, we cannot change the `~~~~RoutingMode` for each cluster. Also, for server side, provided cluster must have **same partition count**. Client cannot failover to a cluster with different partition count.
```
```suggestion
**(3)** Alternative cluster options. Alternative options are only for connection and authentication. Alternative client options cannot override other fields. General configuration should be done in first client options which is step **1**. For example, here **3**, we cannot change the `RoutingMode` for each cluster. Also, for server side, provided cluster must have **same partition count**. Client cannot failover to a cluster with different partition count.
```
Not sure what purpose the `~~~~` is serving here?

doc/dev/doc/network.md
```diff
@@ -13,18 +13,37 @@ You can specify multiple addresses, with or without the port information. If the
...
+### Routing Mode
 
-### Smart Routing
+Routing mode allows you to choose the routing strategy of the client. 
 
-In the smart mode, the clients connect to each cluster member. Since each data partition uses the well known and consistent hashing algorithm, each client can send an operation to the relevant cluster member, which increases the overall throughput and efficiency. Smart mode is the default mode.
+There are three routing modes:
+* `@Hazelcast.Networking.RoutingModes.SingleMember` aka Unisocket mode.
+* `@Hazelcast.Networking.RoutingModes.MultiMember` connects to subset of the cluster.
+* `@Hazelcast.Networking.RoutingModes.AllMembers` connects to all members in the cluster aka Smart mode.
```
JamesHazelcast, line 29:
```suggestion
* `@Hazelcast.Networking.RoutingModes.SingleMember` connects to a single member of the cluster (previously Unisocket routing).
* `@Hazelcast.Networking.RoutingModes.MultiMember` connects to subset of members in the cluster.
* `@Hazelcast.Networking.RoutingModes.AllMembers` connects to all members of the cluster (previously Smart routing).
```

doc/dev/doc/network.md
```diff
@@ -13,18 +13,37 @@ You can specify multiple addresses, with or without the port information. If the
...
 
 In the unisocket client mode, the client will only connect to one of the configured addresses. This single member will behave as a gateway to the other members. For any operation requested from the client, it will redirect the request to the relevant member and return the response back to the client returned from this member.
+#### Multi Member Routing Mode
+Multi member routing mode is a mode where the client connects to a subset of the cluster members. This mode behaves based
+on given strategy on `@Hazecast.Networking.NetworkingOptions.RoutingMode.Strategy`.
+
+Available strategies are:
+* `@Hazelcast.Networking.RoutingStrategy.PartitionGroups` which is the default strategy. Clients connects to subset of the cluster based on partition groups which contains the current connected member. 
+
+Please note that server should support this routing mode which is available with Hazelcast 5.5.0.
```
JamesHazelcast, line 43:
```suggestion
Please note that server should support this routing mode which is available with Hazelcast Enterprise 5.5.0.
```

doc/dev/doc/network.md
```diff
@@ -13,18 +13,37 @@ You can specify multiple addresses, with or without the port information. If the
...
+
+#### All Members Routing Mode
+In all members mode, the clients connect to each cluster member. Since each data partition uses the well known and consistent hashing algorithm, each client can send an operation to the relevant cluster member, which increases the overall throughput and efficiency. Smart mode is the default mode.
```
JamesHazelcast, line 46:
```suggestion
In all members mode, the clients connect to each cluster member. Since each data partition uses the well known and consistent hashing algorithm, each client can send an operation to the relevant cluster member, which increases the overall throughput and efficiency. This mode is the default mode.
```

src/Hazelcast.Net/Clustering/ClusterConnections.cs
```diff
@@ -900,6 +917,7 @@ private async Task<MemberConnection> ConnectAsync(NetworkAddress address, Networ
             if (!connection.Active) await ThrowDisconnected(connection).CfAwait();
             if (!accepted) await ThrowRejected(connection).CfAwait();
 
+
```
JamesHazelcast, line 920: Nit
```suggestion
```

src/Hazelcast.Net/Clustering/ClusterState.cs
```diff
@@ -429,7 +431,17 @@ public void RequestShutdown()
         /// <summary>
         /// Whether smart routing is enabled.
         /// </summary>
-        public bool IsSmartRouting => Options.Networking.SmartRouting;
+        public bool IsSmartRouting => RoutingMode == RoutingModes.AllMembers;
```
JamesHazelcast, line 434: The Java client has dropped all use of "smart routing" except in the client config option, which is maintained for backwards compatibility only - I'd suggest we do the same for C#

This applies to other areas of the PR too, but I will only mention it here.

src/Hazelcast.Net/Exceptions/InvalidPartitionGroupException.cs
```diff
@@ -0,0 +1,48 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+using System;
+namespace Hazelcast.Exceptions
+{
+    /// <summary>
+    /// Represents the exception that is thrown when invalid partition group is set.
+    /// </summary>
+    public sealed class InvalidPartitionGroupException :HazelcastException
```
JamesHazelcast, line 20:
```suggestion
    public sealed class InvalidPartitionGroupException : HazelcastException
```
Nit

src/Hazelcast.Net/Networking/NetworkingOptions.cs
```diff
@@ -91,6 +92,9 @@ private NetworkingOptions(NetworkingOptions other)
     /// <para>Note that it however does not guarantee that the operation will always be
     /// executed on the owner, as the member table is only updated every 10 seconds.</para>
     /// </remarks>
+    [Obsolete("This option has no effect, use RoutingMode instead.", false)]
+    [ExcludeFromCodeCoverage]
+    [EditorBrowsable(EditorBrowsableState.Never)]
     public bool SmartRouting { get; set; } = true;
```
JamesHazelcast, line 98: We need to support the "smart routing" config option for the foreseeable future as some of our customers will not be able to update their client configs easily and we don't want to introduce a breaking change.

In the Java client, we do this by overriding `setSmartRouting(boolean)` to set our `RoutingMode` based on the `boolean` provided.

JamesHazelcast (approved):
While not a C# expert by any means, I've checked over the business logic and it looks solid to me - a few comments to address, but almost all are minor, so approving in advance 👍 Thanks for working on this Emre!

src/Hazelcast.Net/Exceptions/InvalidPartitionGroupException.cs
emreyigit: Fixed.

doc/dev/doc/failover.md
emreyigit: No idea on where it came from. Good catch. Fixed

doc/dev/doc/network.md
emreyigit: Fixed.

doc/dev/doc/network.md
emreyigit: Fixed.

doc/dev/doc/network.md
emreyigit: Fixed.

src/Hazelcast.Net/Clustering/ClusterConnections.cs
emreyigit: Fixed.

src/Hazelcast.Net/Clustering/ClusterState.cs
emreyigit: Fair enough. Refactored that smart routing stays same by setting the `RoutingMode`

src/Hazelcast.Net/Networking/NetworkingOptions.cs
emreyigit: As I mentioned above, smart routing stays same by setting the `RoutingMode` as in Java client.

JamesHazelcast (approved):
Business logic LGTM, thanks for working on this @emreyigit!

PR #902, opened 2024-06-28
Introduce Multi Member Routing Mode [API-2241] (3/4)

emreyigit:
Implemented new `MultiMember` routing functionality and refactored the smart routing and unisocket to `AllMembers` and `SingleMember`.

_A note for reviewers:_ Although I've completed implementation including CP routing, I've created separated PR for Cp part. You may see some basic CP stuff in this PR but most part is in next PR #907


src/Hazelcast.Net/CP/ClusterCPGroups.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// <summary>
+        /// Overrides current leader mappings with the given ones.
+        /// Suitable for first time leader mappings.
+        /// </summary>
+        /// <param name="cpGroupLeaders">Mapped CP Group Ids and their leaders</param>
+        public void SetCPGroupIds(IDictionary<CPGroupId, Guid> cpGroupLeaders)
+        {
+            lock (_mutex)
+            {
+                if(cpGroupLeaders.Count == 0) return;
```
JamesHazelcast, line 119: Is there a scenario where you expect `Count == 0`? Leadership updates are only sent by group leaders, so they should always contain at least 1 group, but I'm curious if you found a scenario where this is possible.

src/Hazelcast.Net/Clustering/ClusterConnections.cs
```diff
@@ -835,6 +835,9 @@ private async Task<MemberConnection> ConnectAsync(NetworkAddress address, Networ
             // connect to the server (may throw and that is ok here)
             var result = await connection.ConnectAsync(_clusterState, cancellationToken).CfAwait();
 
+            // we now that fail over is an ee feature.
+            _clusterState.IsEnterprise = result.FailoverSupported;
```
JamesHazelcast, line 839: In the Java client, we send servers whether or not we are EE [as a client attribute](https://github.com/hazelcast/hazelcast-mono/blob/master/hazelcast/hazelcast/src/main/java/com/hazelcast/client/impl/ClientEndpointImpl.java#L169), and internally we are also [able to query `BuildInfo` to check if we are EE](https://github.com/hazelcast/hazelcast-mono/blob/master/hazelcast/hazelcast/src/main/java/com/hazelcast/client/impl/clientside/DefaultClientExtension.java#L200) - is there nothing similar on .NET client?

src/Hazelcast.Net/Clustering/ClusterEvents.cs
```diff
@@ -623,21 +696,28 @@ private async ValueTask HandleCodecPartitionViewEvent(int version, IList<KeyValu
             // On... does not throw
             await _partitionsUpdated.AwaitEach().CfAwait();
         }
-        
-        private  ValueTask HandleCodecClusterVersionEvent(ClusterVersion version, object state)
+
+        private ValueTask HandleCodecClusterVersionEvent(ClusterVersion version, object state)
         {
-            _logger.LogDebug("Handle ClusterVersion event.");
+            _logger.IfDebug()?.LogDebug("Handle ClusterVersion event");
```
JamesHazelcast, line 702: Applies to several places in this file, but I'd like to see more information provided in these debugging logs - their primary purpose would be to provide information about a production issue, and the current messages provide very little to work with

src/Hazelcast.Net/Models/AuthenticationResult.cs
```diff
@@ -129,5 +132,8 @@ internal sealed class AuthenticationResult
         /// Cluster version. Read from KeyValuePairs["cluster.version"].
         /// </summary>
         public ClusterVersion ClusterVersion { get; }
+
+        /// <summary>
```
JamesHazelcast, line 136: Dangling doc?

src/Hazelcast.Net/CP/CPDistributedObjectBase.cs
```diff
@@ -66,7 +72,33 @@ protected ValueTask<TObject> ToObjectAsync<TObject>(IData valueData)
...
+
+                if (leader != null)
+                {
+                    return Cluster.Messaging.SendToMemberAsync(request, leader.Uuid);
+                }
+            }
+            else
+            {
+                if (Cluster.State.Options.Networking.CPDirectToLeaderEnabled)
+                    Logger.IfWarning()?.LogWarning("CP Direct To Leader is an Enterprise feature, the request will be sent to any member");
```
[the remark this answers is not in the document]
emreyigit: I'm afraid there is no proper way at the moment because there is config check during starting and even I check it on authentication that exceptions doesn't bubble up to user level. We need a moderate refactoring to have that.

src/Hazelcast.Net/CP/ClusterCPGroups.cs
emreyigit: Actually, I used a empty list or object instead of null. You can consider it as a null or emptiness check.

src/Hazelcast.Net/Clustering/ClusterConnections.cs
emreyigit: Yes, in Java client there are options. However, except Java, other clients have no EE version. This is the best effort for EE check.

src/Hazelcast.Net/Clustering/ClusterEvents.cs
emreyigit: Sure, fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/902/commits/b246b6c0752b52869313042c36feb40935239ed6

src/Hazelcast.Net/Models/AuthenticationResult.cs
emreyigit: Yes, it is. Fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/902/commits/b246b6c0752b52869313042c36feb40935239ed6

src/Hazelcast.Net/CP/ClusterCPGroups.cs
```diff
@@ -0,0 +1,131 @@
...
+                _groups.Clear();
+                foreach (var g in newGroup)
+                {
+                    _groups[g.Key] = g.Value;
+                }
+
+            }
+
+            _logger.IfDebug()?.LogDebug("CP groups updated to Version {Version}." +
+                                        " Old groups count: {OldCount}, new count: {NewCount}", Version, oldCount, groups.Count);
```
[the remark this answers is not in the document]
gbarnett-hz: Is it a frequent log message? If it's really infrequent then it might be worth bumping to `INFO` as nobody is running with `DEBUG`. If it's frequent then I would leave as-is.

src/Hazelcast.Net/Clustering/Authenticator.cs
```diff
@@ -268,4 +272,43 @@ internal MemberGroups ParsePartitionMemberGroups(ClientAuthenticationCodec.Respo
...
+                var name = cpGroupJson["name"].GetValue<string>();
+                var leaderId = group["leaderUUID"].GetValue<Guid>();
+
+                var cpGroupId = new CPGroupId(name, seed, id);
+                cpGroupLeaderIds[cpGroupId] = leaderId;
+            }
+        }
+        catch (Exception e)
+        {
+            _logger.IfDebug()?.LogDebug("Failed to parse CP group leader ids [{JsonMessage}]: {Message}. ", jsonMessage, e.Message);
```
gbarnett-hz, line 308: Should this not be > `DEBUG`? We control the structure of the message and its content -- if it doesn't have what we expect then it's an `ERROR`?

gbarnett-hz (approved):

JamesHazelcast (approved):
Thanks @emreyigit!

PR #908, opened 2024-07-11
Make `git` branch resolution in script more more tolerant

JackPGreen:
The script queries if the branch name starts with `release`, using the `git symbolic-ref` command.

[While testing the `client-compatibility-suites`](https://github.com/hazelcast/client-compatibility-suites/actions/runs/9908837208/job/27376177353), it was noticed that for detached branches, [this command throws an error (`fatal: ref HEAD is not a symbolic ref`) and stops the script](https://stackoverflow.com/a/45028375).

### DIFF
--- a/hz.ps1
+++ b/hz.ps1
@@ -1541,7 +1541,12 @@ function hz-build {
         $options.constants = $options.constants.Replace(";", "%3B") # escape ';'
     }
 
-    $branchName = git name-rev --name-only HEAD
+    try {
+        $branchName = git symbolic-ref --short HEAD
+    } catch {
+        # In the case of a detached branch
+        $branchName = ""
+    }
     $isReleaseBranch = $branchName.StartsWith("release/")
 
     Write-Output "Build"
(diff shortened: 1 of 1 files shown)


hz.ps1
```diff
@@ -1541,7 +1541,7 @@ function hz-build {
         $options.constants = $options.constants.Replace(";", "%3B") # escape ';'
     }
 
-    $branchName = git symbolic-ref --short HEAD
+    $branchName = git name-rev --name-only HEAD
```
emreyigit, line 1544: It doesn't produce same result. Instead, could you add `-q` to suppress the error?

JackPGreen: > Instead, could you add `-q` to suppress the error?
No, because it still sets the error level / exit code to `1`.

emreyigit: Well, this won't work. `name-rev` doesn't give the full refs path which includes the _release_ directory.

JackPGreen: > Well, this won't work. `name-rev` doesn't give the full refs path which includes the _release_ directory.

Ok, good point.

b72e3a9de98017e484299a293e8dc35318b37413

PR #909, opened 2024-07-12
Make `ensure-server-files` use `hazelcast-mono`

JackPGreen:
`ensure-server-files` downloads `hazelcast-default.xml` from the `hazelcast` repo.

[This failed](https://github.com/hazelcast/client-compatibility-suites/actions/runs/9882576769/job/27296050800), now that we have EE specific patches, where corresponding tags do not exist in `hazelcast-mono`:
> Downloading hazelcast-default.xml -> hazelcast-5.4.1-SNAPSHOT.xml...
> Failed to download hazelcast-default.xml (404) from tag v5.4.1

Updated script to:
- Use `hazelcast-mono`, not `hazelcast` as source
- Centralise downloading into a single function

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -156,6 +156,7 @@ jobs:
       - name: Test
         id: test
         env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           HAZELCAST_ENTERPRISE_KEY: ${{ secrets.HAZELCAST_ENTERPRISE_KEY }}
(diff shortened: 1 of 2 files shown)


emreyigit:
I see the changes but this is a public repo. How user will run it?

JackPGreen:
> I see the changes but this is a public repo. How user will run it?

They won't be able to - but isn't that where we already are with private `SNAPSHOT` repositories etc?

You _could_ make it check `hazelcast` first, and if it's not found _then_ check `hazelcast-mono`

emreyigit (approved):

srknzl:
one build failed, (windows-latest+net8.0) I restarted it

PR #913, opened 2024-08-06
Fix CP Group Name in CP Objects [API-2252]

emreyigit:
The PR fixes wrong usage of CP groups. Before the introduced fix, the client was sending the CP group creation requests in wrong format. This was resulting that all CP objects created in `default` CP group. Now, it's fixed.

### DIFF
--- a/src/Hazelcast.Net.Tests/CP/CPRoutingTests.cs
+++ b/src/Hazelcast.Net.Tests/CP/CPRoutingTests.cs
@@ -40,6 +40,8 @@ namespace Hazelcast.Tests.CP
         [Test]
         public async Task TestCPRequestRoutingToLeader()
         {
+            const string groupName = "myGroup";
+            const string cpObjectName = "myAtomicLong";
             var options = new HazelcastOptionsBuilder()
                 .With((config =>
                 {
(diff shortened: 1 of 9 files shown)


Patras3 (approved):
LGTM, thanks for the fix.

src/Hazelcast.Net.Tests/CP/CPRoutingTests.cs
```diff
@@ -51,49 +53,63 @@ public async Task TestCPRequestRoutingToLeader()
                     config.Networking.ConnectionRetry.ClusterConnectionTimeoutMilliseconds = 10_000;
                 })).Build();
 
+
+            // Assert multiple that the addAndGet operation is sent to the leader
+            // First round creates new group
+            // Second round uses the existing group
+            for (var i = 0; i < 2; i++)
```
ihsandemir, line 60: is this related to the change?

src/Hazelcast.Net.Tests/Resources/Cluster/cp.xml
```diff
@@ -43,8 +43,7 @@
 
   <cp-subsystem>
     <base-dir>/custom-cp-dir</base-dir>
-    <cp-member-count>3</cp-member-count>
-    <group-size>3</group-size>
```
ihsandemir, line 47: why removed?

src/Hazelcast.Net.Tests/CP/CPRoutingTests.cs
emreyigit: Yes.

src/Hazelcast.Net.Tests/Resources/Cluster/cp.xml
emreyigit: The default group size is already equals to cp member count at default.

ihsandemir (approved):

PR #931, opened 2024-11-20
Update IHMap.Setting.cs

rejdzu:
Fix order of parameters on interface compared to implementation

Issue: Interface of the HMap has different order of parameters than the implementation, which is causing confusion and wrong behavior


emreyigit:
Hi @rejdzu, thank you for your contribution ❤️. Although it's a nice touch, I can't merge it know because it will break backward compatibility. Also, the implementation should be refactored accordingly. I've put it to backlog. Whenever there is chance to break backward compatibility, I'll process it ✋

PR #933, opened 2025-02-07
Fix Missing Patch Version on Client Version Reporting [API-2245][HZG-327]

emreyigit:
- Updated ClientVersion class to handle version parsing more robustly.
- Added methods to retrieve the major, minor, and patch versions separately.


src/Hazelcast.Net/Core/ClientVersion.cs
```diff
@@ -23,7 +23,7 @@ namespace Hazelcast.Core
     internal static class ClientVersion
     {
         private static string _clientVersion;
-        private static string _clientVersionPure;
+        private static string _clientMajorMinorPatchVersion;
```
ihsandemir, line 26: For this class, should we rather port the corresponding Java class [Version.java](https://github.com/hazelcast/hazelcast-mono/blob/master/hazelcast/hazelcast/src/main/java/com/hazelcast/version/Version.java) so that the two code base uses same logics?

src/Hazelcast.Net/Core/ClientVersion.cs
```diff
@@ -59,18 +59,38 @@ internal static string GetVersion(AssemblyInformationalVersionAttribute informat
...
+            var pos2 = -1;
+            
+            // Check the patch version segment, expecting max 2 digits
+            if (pos1 + 1 < version.Length && int.TryParse(version[pos1 + 1] + "", out _))
             {
-                // two dots = major.minor.whatever
-                // take everything until the second '.' = major.minor
+                pos2 = pos1 + 1; 
+
+                if (pos2 + 1 < version.Length && int.TryParse(version[pos2 + 1] + "", out _))
```
ihsandemir, line 77: do you have unit tests to validate the changes?

emreyigit: Yes, please see the `ClientVersionReturnsMajorMinorVersion` in the PR.

src/Hazelcast.Net/Core/ClientVersion.cs
emreyigit: I have used the canonical form of the client version. Which prints `5.5.0+1ce4f2`. Is that OK for you?

ihsandemir: should we not include the commit id to be similar to what java client reports?

emreyigit: Java reports the whole version info except the building part, such as `6.0.0-SNAPSHOT`. I've refactor based on this. You can see that Authenticator and ClientMetricSource sends the client version similar to Java client.

src/Hazelcast.Net.Tests/Core/ClientVersionTests.cs
```diff
@@ -50,15 +51,32 @@ public void ClientVersionReturnsVersion()
...
+
+            Assert.That(Authenticator.ClientVersion, Is.EqualTo(ClientVersion.MajorMinorPatchVersion));
+            var assemblyVersion = typeof(ClientVersion).Assembly.GetName().Version;
+            Assert.That(Authenticator.ClientVersion, Is.EqualTo(assemblyVersion.Major + "." + assemblyVersion.Minor + "." + assemblyVersion.MajorRevision));
+            Assert.That(Authenticator.ClientVersion.Split('.').Length, Is.GreaterThanOrEqualTo(3));
+        }
+
+
+        [TestCase("1.2.3", "1.2.3")]
+        [TestCase("1.2.3-preview.0", "1.2.3")]
```
ihsandemir, line 64: You commented below java reports as 6.0.0-SNAPSHOT. In this case, I was expecting to see this test expects to be `1.2.3-preview.0`

Can you also add the `6.0.0-SNAPSHOT` test case?

emreyigit: This is only for `ClientVersion.GetMajorMinorPatchVersion`. I've introduced another method `ClientVersion.GetSemVerWithoutBuildingMetadata` which has tests below. Also, added the case.

PR #934, opened 2025-02-07
Fix `RoutingModes` Enum Values [API-2266]

emreyigit:
Introduced a fix to sync enum values with Java implementation. Hence, connection mode the reporting to the cluster is corrected.

### DIFF
--- a/src/Hazelcast.Net/Networking/RoutingModes.cs
+++ b/src/Hazelcast.Net/Networking/RoutingModes.cs
@@ -24,11 +24,6 @@ namespace Hazelcast.Networking
         SingleMember = 0,
 
         /// <summary>
-        /// Connects only to selected group of members in the cluster based on selected grouping strategy.
-        /// </summary>
-        MultiMember = 1,
-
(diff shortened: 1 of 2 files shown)


src/Hazelcast.Net/Networking/RoutingModes.cs
```diff
@@ -36,7 +31,12 @@ public enum RoutingModes : byte
         /// <para>Note that it however does not guarantee that the operation will always be
         /// executed on the owner, as the member table is only updated every 10 seconds.</para>
         /// </summary>
-        AllMembers = 2
+        AllMembers = 1,
```
ihsandemir, line 34: is the existing release broken due to this mismatch? Will the fix be backported?

emreyigit: Yes, this was the reason. I'm not sure about the backport. Since it's a cosmetic issue, another release may not be needed.

ihsandemir (approved):

PR #956, opened 2025-03-14
Introduce Custom Build Analyzer for Hazelcast .Net Client

emreyigit:
I believe we need a custom analyzers for more consistent development quality. Introduced the `Hazelcast.Net.BuildAnalyzers` with an initial rule. The project is an internal tool. It runs on compile time and not part of the release. Currently, disabled the analyzer since the first rule which checks the all public APIs whether they have `NotNull` or `MayBeNull` attributes. Since the attributes are not used widely enough, triggers a lot of warnings. When I find time, I'll fix the warnings with another PR.


gbarnett-hz (approved):
LGTM - though I would hold off merging until it's actually used.

What's the empty `AnalyzerReleases.Unshipped.md` for?

src/Hazelcast.Net.BuildAnalyzers/PublicApiNullAttributeAnalyzer.cs
```diff
@@ -0,0 +1,101 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Hazelcast.Net.BuildAnalyzers
+{
+    /// <summary>
+    /// Checks that all public APIs have either NotNull or MaybeNull parameter. 
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class PublicApiNullAttributeAnalyzer : DiagnosticAnalyzer
```
ihsandemir, line 28: Does Microsoft release any builtin analyzers in .NET framework which we can re-use rather than doing a custom one?

emreyigit:
Thanks @gbarnett-hz. It's used for version logging. The analyzer dependency requires these files before making the rule available. It's a kind a check point before making the rules available and for version history. The new added rule is reflected to the unshipped file and moved to shipped file manually. 

Also, [similar approach](https://github.com/hazelcast/hazelcast-csharp-client/tree/master/src/Hazelcast.Net/PublicAPI) exists on the client as well. Public APIs are inserted to unshipped file by the IDE and it prevents the release if they are not moved to shipped file manually. It's a nice reminder checking the public APIs one last time.

The rule is disabled for now, so that we can merge the PR, IMO. We can enable it when we have resource to drain warnings. Otherwise, opening PR diverges from master by time :|

src/Hazelcast.Net.BuildAnalyzers/PublicApiNullAttributeAnalyzer.cs
emreyigit: I couldn't see such rule unfortunately.

PR #997, opened 2025-11-05
Implement `UpdateTimeToLive` [HZ-5249][API-1789]

emreyigit:
Implemented missing `Map.UpdateTimeToLive` API.

### DIFF
--- a/src/Hazelcast.Net/DistributedObjects/IHMap.Setting.cs
+++ b/src/Hazelcast.Net/DistributedObjects/IHMap.Setting.cs
@@ -250,8 +250,8 @@ namespace Hazelcast.DistributedObjects
         /// <remarks>
         /// <para>The value is automatically expired, evicted and removed after the
         /// <paramref name="timeToLive"/> has elapsed. If <paramref name="timeToLive"/> is infinite (0ms),
-        /// the value is retained indefinitely. If it is -1ms,
-        /// it lives for the duration of the server-configured time-to-live.</para>
+        /// the value is retained indefinitely. If it is negative,
+        /// it lives for the duration of the server-configured (default: forever) time-to-live.</para>
         /// <para>The new time-to-live value is valid starting from the time this operation is invoked,
         /// not since the time the entry was created.</para>
         /// <para>


src/Hazelcast.Net.Tests/Remote/ClientMapTest.cs
```diff
@@ -1407,5 +1408,42 @@ public async Task TestValuesPredicate()
...
+
+            await latch.WaitOneAsync();
+            starTime.Stop();
+            var elapsed = starTime.ElapsedMilliseconds;
+
+            // Definitively, less then maxTTL
+            Assert.Less(elapsed, maxTTL * 1000);
+
+            // Close to min TTL
+            Assert.Less(elapsed, minTTL * 1000 * 2);
```
ihsandemir: i am not sure if this will always work if the server is working slow, etc.
Does the ContainsKeyAsync trigger eviction check on expiry, if so, it may be safe but if we do it on scheduled basis on the server side, there may be a time diff until that scheduler is triggered and entry is evicted. If that is the case, this check may fail depending on the scheduler interval.

emreyigit: Right, it may not work well in slow cases. Do you have any suggestion?

emreyigit: Removed the assertion at https://github.com/hazelcast/hazelcast-csharp-client/pull/997/commits/147b40ab53b751bef4acdc003c0d22faf87dad87 There is a public api coeverage work upcoming. I don't know the tests but I'll look for more according to results of that work.

src/Hazelcast.Net/DistributedObjects/Impl/HMap.Setting.cs
```diff
@@ -392,9 +392,15 @@ Task SetTransientAsync(IData keyData, IData valueData, TimeSpan timeToLive, Time
 #endif
         }
 
-        public Task<bool> UpdateTimeToLive(TKey key, TimeSpan timeToLive)
+        public async Task<bool> UpdateTimeToLive(TKey key, TimeSpan timeToLive)
         {
-            throw new NotImplementedException();
+            var keyData = ToSafeData(key);
```
ihsandemir: At java the key and TimSpan are checked for null? will you add the checks?

emreyigit: It's a struct cannot be null.

src/Hazelcast.Net.Tests/Remote/ClientMapTest.cs
```diff
@@ -1407,5 +1408,39 @@ public async Task TestValuesPredicate()
             Assert.IsTrue(enumerator.MoveNext());
             Assert.AreEqual("value1", enumerator.Current);
         }
+
```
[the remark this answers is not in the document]
emreyigit: Added https://github.com/hazelcast/hazelcast-csharp-client/pull/997/commits/b20446198d9edf5539c2ebf336786398a31aa63f

src/Hazelcast.Net/DistributedObjects/Impl/HMap.Setting.cs
```diff
@@ -392,9 +392,15 @@ Task SetTransientAsync(IData keyData, IData valueData, TimeSpan timeToLive, Time
...
+            var seconds = Convert.ToInt64(timeToLive.TotalSeconds);
```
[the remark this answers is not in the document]
emreyigit: Yes, fixed at https://github.com/hazelcast/hazelcast-csharp-client/pull/997/commits/b20446198d9edf5539c2ebf336786398a31aa63f

src/Hazelcast.Net/DistributedObjects/IHMap.Setting.cs
```diff
@@ -265,6 +265,6 @@ public partial interface IHMap<TKey, TValue> // Setting
         /// Time resolution for <paramref name="timeToLive"></paramref> is seconds. The given value is rounded to the next closest second value.
         /// </para>
         /// </remarks>
```
ihsandemir: Javadoc sync from Java missing: exception throws missing.
server config if negative, can be different from -1. can you double check with javadoc?

